PR #14, opened 2023-08-29
Integrate PQ to recall Bench

jkni:
I'll start with an apology -- missing configuration meant the previous commit of Bench didn't get line endings mangled correctly. I've fixed that in this PR and confirmed no other files are affected. Adding w=1 to the query parameters of the diff will ignore the whitespace fixes in Bench and allow easier review.

This adds Product Quantization to Bench and tests every configuration with PQ on and off at search time. The goal here is to get this functionality back in, with the expectation that changes will be needed as JVector stabilizes. I've left a few inline comments on questionable parts.

### DIFF
--- a/src/main/java/com/github/jbellis/jvector/example/Bench.java
+++ b/src/main/java/com/github/jbellis/jvector/example/Bench.java
@@ -1,12 +1,11 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
- * The ASF licenses this file to You under the Apache License, Version 2.0
- * (the "License"); you may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
+ * Copyright DataStax, Inc.
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -92,7 +92,7 @@
               <arguments>
                 <argument>-classpath</argument>
                 <classpath/>
-                <!--<argument> - -add-modules=jdk.incubator.vector</argument # Reenable once vectorization provider fixed-->
+                <argument>--add-modules=jdk.incubator.vector</argument>
```
jkni, line 109: We need this for PQ because clustering depends on this module.

src/main/java/com/github/jbellis/jvector/vector/VectorizationProvider.java
```diff
@@ -109,7 +109,8 @@ static VectorizationProvider lookup(boolean testMode) {
         throw new LinkageError(
             "PanamaVectorizationProvider is missing correctly typed constructor", e);
       } catch (ClassNotFoundException cnfe) {
-        throw new LinkageError("PanamaVectorizationProvider is missing in Lucene JAR file", cnfe);
+        LOG.warning("PanamaVectorizationProvider is missing in Lucene JAR file.");
```
jkni: Since we need vector module for PQ and we don't have PanamaVectorizationProvider yet, we need this to be nonfatal and fall back to the slow default.

src/main/java/com/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -92,7 +92,7 @@ public GraphSearcher build() {
   }
 
   public NeighborQueue search(
-      NeighborSimilarity.ExactScoreFunction scoreFunction,
```
jkni, line 95: Loosened these to allow searching with PQ approximate scores.

jbellis (approved):
LGTM, except the changes to VectorizationP shouldn't be necessary now that Jake's PR is in

Issue #22, opened 2023-08-30
update SiftTest to perform both in-memory and from-disk tests


jbellis:
Recall should be ~0.975 for in-memory and ~0.96 for from-disk (with PQ)

Issue #12, opened 2023-08-28
Put the Disk in DiskANN


jbellis:
It *may* be useful to look at https://github.com/datastax/cassandra/tree/diskann, but this only added PQ to the existing HNSW structure.  What we want to do here is implement full diskann, which means saving vectors as part of the graph structure, and fetching them as we perform BFS to minimize "seeks" (mmap paging-in) compared to reading them separately at the end of the search to re-rank in exact order.

Will also need to add a `isApproximate()` flag (or the opposite, `isExact()`) so that search knows when it needs to do the re-rank.

I *think* the right place for this is `NeighborSimilarity`.

jbellis:
(GraphIndex.getVector and ScoreFunction.isExact have been added)

jbellis:
I pushed some reorganization that I think will make DiskANN implementation straightforward -- if that is not the case then let's talk

PR #24, opened 2023-08-30
Work on putting the disk in DiskANN

jkni:
Update Searcher for re-ranking, SiftSmall to compare heap/disk, and misc assorted changes to allow that to be plumbed together.

This includes contributions for #12 and #22 . Most of the work here is quite naive as I start to put the picture together, but this is enough to start round-tripping to disk and start plugging in the DiskANN optimizations before they make sense.


pom.xml
```diff
@@ -91,6 +91,7 @@
               <arguments>
                 <argument>-classpath</argument>
                 <classpath/>
+                <argument>--add-modules=jdk.incubator.vector</argument>
```
jkni, line 94: Sift needs vector module now that it round trips to PQ on disk.

src/main/java/com/github/jbellis/jvector/disk/OnDiskGraphIndex.java
```diff
@@ -60,25 +64,36 @@ public int maxEdgesPerNode() {
...
         }
 
         public T getVector(int node) {
+            var cachedVector = vectorCache.get(node);
+            if (cachedVector != null) {
+                return cachedVector;
+            }
             try {
-                reader.seek(neighborsOffset + node * (Integer.BYTES + (long) dimension * Float.BYTES + (long) Integer.BYTES * M));
+                reader.seek(neighborsOffset +
```
jkni, line 92: We need (M+1) here because there's also a leading int for the size of the neighbors list.

src/main/java/com/github/jbellis/jvector/disk/OnDiskGraphIndex.java
```diff
@@ -88,7 +103,11 @@ public T getVector(int node) {
 
         public NodesIterator getNeighborsIterator(int node) {
             try {
-                reader.seek(neighborsOffset + (node + 1) * (Integer.BYTES + (long) dimension * Float.BYTES) + node * (long) Integer.BYTES * M);
+                // Cache full-precision coordinates per DiskANN 3.5
+                reader.seek(neighborsOffset +
+                        node * (Integer.BYTES + (long) dimension * Float.BYTES + (long) Integer.BYTES * (M + 1))
```
jkni, line 108: Same M+1 as above.

src/main/java/com/github/jbellis/jvector/disk/OnDiskVectors.java
```diff
@@ -73,7 +73,10 @@ public float[] vectorValue(int i)
 
     void readVector(int i, float[] v) throws IOException
     {
-        in.readFloatsAt(segmentOffset + 8L + i * dimension * 4L, v);
+        in.seek(segmentOffset + 8L + i * dimension * 4L);
```
jkni, line 76: Not sure about pulling this out of the interface but kept the I/O hierarchy simple while I try to figure it out.

src/main/java/com/github/jbellis/jvector/example/Bench.java
jkni, line 1: Changes here clean up some things no longer needed, take into account the reranking in search, and prep for Bench going to disk.

jbellis:
Moving this to a branch against the main repo so we can both push to it

PR #28, opened 2023-08-31
JVector 24: DiskANN

jbellis:


src/main/java/com/github/jbellis/jvector/disk/OnDiskGraphIndex.java
```diff
@@ -60,25 +64,36 @@ public int maxEdgesPerNode() {
...
     public GraphIndex.View<T> getView()
     {
         return new OnDiskView(readerSupplier.get());
     }
 
+    // TODO: This is fake generic until the reading functionality uses T
     public class OnDiskView implements GraphIndex.View<T>, AutoCloseable
     {
         private final RandomAccessReader reader;
+        private final Map<Integer, T> vectorCache;
```
jbellis: Shouldn't need the cache here.  We're guaranteed to touch each node only once in the search (Searcher's visited set makes sure of that).  Searcher should ask for the vectors when it expands a new node (right before it asks for the neighbors).

jbellis:
It's a little surprising how much slower the disk is on the toy sift dataset, but we can probably blame that on the very simple i/o provider.  If so, we don't need to fix that.

src/main/java/com/github/jbellis/jvector/example/SiftSmall.java
```diff
@@ -51,9 +88,15 @@ public static void main(String[] args) throws IOException, ExecutionException, I
         System.out.format("%d base and %d query vectors loaded, dimensions %d%n",
                 baseVectors.size(), queryVectors.size(), baseVectors.get(0).length);
 
-        var recall = testRecall(baseVectors, queryVectors, groundTruth);
+        var testDirectory = Files.createTempDirectory("SiftSmallGraphDir");
+        var testOutputPath = testDirectory.resolve("graph_test");
+        try {
+            testRecall(baseVectors, queryVectors, groundTruth, testOutputPath.toFile());
+        } finally {
+            Files.delete(testOutputPath);
```
jbellis: this always throws for me

src/main/java/com/github/jbellis/jvector/disk/OnDiskGraphIndex.java
```diff
@@ -60,25 +64,36 @@ public int maxEdgesPerNode() {
...
         public T getVector(int node) {
+            var cachedVector = vectorCache.get(node);
+            if (cachedVector != null) {
+                return cachedVector;
+            }
             try {
-                reader.seek(neighborsOffset + node * (Integer.BYTES + (long) dimension * Float.BYTES + (long) Integer.BYTES * M));
+                reader.seek(neighborsOffset +
+                        node * (Integer.BYTES + (long) dimension * Float.BYTES + (long) Integer.BYTES * (M + 1)) // earlier entries
+                        + Integer.BYTES // skip the ID
```
jbellis, line 84: thanks :)

src/main/java/com/github/jbellis/jvector/example/SiftSmall.java
jkni: Huh. I'll take a look and sort it out tomorrow if you're not already on it.

src/main/java/com/github/jbellis/jvector/disk/OnDiskGraphIndex.java
jkni: That makes sense.

src/main/java/com/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -92,17 +92,49 @@ public GraphSearcher build() {
...
+    if (initialEp == -1) {
+      return new NeighborQueue(1, false);
+    }
+    NeighborQueue results;
+    results = new NeighborQueue(topK, false);
+    searchInternal(approximateScoreFunction, results, topK, view.entryNode(), acceptOrds, visitedLimit);
+
+    // re-rank the quantized results by the original similarity function
+    // Decorate
+    // TODO: if we capture ef/actual topK here, we can avoid rerank when nn.size < topK
```
jbellis: If you mean, because the definition of recall doesn't care where it is in the results, that's true, but I think it would surprise users not to get them back in actual similarity order

jkni: Good point. I was thinking of just recall benchmarking.

jbellis:
I pushed a sketch of what I *think* is approximately the right api for the Searcher.  Reasoning:

- We should optimize the shape of the results for the common case, which is reranking
- We can throw away visitedLimit; that was a Lucene-ism that C* doesn't use.  I think Lucene uses it to prevent runaway queries, but a better way (and more consonant with how C* works) would be to pass a time limit.  We can add that back if/when we need it.
- I had to add ReRanker because I'm convinced that the vectors read during the search shouldn't leak outside of the Searcher class, so we need a way to pass those to the scoring calculation.  (Maybe it should just take a Map instead of an RAVV though, now that I think of it.)

Besides the TODOs and FIXMEs in the code, I'm also not sure where NodeScore definition should live.

jbellis:
let's add back the readFloatsAt method to RAR so that we can use Cassandra's optimized code on that side of the house

src/main/java/com/github/jbellis/jvector/disk/OnDiskGraphIndex.java
```diff
@@ -48,6 +50,8 @@ public OnDiskGraphIndex(ReaderSupplier readerSupplier, long offset)
         } catch (Exception e) {
             throw new RuntimeException("Error initializing OnDiskGraph at offset " + offset, e);
         }
+
+        //TODO: DiskANN 3.4 Cache Frequently Visited Vertices (by hops from entryNode) at load time, share between views
```
jbellis: You're right, we should add this back in.  I hacked out the old cache since the new one will be vector + neighbors instead of an overcomplicated mix of stuff, but the algorithm (populate by BFS from entry) will be the same.

src/main/java/com/github/jbellis/jvector/example/Bench.java
```diff
@@ -16,17 +16,24 @@
...
+import com.github.jbellis.jvector.disk.CompressedVectors;
+import com.github.jbellis.jvector.disk.OnDiskGraphIndex;
+import com.github.jbellis.jvector.example.util.MappedRandomAccessReader;
 import com.github.jbellis.jvector.graph.*;
 import com.github.jbellis.jvector.pq.ProductQuantization;
 import com.github.jbellis.jvector.vector.VectorEncoding;
 import com.github.jbellis.jvector.vector.VectorSimilarityFunction;
 import com.github.jbellis.jvector.vector.VectorUtil;
 import io.jhdf.HdfFile;
+import org.apache.commons.io.FileUtils;
```
jbellis: is it worth adding this dependency for a single, straightforward method?

src/main/java/com/github/jbellis/jvector/example/Bench.java
```diff
@@ -84,7 +84,7 @@ private record ResultSummary(int topKFound, int nodesVisited) { }
     private static long topKCorrect(int topK, int[] resultNodes, Set<Integer> gt) {
         int count = Math.min(resultNodes.length, topK);
         // stream the first count results into a Set
-        var resultSet = Arrays.stream(resultNodes, 0, count)
```
jbellis: Need to update the javadoc for search if you want to put it back to worst-first (which is probably the right call)

jbellis: thinking about this more.  the main benefit to worst-first is that it's similar to the old code.  from a blank slate perspective, if you're going to return an array of results it makes the most sense to have the best at the beginning.

i'm 60% sure that my first impulse was correct and we should just take the one-time hit in complexity, which is just "remember to change this on the cassandra side" and go forward with best-first.  but i'm open to being convinced otherwise.

src/main/java/com/github/jbellis/jvector/example/Bench.java
jkni: Probably not. I'll rip it back out.

src/main/java/com/github/jbellis/jvector/example/Bench.java
jkni: Switching back to best-first.

PR #37, opened 2023-09-01
Faster on-disk index flush in Bench

jkni:
This speeds up index flush by a couple orders of magnitude by using buffered I/O.


tjake (approved):
BufferedInputStream FTW

PR #40, opened 2023-09-05
Issue 38 multiple jdk builds

zznate:
Uses executions in both compile and jar tasks to do toggle which JVM is used. I most likely missed something, but the block and tackle stuff of compile and exec is working provided you follow the updated instructions in the readme. I'll add a section about how to reference the jar for the JDK version you want once I verify that with C*.


jbellis:
looks surprisingly straightforward, please lmk how getting C* to use it goes

jbellis (approved):

PR #41, opened 2023-09-05
JVECTOR-20

jkni:
Introduce Shared/Unshared annotations. Annotate RAVV methods appropriately.

Using ElementType.TYPE_USE means this travels with the returned value better through most IDEs, but I don't have a strong opinion on the choice between annotating the method and the return type.

### DIFF
--- a/src/main/java/com/github/jbellis/jvector/annotations/Shared.java
+++ b/src/main/java/com/github/jbellis/jvector/annotations/Shared.java
@@ -3,11 +3,11 @@ package com.github.jbellis.jvector.annotations;
 import java.lang.annotation.*;
 
 /**
- * Type used marked Shared indicate the Object may be reused across returning method invocations.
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/jbellis/jvector/annotations/Shared.java
```diff
@@ -0,0 +1,13 @@
+package com.github.jbellis.jvector.annotations;
+
+import java.lang.annotation.*;
+
+/**
+ * Type used marked Shared indicate the Object may be reused across returning method invocations.
```
jbellis: typo (uses)

jbellis:
> Using ElementType.TYPE_USE means this travels with the returned value better through most IDEs

LGTM, but let's add that as a comment to the interface, because METHOD is the more obvious choice.

jbellis (approved):

PR #46, opened 2023-09-06
JVECTOR-31

jkni:
Add log output for VectorizationProvider selection on old JDKs and JDK 20 happy path. Tested with hacked up local builds to get JDK 11 actually working enough to load the provider.


jbellis:
I believe this should work with 19, which has the same Panama API as 20 (but 18 does not)

jbellis:
okay, we can do it in a separate ticket

jbellis (approved):

PR #47, opened 2023-09-07
Perf improvements

tjake:
There are a number of changes here:
  
  * Add Slf4j since it was already being loaded by hdf5
  * Per species versions of dot product, this is needed because I was hitting https://bugs.openjdk.org/browse/JDK-8276212 otherwise
  * Handles any size dims PQ
  * Avoids any non parallel use of streams and moved to primarily pure java arrays in build and PQ code.

```
Before:

hdf5/nytimes-256-angular.hdf5: 289761 base and 9991 query vectors loaded, dimensions 256
PQ@128 build 65.73s,
PQ encode 26.79s,
Index   M=8 ef=60 PQ=false: top 100/1 recall 0.5890, build 14.65s, query 6.17s. -1 nodes visited
Index   M=8 ef=60 PQ=true: top 100/1 recall 0.5813, build 14.65s, query 15.90s. -1 nodes visited
Index   M=8 ef=60 PQ=false: top 100/2 recall 0.6618, build 14.65s, query 10.09s. -1 nodes visited
Index   M=8 ef=60 PQ=true: top 100/2 recall 0.6622, build 14.65s, query 26.87s. -1 nodes visited

After:
hdf5/nytimes-256-angular.hdf5: 289761 base and 9991 query vectors loaded, dimensions 256
PQ@128 build 52.59s,
PQ encode 4.56s,
Index   M=8 ef=60 PQ=false: top 100/1 recall 0.5897, build 14.19s, query 6.23s. -1 nodes visited
Index   M=8 ef=60 PQ=true: top 100/1 recall 0.5811, build 14.19s, query 13.48s. -1 nodes visited
Index   M=8 ef=60 PQ=false: top 100/2 recall 0.6631, build 14.19s, query 10.36s. -1 nodes visited
Index   M=8 ef=60 PQ=true: top 100/2 recall 0.6628, build 14.19s, query 23.38s. -1 nodes visited


```


src/main/java/com/github/jbellis/jvector/example/Bench.java
```diff
@@ -42,11 +42,6 @@
  * Tests GraphIndexes against vectors from various datasets
  */
 public class Bench {
-
```
jbellis: was this an experiment or did it make a difference?

src/main/java/com/github/jbellis/jvector/vector/VectorizationProvider.java
```diff
@@ -22,7 +22,9 @@
 import java.security.PrivilegedAction;
 import java.util.Locale;
 import java.util.Objects;
-import java.util.logging.Logger;
+
+import org.slf4j.Logger;
```
jbellis: let's not do this, it's just at startup and the hdf5 dependency is going to be moved to a separate module so core will stay clean

jbellis (approved):
commits 3 and 4 look good

src/main/java/com/github/jbellis/jvector/vector/VectorizationProvider.java
tjake: Why? do you not like slf4j? I think it's quite good compared to standard.

jbellis:
Just want to keep dependencies list small.  Ideally empty.

src/main/java/com/github/jbellis/jvector/example/Bench.java
tjake: it does help

jbellis: should we move it into SimdOps then?

tjake: Sure

PR #51, opened 2023-09-07
Track numVisited when performing graph searches

jkni:
This PR introduces a results container to return nearest neighbors alongside associated metrics from the search.


jbellis:
Ship it!

PR #55, opened 2023-09-08
Switch to maven profiles, add gh workflow for tests and add package f…

tjake:
…or jdk20 specific code


src/main/java/com/github/jbellis/jvector/vector/VectorizationProvider.java
```diff
@@ -80,12 +81,16 @@ static VectorizationProvider lookup(boolean testMode) {
...
+        var provider = (VectorizationProvider) Class.forName("com.github.jbellis.jvector.vector.jdk20.PanamaVectorizationProvider").getConstructor().newInstance();
         LOG.info("Java incubating Vector API enabled. Using PanamaVectorizationProvider.");
         return provider;
       } catch (UnsupportedOperationException uoe) {
         // not supported because preferred vector size too small or similar
-        LOG.warning("Java vector incubator API was not enabled. " + uoe.getMessage());
+        LOG.warning("Java vector API was not enabled. " + uoe.getMessage());
+        return new DefaultVectorizationProvider();
+      } catch (ClassNotFoundException e) {
+        // not supported because preferred vector size too small or similar
```
jkni: clean up copied comment that doesn't apply to this catch

pom.xml
```diff
@@ -70,116 +70,153 @@
...
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
-  </build>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-jar-plugin</artifactId>
+            <version>3.3.0</version>
```
jkni, line 185: Should we keep the jdk11 classifier on the jar in this profile? It looks to me like this will currently produce output with the same name for two jar builds with different contents

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,26 @@
...
+    paths:
+      - .github/workflows/unit-tests.yaml
+      - '**.java'
+      - '**/pom.xml'
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        jdk: [ 11, 20 ]
```
jkni, line 15: I suspect JDK 11 won't work yet. Last I looked at this (a couple days ago), a few language-level APIs from 12+ had snuck in to the codebase.

pom.xml
tjake: Why should that matter? You are gonna use one profile or the other

.github/workflows/unit-tests.yaml
tjake: I fixed this in the PR

tjake: Oh I see I would need to change all the var stuff too

pom.xml
jkni: Having distinct artifacts for publishing/use. This is the general convention for classifier. From the docs:
```classifier:
The classifier distinguishes artifacts that were built from the same POM but differ in content. It is some optional and arbitrary string that - if present - is appended to the artifact name just after the version number.
As a motivation for this element, consider for example a project that offers an artifact targeting Java 11 but at the same time also an artifact that still supports Java 1.8. The first artifact could be equipped with the classifier jdk11 and the second one with jdk8 such that clients can choose which one to use.

Another common use case for classifiers is to attach secondary artifacts to the project's main artifact. If you browse the Maven central repository, you will notice that the classifiers sources and javadoc are used to deploy the project source code and API docs along with the packaged class files.```

.github/workflows/unit-tests.yaml
jkni: var is fine, just some API conveniences we use. `var` is a language feature and would get rejected by the -source/-target compile

jkni: now that the java 20 vector stuff is actually separated from the compile, we should switch from `-source`/`-target` to `-release` for the jdk11 profile, which will also reject APIs that aren't available.

PR #56, opened 2023-09-08
Switch to ThreadLocal cached searcher in GraphIndexBuilder

jkni:
This TODO looked essentially resolved, as the ThreadLocal graphSearcher field already existed. Swapping to using a cached searcher with distinct view per thread measurably improved graph building times on my machine. Am I missing something?


jbellis:
I think I wrote that comment back when View still interacted with the (now-deleted) CompletionTracker so it would only see nodes that were added before the View was created.  If you're not seeing recall rates plummet then you should be good.  (Can't test right this second.)

jbellis (approved):

PR #61, opened 2023-09-11
Migrate to Maven multi-module project and produce multirelease jar

jkni:
This PR reworks the build system to build a multirelease jar by using Maven's multimodule features. This prevents the need to publish multiple artifacts with classifiers. It also allows integration with IDEs with the appropriate language features/APIs per module.

Generally, the top-level project `jvector-parent` contains common configuration. `jvector-base` includes all common code. `jvector-twenty` includes code using Java 20+ language features/APIs. This will be included in the multirelease jar but only run on JVMs supporting this release level. The `jvector-multirelease` module is the only module that produces actual installed artifacts. It installs a multirelease jar with coordinates io.github.jbellis:jvector and a corresponding sources jar. The README has been updated for common uses.

Please let me know any workflows that you are concerned won't be supported by this approach. It seems to be the best Maven-based approach for the needs of the project and its developers, but I understand it introduces some additional complexity. The goal is to capture this complexity at build-system-write-time rather than when the developer interacts with the build system. For common use cases, the developer will use the same `mvn` commands as if this were a single module project.


jvector-multirelease/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.github.jbellis</groupId>
+        <artifactId>jvector-parent</artifactId>
+        <version>0.9-SNAPSHOT</version>
+    </parent>
+    <artifactId>jvector</artifactId>
+    <name>JVector multirelease jar</name>
+    <properties>
+        <maven.install.skip>false</maven.install.skip>
```
zznate, line 15: Nit: This is the default, do we need it as an explicit property? With the more obscure plugins, im +1 to be explicit about key settings, but can probably remove this.

jvector-multirelease/src/assembly/mrjar.xml
```diff
@@ -0,0 +1,37 @@
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.3"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.3 http://maven.apache.org/xsd/assembly-1.1.3.xsd">
+    <id>mvjar</id>
+    <formats>
+        <format>jar</format>
+    </formats>
+    <includeBaseDirectory>false</includeBaseDirectory>
+    <moduleSets>
```
zznate, line 9: I like this moduleSet approach - seems much cleaner.

pom.xml
```diff
@@ -1,242 +1,131 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.35</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+    <modules>
+        <module>jvector-base</module>
```
zznate, line 127: We are going to want to make an `examples` so we dont ship that code in the core jar, but that is out of scope for this issue.

zznate (approved):
Small nit or two with extraneous properties - up to you. Otherwise looks good and does what we want: 
```
jar -tf jvector-twenty/target/jvector-twenty-0.9-SNAPSHOT.jar
META-INF/
META-INF/MANIFEST.MF
com/
com/github/
com/github/jbellis/
com/github/jbellis/jvector/
com/github/jbellis/jvector/vector/
META-INF/maven/
META-INF/maven/io.github.jbellis/
META-INF/maven/io.github.jbellis/jvector-twenty/
META-INF/jpms.args
com/github/jbellis/jvector/vector/PanamaVectorUtilSupport.class
com/github/jbellis/jvector/vector/SimdOps.class
com/github/jbellis/jvector/vector/PanamaVectorizationProvider.class
META-INF/maven/io.github.jbellis/jvector-twenty/pom.xml
META-INF/maven/io.github.jbellis/jvector-twenty/pom.properties

```

pom.xml
jkni: Agreed. I think we should pull that out shortly after this PR is merged.

jvector-multirelease/pom.xml
jkni: This overrides the top-level skip in the parent pom, so we do need it here. I originally pushed the opt-in skipping to all other modules, but setting it at the parent and explicitly overriding it in the `jvector-multirelease` module is less verbose and seemed more aligned with our intent, which is that this should be the _only_ module installing.

zznate: That makes sense. LGTM!

PR #68, opened 2023-09-14
Build rework for release. Split examples out into separate project.

jkni:
Tagging @jbellis for review just to make sure we're on the same page with what's getting pulled into jvector-examples as a separate project. Don't worry about the build parts any more than you want to.


jbellis:
#collaborating my thought was that we'd include SimpleMappedReader (I guess we should include a SimpleMappedSupplier too) and leave the rest of it in example, both to keep the surface area small and to avoid the impression we're trying to provide something production-ready.  Do you think it makes more sense to include the MMapReader too?

jbellis:
structure LGTM

jbellis (approved):

PR #71, opened 2023-09-15
Change package from com.github.jbellis to io.github.jbellis

jkni:
This is the automated IntelliJ refactor. Along the way, it also refactored some imports to the IntelliJ preferred style, which we might as well take until we dictate a specific import style.


jbellis (approved):
```
$ find . -name "*.java" |xargs grep com.github
```

LGTM

PR #76, opened 2023-09-15
JVECTOR-62: various PQ options, support for fvec files

dlg99:
https://github.com/jbellis/jvector/issues/62

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
@@ -31,8 +31,12 @@ import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
 import java.io.BufferedOutputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
+import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/Hdf5Loader.java
```diff
@@ -20,33 +20,49 @@
...
         float[][] baseVectors;
         float[][] queryVectors;
         int[][] groundTruth;
+
         Path path = Paths.get(pathStr);
-        try (HdfFile hdf = new HdfFile(path)) {
-            baseVectors = (float[][]) hdf.getDatasetByPath("train").getData();
-            queryVectors = (float[][]) hdf.getDatasetByPath("test").getData();
-            groundTruth = (int[][]) hdf.getDatasetByPath("neighbors").getData();
+        if (isHdf5) {
```
jbellis: don't do this, make the caller use SiftLoader where appropriate

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/Hdf5Loader.java
```diff
@@ -100,10 +116,31 @@ else if (pathStr.contains("euclidean")) {
...
-        System.out.format("%n%s: %d base and %d query vectors loaded, dimensions %d%n",
-                          pathStr, scrubbedBaseVectors.size(), scrubbedQueryVectors.size(), scrubbedBaseVectors.get(0).length);
+        String details = String.format("%s: %d base and %d query vectors loaded, dimensions %d",
+                pathStr, scrubbedBaseVectors.size(), scrubbedQueryVectors.size(), scrubbedBaseVectors.get(0).length);
 
-        return new DataSet(path.getFileName().toString(), similarityFunction, scrubbedBaseVectors, scrubbedQueryVectors, gtSet);
+        return new DataSet(path.getFileName().toString(), similarityFunction, scrubbedBaseVectors, scrubbedQueryVectors, gtSet, details);
+    }
+
+    private static int[][] toIntArr(ArrayList<HashSet<Integer>> hashSets) {
```
jbellis: this won't be necessary if using SiftLoader directly

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/Hdf5Loader.java
jbellis: it looks like you did this to re-use the scrub code, but (1) the nytimes dataset in hdf5 format is the only one that we've needed to scrub, (2) if we do need to re-use it, let's move it out of Hdf5Loader (into DataSet perhaps)

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/DataSet.java
```diff
@@ -28,7 +28,9 @@ public class DataSet {
     public final List<float[]> queryVectors;
     public final List<? extends Set<Integer>> groundTruth;
 
-    public DataSet(String name, VectorSimilarityFunction similarityFunction, List<float[]> baseVectors, List<float[]> queryVectors, List<? extends Set<Integer>> groundTruth) {
+    public final String details;
```
jbellis, line 37: i don't love jamming the parsing information into DataSet, what was wrong with printing it when it's loaded?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
```diff
@@ -43,37 +47,46 @@
...
+
+    private static void testRecall(GraphResult graph, int topK, int M, int efConstruction, boolean useDisk, int overquery, DataSet ds, CompressedVectors cv) throws IOException {
+        long start = System.nanoTime();
+        var pqr = performQueries(ds, graph.floatVectors, useDisk ? cv : null, useDisk ? graph.onDiskGraph : graph.onHeapGraph, topK, topK * overquery, queryRuns);
+        var recall = ((double) pqr.topKFound) / (queryRuns * ds.queryVectors.size() * topK);
+        System.out.format("  Query PQ=%b top %d/%d recall %.4f in %.2fs after %s nodes visited%n",
+                useDisk, topK, overquery, recall, (System.nanoTime() - start) / 1_000_000_000.0, pqr.nodesVisited);
+    }
 
+    private static GraphResult buildGraph(int M, int efConstruction, DataSet ds, Path graphPath) throws IOException {
```
jbellis, line 61: splitting out buildGraph is a good idea and probably overdue

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
```diff
@@ -131,55 +144,107 @@ private static ResultSummary performQueries(DataSet ds, RandomAccessVectorValues
     }
 
     public static void main(String[] args) throws IOException {
-        System.out.println("Heap space available is " + Runtime.getRuntime().maxMemory());
```
jbellis, line 135: please review and either delete commented-out lines, or restore them, before submitting PR

jbellis:
i pushed some minimal code to get the ada002 dataset running in Bench (does not do a PQ grid tho)

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/DataSet.java
dlg99: It's more of stats info (ideally, if we need it in different format).
It is better to have an option to construct output later/as needed, in the format that suitable for parsers instead of changing the parsers after every refactoring that changes the order/frequency of output.

PR #79, opened 2023-09-18
JVECTOR-78 Make Graph and Graph.View AutoCloseable

mike-tr-adamson:


jbellis:
I ended up removing the default method, I think it's good to make people think explicitly about how they implement that

PR #84, opened 2023-09-18
Fix recall regression for centered PQ with non-dot product metrics

jkni:
Now that decodedSimilarity caller is responsible for centering comparison vector, decodedSimilarity shouldn't uncenter decoded vector.


jbellis:
good catch.  i'd lean towards leaving decodeCentered package-private until/unless we need it somewhere else.  otherwise LGTM.

jbellis:
ah, right.  I guess I'm 51% in favor of moving CV to the pq package but leaving it public also wfm.

PR #88, opened 2023-09-19
Fix code coverage in IntelliJ

jkni:
jvector-base/jvector-twenty must be compile time deps for the classes files to be included in the scope for code coverage generation.


jbellis:
Screenshot from 2023-09-19 12-09-04

jbellis (approved):

PR #89, opened 2023-09-19
Implement optimized decoded square distance

jkni:
This also implements the necessary SimdOps backing methods, mirroring the dotProduct implementation.

As a side benefit, PQ encoding is about twice as fast.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/ProductQuantization.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/ProductQuantization.java
@@ -157,13 +157,14 @@ public class ProductQuantization {
      */
     public float decodedSquareDistance(byte[] encoded, float[] other) {
         float sum = 0.0f;
+        var a = scratch.get();
         for (int m = 0; m < M; ++m) {
             int offset = subvectorSizesAndOffsets[m][1];
             int centroidIndex = Byte.toUnsignedInt(encoded[m]);
             float[] centroidSubvector = codebooks[m][centroidIndex];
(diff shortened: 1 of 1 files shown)


jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/SimdOps.java
```diff
@@ -128,8 +128,6 @@ static float dotProduct64(float[] v1, int v1offset, float[] v2, int v2offset, in
         if (length == FloatVector.SPECIES_64.length())
             return dot64(v1, v1offset, v2, v2offset);
 
-        assert length == 3; // or we should be calling dot 128
```
jkni, line 131: note: this is removed because it's not quite right. You could be down this code path with a one-dimensional vector, which is valid everywhere else in the codebase

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/SimdOps.java
```diff
@@ -316,31 +314,160 @@ static float cosineSimilarity(byte[] v1, byte[] v2) {
...
+    }
+
+    static float squareDistance256(float[] v1, int offset1, float[] v2, int offset2) {
+        var a = FloatVector.fromArray(FloatVector.SPECIES_256, v1, offset1);
+        var b = FloatVector.fromArray(FloatVector.SPECIES_256, v2, offset2);
+        var diff = a.sub(b);
+        return diff.mul(diff).reduceLanes(VectorOperators.ADD);
+    }
+
+    static float squareDistance512(float[] v1, int offset1, float[] v2, int offset2) {
```
jkni: This should be `squareDistancePreferred`, will fix

PR #90, opened 2023-09-20
Initial refactor of vector types to allow on-heap and off-heap repres…

tjake:
…entations


tjake:
Sadly, this approach is much slower:

```
Benchmark                                 Mode  Cnt        Score        Error   Units
SimilarityBench.dotProductArray          thrpt    3  4551812.518 ± 751877.546  ops/ms
SimilarityBench.dotProductArray:·async   thrpt               NaN                  ---
SimilarityBench.dotProductMemory         thrpt    3  2195843.354 ± 177117.603  ops/ms
SimilarityBench.dotProductMemory:·async  thrpt               NaN                  ---
SimilarityBench.dotProductNative         thrpt    3   483524.241 ±  29822.702  ops/ms
SimilarityBench.dotProductNative:·async  thrpt               NaN                  ---
```

PR #94, opened 2023-09-22
Fragment cache

jbellis:


jbellis:
I'm not worrying about the case where the graph is too small to benefit from caching, b/c C* already tries to detect tiny graphs and do a brute force comparison anyway.  Can revisit later if necessary.

jvector-base/src/main/java/io/github/jbellis/jvector/pq/CompressedDecoder.java
```diff
@@ -0,0 +1,109 @@
...
+                    int offset = pq.subvectorSizesAndOffsets[i][1];
+                    float[] centroidSubvector = pq.codebooks[i][j];
+                    switch (vsf) {
+                        case DOT_PRODUCT:
+                            cache[i][j] = VectorUtil.dotProduct(centroidSubvector, 0, centeredQuery, offset, centroidSubvector.length);
+                            break;
+                        case EUCLIDEAN:
+                            cache[i][j] = VectorUtil.squareDistance(centroidSubvector, 0, centeredQuery, offset, centroidSubvector.length);
+                            break;
+                        default:
```
jkni, line 38: cosine too expensive to be worth precomputing?

jvector-base/src/main/java/io/github/jbellis/jvector/pq/CompressedDecoder.java
```diff
@@ -0,0 +1,109 @@
...
+                            throw new UnsupportedOperationException("Unsupported similarity function " + vsf);
+                    }
+                }
+            }
+            return cache;
+        }
+
+        @Override
+        public float similarityTo(int node2) {
+            return (1 + decodedSimilarity(cv.get(node2), cache)) / 2;
```
jkni: This changes the Euclidean similarity definition. Is that intentional?

jbellis: No, thanks for the catch.

PR #100, opened 2023-09-25
KMeansPlusPlusClusterer optimizations

jkni:
This delivers about a 5-7% reduction in the time needed to generate PQ for every dataset we have in Bench.

About half of this PR we should do regardless (removing centroidDistances, doc updates). The rest of the PR removes some collections work that consistently showed up in profiles and reduces the amount of repeated work we do as PQ converges by keeping running averages that can be updated with each clustering round.


jbellis (approved):
LGTM

PR #103, opened 2023-09-27
Give non default initial capacity

tjake:


jbellis:
nsize0 is the number of neighbors each member of this.nodes is allowed, so they're not related like this

better default would be 100 or 1000

tjake:
This was a red herring.

PR #117, opened 2023-10-09
Deletes

### LINKED ISSUE #115: How to remove vector from the index?
how do I remove a vector from the index?

jbellis:
Adds markNodeDeleted to GraphIndexBuilder, and adds functionality to cleanup() that removes deleted nodes and repairs the connections of their neighbors.

Fixes #115


jbellis:
Not yet done:
- Tests that exercise RenumberingGraphIndex 
- A way to communicate to the caller how their vectors got renumbered
- (Optional) integrate this with other renumbering the caller may want to perform (e.g. Cassandra prefers to have vector ids match row ids when there is a 1:1 correspondence)

jbellis:
PR is a bit of a mess, largely because I started writing it on my desktop, wrote a different part on my macbook, and then had to merge both parts with the big threadlocal refactor, so apologies in advance for that.

jbellis:
> A way to communicate to the caller how their vectors got renumbered

I think there's actually two different use cases.

The first is Cassandra-like, where the caller wants to make the vector ordinals match some other data source (C* rowIds).  For that we'll have the caller pass a a mapping function.

But if you're building, say, vector embeddings of the files in an intellj project, you don't care about that, you just need to map vectors to files or methods.  And there isn't really a state where the index is "done," you're going to keep modifying it as edits are made to the files.  In that case it's a better fit to save the index, "holes" and all, and be able to reload it in the same state to continue modifying it.

The most recent commit here adds save and load methods to handle this.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
```diff
@@ -100,6 +110,35 @@ public void cleanup() {
     neighborsRef.getAndUpdate(this::removeAllNonDiverse);
   }
 
+  /**
+   * @return true if we had deleted neighbors
+   */
+  public boolean removeDeletedNeighbors(Bits deletedNodes) {
+    AtomicBoolean found = new AtomicBoolean();
```
jkni, line 117: I don't follow why we need an AtomicBoolean here

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
```diff
@@ -141,18 +180,35 @@ public void insertDiverse(NeighborArray natural, NeighborArray concurrent) {
...
+        toMerge = mergeNeighbors(natural, concurrent);
+      }
+
       // merge all the candidates into a single array and compute the diverse ones to keep
       // from that.  we do this first by selecting the ones to keep, and then by copying
       // only those into a new NeighborArray.  This is less expensive than doing the
       // diversity computation in-place, since we are going to do multiple passes and
       // pruning back extras is expensive.
-      var merged = mergeNeighbors(mergeNeighbors(natural, current), concurrent);
+      var merged = mergeNeighbors(natural, toMerge);
```
jkni: `current` seems to have been lost in the shuffle

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
jbellis: it's just the easiest way to represent a reference to a boolean, so we can use it outside the getAndUpdate

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
jbellis: you're right.  added a failing test and fixed.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -16,267 +16,469 @@
...
+
+        // repair affected nodes
+        for (var node : affectedLiveNodes) {
+            addNNDescentConnections(node);
+        }
+
+        // reset deleted collection
+        deletedNodes.clear();
+
+        return graph.ramBytesUsedOneNode(0);
```
jkni: this doesn't seem to match the documented return value. That should depend on the number of deletedNodes, and graph.ramBytesUsedOneNode does not.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -16,267 +16,469 @@
...
+
+                // add random connections if we've dropped below minimum
+                int minConnections = 1 + graph.maxDegree() / 2;
+                if (neighbors.size() < minConnections) {
+                    // create a NeighborArray of random connections
+                    NeighborArray randomConnections = new NeighborArray(graph.maxDegree() - neighbors.size());
+                    // doing actual sampling-without-replacement is expensive so we'll loop a fixed number of times instead
+                    for (int i = 0; i < 2 * graph.maxDegree(); i++) {
+                        int randomNode = liveNodes[R.nextInt(liveNodes.length)];
+                        if (randomNode != node && !randomConnections.contains(randomNode)) {
```
jkni, line 283: Should we also filter out whether neighbors already contains the randomNode selected here?

jvector-base/src/main/java/io/github/jbellis/jvector/disk/OnDiskGraphIndex.java
```diff
@@ -154,15 +182,22 @@ public static <T> void write(GraphIndex<T> graph, RandomAccessVectorValues<T> ve
         out.writeInt(graph.maxDegree());
 
         // for each graph node, write the associated vector and its neighbors
-        for (int node = 0; node < graph.size(); node++) {
-            out.writeInt(node); // unnecessary, but a reasonable sanity check
-            Io.writeFloats(out, (float[]) vectors.vectorValue(node));
+        var newOrdinals = new HashSet<Integer>();
+        for (int originalOrdinal = 0; originalOrdinal <= graph.getMaxNodeId(); originalOrdinal++) {
```
jkni: I don't think renumbering works because we flush in original order. Consider some graph node with originalOrdinal 0 that gets renumbered to `x`, both as its ordinal and in neighbor lists. We'll use the offset `x` when getting the graph node with originalOrdinal 0 from disk, but its on disk at its original ordinal offset, not `x`. You can see this in Bench if you supply a derangement instead of the identity function. Recall will drop to near zero.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jbellis: yes.  i will merge ConcurrentNeighborArray into NA so we get deduplication everywhere, the only reason they were split apart at all was to reduce my blast radius on Lucene.

jvector-base/src/main/java/io/github/jbellis/jvector/disk/OnDiskGraphIndex.java
jbellis: ugh.

jbellis: fixed.

jvector-base/src/main/java/io/github/jbellis/jvector/disk/OnDiskGraphIndex.java
```diff
@@ -171,17 +174,34 @@ public void close() throws IOException {
...
      * @param vectors the vectors associated with each node
-     * @param ordinalMapper A function that maps from the graph's ordinals to the ordinals in the output.
-     *                      For simple use cases this can be the identity function.  To deal with deleted nodes,
-     *                      or to renumber the nodes to match rows or documents elsewhere, you may provide
-     *                      something custom.  The mapper must map from the graph's ordinals
-     *                      [0..getMaxNodeId()], to the output's ordinals [0..size()), with no gaps and
-     *                      no duplicates.
+     * @param out the output to write to
+     *
+     * If any nodes have been deleted, you must use the overload specifying `oldToNewOrdinals` instead.
```
jkni, line 179: I'm not sure why this is necessary, since `getSequentialRenumbering` from line 191 should work for this case. Is it to force the user to be aware and make a choice?

jvector-tests/src/test/java/io/github/jbellis/jvector/disk/TestOnDiskGraphIndex.java
```diff
@@ -0,0 +1,187 @@
...
+        {
+            // 0 -> 1
+            assertTrue(getNeighborNodes(onDiskView, 0).contains(1));
+            // 1 -> 0
+            assertTrue(getNeighborNodes(onDiskView, 1).contains(0));
+        }
+    }
+
+    @Test
+    public void testReordingReumbering() throws IOException {
```
jkni: rename? Not sure how much of this is a typo but at least Reumbering -> Renumbering

jvector-base/src/main/java/io/github/jbellis/jvector/disk/OnDiskGraphIndex.java
jbellis: yes.  (and that is why gSR is public, so they can just use that if they want.)

jvector-tests/src/test/java/io/github/jbellis/jvector/disk/TestOnDiskGraphIndex.java
jbellis: wow, two typos in the same method name.  I don't *think* I was drunk when I wrote this ... :)

siddhsql:
does this use the same algorithm for deletes as in https://arxiv.org/abs/2105.09613 or something different? looking at it, it seems to be doing something different. could you explain?

jbellis:
JVector's approach is essentially "pretend each node that loses connections due to the deletion is newly added to the graph and rebuild its connections that way."

FreshDiskANN's approach is definitely less expensive, if it actually maintains link quality I'd be happy to switch to that.

siddhsql:
i tried deleting nodes from the graph. the cleanup operation takes more time than rebuilding the entire index. with 1M vectors, building the index took 3 minutes. then i deleted 20% of the vectors and the cleanup took 17 minutes.

jbellis:
That does sound excessive.

PR #121, opened 2023-10-09
Fix num/denom distortion for first round of assignments when clustering

jkni:
The old assignments on first run do not have any value, but they lie in the valid space of assignments (centroid 0), so centroid 0 gets distorted during the first round of assignment. This PR also includes a test that exposes this issue and exercises a surprising amount of the interesting bits of PQ/KMPP.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/KMeansPlusPlusClusterer.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/KMeansPlusPlusClusterer.java
@@ -157,7 +157,8 @@ public class KMeansPlusPlusClusterer {
 
     /**
      * Assigns points to the nearest cluster.  The results are stored as ordinals in `assignments`.
-     * This method relies on valid assignments existing following a run of initializeAssignedPoints.
+     * This method relies on valid assignments existing from either initializeAssignedPoints or
+     * a previous invocation of this method.
      *
      * @return the number of points that changed clusters
      */


jvector-base/src/main/java/io/github/jbellis/jvector/pq/KMeansPlusPlusClusterer.java
```diff
@@ -143,18 +143,24 @@ private float[][] chooseInitialCentroids(float[][] points) {
...
+     * @return the number of points that changed clusters
      */
-    private int assignPointsToClusters() {
+    private int assignPointsToClusters(boolean firstRun) {
         int changedCount = 0;
 
         for (int i = 0; i < points.length; i++) {
             float[] point = points[i];
-            var oldAssignment = assignments[i];
+            var oldAssignment = firstRun ? -1 : assignments[i];
```
jbellis, line 155: could we get rid of the firstRun flag and the if statements by just calling Arrays.fill(assignments, -1) to start with?

jbellis (changes requested):

jvector-base/src/main/java/io/github/jbellis/jvector/pq/KMeansPlusPlusClusterer.java
```diff
@@ -143,18 +143,24 @@ private float[][] chooseInitialCentroids(float[][] points) {
 
     /**
      * Assigns points to the nearest cluster.  The results are stored as ordinals in `assignments`
+     *
+     * @param firstRun whether this is the first assignment of points, which means the previous assignment is meaningless
+     * @return the number of points that changed clusters
      */
-    private int assignPointsToClusters() {
+    private int assignPointsToClusters(boolean firstRun) {
```
jbellis: i don't love "flag parameters" that change behavior, could we split this into assignInitialPoints and assignIteratedPoints (names subject to change)?

jbellis (changes requested):

jvector-base/src/main/java/io/github/jbellis/jvector/pq/KMeansPlusPlusClusterer.java
jkni: Sure, I went back-and-forth on exactly this and didn't have strong feelings. Pushing an update.

jbellis (approved):

Issue #36, opened 2023-09-01
mvn release task that checks for license headers?


jbellis:
Not sure if this is a big ask or a small one.  If it's a lot of work we can skip it.

zznate:
Let me check - there should be a couple of different plugins that make this straight forward.

zznate:
https://github.com/jbellis/jvector/pull/129 Should address license checking with added licenses on missing files and sensible excludes list.

PR #129, opened 2023-10-18
Addresses issue #36 by adding license header checks. Added headers on…

zznate:
… missing files, added exlsion list for files that dont necessarily need them

### DIFF
--- a/jvector-base/pom.xml
+++ b/jvector-base/pom.xml
@@ -11,4 +11,15 @@
     </parent>
     <artifactId>jvector-base</artifactId>
     <name>Base</name>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <configuration>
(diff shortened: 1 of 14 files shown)


jvector-base/pom.xml
```diff
@@ -11,4 +11,15 @@
     </parent>
     <artifactId>jvector-base</artifactId>
```
jbellis, line 12: does this need to go in Base?

zznate: It's sort of annoying, but by design, Maven modules have all their variables scoped, so there is not a real easy way to refer to the configuration settings in the parent.  Placed here, it gets included as a transitive dependency by all the other modules.

jbellis: okay, and this is a "plugin" which is distinct from a "dependency" so jvector users aren't going to pull down rat as well, is that right?

zznate: Correct. The plugin machinery is specific to Maven workflows. It's stored in the local repository, but has nothing to do with the produced artifacts.

jbellis: Ship it!

PR #133, opened 2023-10-27
Cherry-pick various bench improvements from PR #76.

jkni:
Split out graph builds in Bench, provide dataset filtering via args, match all dataset types in plot_output.

In addition, when applying changes, I discovered cached PQ builds don't output when selected, which breaks plot_output. This is fixed in this PR.

Credit to @dlg99 for his work here. I'm really only bringing over changes and shuffling a few things around.

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
@@ -205,14 +205,14 @@ public class Bench {
         var pattern = Pattern.compile(regex);
 
         if (pattern.matcher("wikipedia_squad/100k/e5-small-v2").find()) {
-            DownloadHelper.maybeDownloadFvecs(List.of("intfloat_e5-small-v2_100000"));
+            DownloadHelper.maybeDownloadFvecs("intfloat_e5-small-v2_100000");
             var e5set = loadE5SmallData("wikipedia_squad/100k");
             gridSearch(e5set, compressionGrid, mGrid, efConstructionGrid, efSearchGrid);
             cachedCompressors.clear();
         }
 
         if (pattern.matcher("wikipedia_squad/100k/ada_002").find()) {
-            DownloadHelper.maybeDownloadFvecs(List.of("ada_002_100000"));
+            DownloadHelper.maybeDownloadFvecs("ada_002_100000");
             var adaSet = loadWikipediaData("wikipedia_squad/100k");
             gridSearch(adaSet, compressionGrid, mGrid, efConstructionGrid, efSearchGrid);
             cachedCompressors.clear();
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/DownloadHelper.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/DownloadHelper.java
@@ -51,23 +51,21 @@ public class DownloadHelper {
                 .credentialsProvider(AnonymousCredentialsProvider.create());
     }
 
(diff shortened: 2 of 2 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
```diff
@@ -24,13 +24,7 @@
...
 import io.github.jbellis.jvector.example.util.ReaderSupplierFactory;
 import io.github.jbellis.jvector.example.util.SiftLoader;
-import io.github.jbellis.jvector.graph.GraphIndex;
-import io.github.jbellis.jvector.graph.GraphIndexBuilder;
-import io.github.jbellis.jvector.graph.GraphSearcher;
-import io.github.jbellis.jvector.graph.ListRandomAccessVectorValues;
-import io.github.jbellis.jvector.graph.NeighborSimilarity;
-import io.github.jbellis.jvector.graph.RandomAccessVectorValues;
-import io.github.jbellis.jvector.graph.SearchResult;
+import io.github.jbellis.jvector.graph.*;
```
jbellis: can we not churn this?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
```diff
@@ -220,7 +240,14 @@ private static void gridSearch(DataSet ds, List<Integer> pqGrid, List<Integer> m
         try {
             for (int M : mGrid) {
                 for (int efC : efConstructionGrid) {
-                    testRecall(M, efC, pqGrid, diskOptions, efSearchFactor, ds, testDirectory);
+                    var graphPath = testDirectory.resolve("graph" + M + "-" + efC + ds.name);
+                    try {
+                        GraphResult graph = buildGraph(M, efC, ds, graphPath);
```
jbellis: is this actually an improvement?  we have to create a wrapper class in order to ... avoid building the graph close to where it's used?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
jkni: Agreed on not churning, but in this case, IntelliJ automatically collapsed them as an additional import from this package pushed past the default threshold for wildcard imports. I'm happy to uncollapse them/roll in a project style guide on when to collapse, but I don't want to be fighting IntelliJ constantly. I don't have an opinion on the magic number to collapse at

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
jkni: TBH, I retained this largely due to this comment https://github.com/jbellis/jvector/pull/76#discussion_r1329100998 I don't have particularly strong feelings here and would be happy to go either direction

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
jbellis: Yes, I changed this to a really big number because it bugs me to see it churn between wildcard to not wildcard as requirements change.  Not sure of the best way to have a project-wide code style.

jkni: Cool, I'm happy with just socializing it as a style norm. I'll update my JVector project and push the change. Thanks!

jbellis:
Looks like it makes more sense now to make maybeDownloadFvecs take a single parameter.  Otherwise LGTM.

jbellis (approved):

PR #137, opened 2023-10-31
add threshold parameter to search, and add futureProbabilityAboveThreshold to determine when we've searched deep enough

jbellis:


jvector-base/pom.xml
```diff
@@ -11,6 +11,15 @@
     </parent>
     <artifactId>jvector-base</artifactId>
     <name>Base</name>
+
+    <dependencies>
```
jkni: We'll want to declare this as a dependency in the top-level pom.xml for it to get published correctly with the multirelease jar in the package phase. With how the build is currently structure, jvector-base is available to jvector-multirelease, but it won't inherit its dependencies (dependencies gets flattened down from the parent). We should also update the README to remove the zero dep note.

jbellis:
(merged manually)

PR #138, opened 2023-11-02
Reconnect orphaned nodes in cleanup()

jbellis:


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -165,11 +168,77 @@ public void cleanup() {
...
+            IntStream.range(0, entryNeighbors.size).parallel().forEach(node -> {
+                findConnected(connectedNodes, entryNeighbors.node[node]);
+            });
+
+            // reconnect unreachable nodes
+            try (var gs = graphSearcher.get();
+                 var v1 = vectors.get();
+                 var v2 = vectorsCopy.get())
+            {
+                var connectionTargets = new HashSet<Integer>(nReconnected.get());
```
jkni: Isn't `nReconnected` always 0 here, since it's reinitialized each time through the loop?

jbellis: you're right, thanks

Issue #139, opened 2023-11-03
Unable to build: Too many files with unapproved license


Holmistr:
Hi,

I was trying to build the project from the latest `main` branch using `mvn clean install -DskipTests`. I got the following build error:

```
10:36:18,286 [WARNING] Files with unapproved licenses:
  /home/jholusa/IdeaProjects/research/jvector/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/Test2DThreshold.java
  /home/jholusa/IdeaProjects/research/jvector/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestSearchProbability.java
10:36:18,287 [INFO] ------------------------------------------------------------------------
10:36:18,287 [INFO] Reactor Summary for JVector parent 1.0.3-SNAPSHOT:
10:36:18,287 [INFO] 
10:36:18,287 [INFO] JVector parent ..................................... SUCCESS [  8.407 s]
10:36:18,287 [INFO] Base ............................................... SUCCESS [  2.038 s]
10:36:18,287 [INFO] Twenty ............................................. SUCCESS [  0.832 s]
10:36:18,288 [INFO] JVector Examples ................................... SUCCESS [  1.396 s]
10:36:18,288 [INFO] Tests .............................................. FAILURE [  4.418 s]
10:36:18,288 [INFO] JVector ............................................ SKIPPED
10:36:18,288 [INFO] ------------------------------------------------------------------------
10:36:18,288 [INFO] BUILD FAILURE
10:36:18,288 [INFO] ------------------------------------------------------------------------
10:36:18,288 [INFO] Total time:  17.150 s
10:36:18,289 [INFO] Finished at: 2023-11-03T10:36:18+01:00
10:36:18,289 [INFO] ------------------------------------------------------------------------
10:36:18,289 [ERROR] Failed to execute goal org.apache.rat:apache-rat-plugin:0.15:check (default) on project jvector-tests: Too many files with unapproved license: 2 See RAT report in: /home/jholusa/IdeaProjects/research/jvector/jvector-tests/target/rat.txt -> [Help 1]
```

Looks like a trivial omit of licenses in those files. I'm sending a PR in a minute.

PR #140, opened 2023-11-03
[#139] Fix missing licenses preventing from successfull build

Holmistr:
Fix for https://github.com/jbellis/jvector/issues/139


jbellis:
Thanks!  I added the missing headers independently so we don't need to set up a CLA for this one.

Holmistr:
Sure, makes total sense :)

PR #141, opened 2023-11-06
Fix running single test using Maven commandline

jkni:
This surefire configuration was vestigial, since all tests are in the jvector-tests module.


jbellis:
can you document the right way to run a single test in README?

jbellis (approved):

PR #150, opened 2023-11-10
vectorsEncountered not always in sync with resultQueue, causing NPE when breaking out of loop due to threshold probability

jbellis:


jbellis:
I believe the original, somewhat clunky structure we inherited from Lucene is due to premature optimization.  Bench doesn't see a difference in performance for 2D vectors (which should be some of the fastest comparisons / the rest of searchInternal should matter relatively more).

PR #152, opened 2023-11-14
Fix usage of null acceptOrds in SiftSmall example

vbekiaris:
Since #117, `acceptOrds` should not be `null`. Instead, `Bits.ALL` should be used.
Also updated `GraphSearcher#search` javadoc to match implementation.


jbellis:
Hi Vassilis, thanks for catching this!  Would you mind signing the DataStax CLA at https://cla.datastax.com/ ?

vbekiaris:
> Would you mind signing the DataStax CLA at https://cla.datastax.com/ ?

Done

jbellis:
Merged. Thanks!

PR #161, opened 2023-12-01
Add a utility to conditionally submit parallel tasks to ForkJoinPool

mdogan:
Added `ParallelExecutor` utility to submit parallel stream tasks to the `PhysicalCoreExecutor` only if the caller thread is not a `ForkJoinPool` worker thread.

If the caller is already inside a `ForkJoinPool`, then task will be run on the caller thread directly. That way, parallelization of the streams (index building or compressed vector computation) can be defined dynamically by the caller `ForkJoinPool`.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -20,7 +20,7 @@ import io.github.jbellis.jvector.annotations.VisibleForTesting;
 import io.github.jbellis.jvector.disk.RandomAccessReader;
 import io.github.jbellis.jvector.util.AtomicFixedBitSet;
 import io.github.jbellis.jvector.util.Bits;
-import io.github.jbellis.jvector.util.PhysicalCoreExecutor;
+import io.github.jbellis.jvector.util.ParallelExecutor;
 import io.github.jbellis.jvector.util.PoolingSupport;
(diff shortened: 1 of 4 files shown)


jbellis:
Are you building multiple graphs in parallel?  We actually have a synchronized { } block in Cassandra so we can max out the parallel PQ computations on one graph at a time, I think that is going to be better for most use cases.

```
        synchronized (CassandraOnHeapGraph.class)
        {
            compressor = ProductQuantization.compute(vectorValues, bytesPerVector, false);
            assert !vectorValues.isValueShared();
            encoded = compressVectors(reverseOrdinalMapper, compressor);
        }

        // save (outside the synchronized block, this is io-bound not CPU)
        CompressedVectors cv;
        if (compressor instanceof BinaryQuantization)
            cv = new BQVectors((BinaryQuantization) compressor, (long[][]) encoded);
        else
            cv = new PQVectors((ProductQuantization) compressor, (byte[][]) encoded);
        cv.write(writer);
```

mdogan:
Yes, I want to allow building multiple graphs in parallel. But also I want to reserve some cpu cores to each graph and get a stable build time regardless of the other indexing jobs. 

Say there are 48 cores and three graphs... They will allocate different number of cores due to their size/budget. For instance 24 cores for the first one, 16 cores for the second and 8 cores for the other. So instead of maximizing indexing throughput for any of these graphs, each of them will get ~constant build times.

jbellis:
Yes.

PR #162, opened 2023-12-01
Add optional FJP args for indexing and quantization

mdogan:
By default, SIMD operations are using `PhysicalCoreExecutor` and non-SIMD operations are using `FJP.commonPool()`.

With this addition, one can define custom fork-join pools for these tasks in `GraphIndexBuilder`, `ProductQuantization` and `BinaryQuantization`.


jbellis:
Thanks, looks reasonable in principle.  I'll ask @jkni to review.

In the meantime, would you mind signing the CLA at https://cla.datastax.com/ ?

PR #169, opened 2023-12-15
Call threadLocal.remove() for ThreadPooling#onClosed

michaeljmarshall:
There is an apparent leak in the `ThreadPooling` class. We pass a reference to the `ThreadPooling` class into the `ThreadLocal`'s value, which creates a strong reference to the class, which in turn creates a strong reference to the `threadLocal` object and therefore prevents the class from being garbage collected.

Assuming my understanding of references is correct (which is not a given), we just need to call `threadLocal.remove()`.


michaeljmarshall:
@jkni - after reviewing the class more, I agree with your conclusion. Thanks for taking a look. This is ready for re-review.

PR #172, opened 2023-12-21
Parallelize GraphIndexBuilder.removeDeletedNodes()

mdogan:
`GraphIndexBuilder.addNNDescentConnections()` was being called serially for each effected node during cleanup after some nodes are marked as deleted.

When there are too many (like hundreds of thousands), `GraphIndexBuilder.cleanup()` was taking too much time, and most of it was in `removeDeletedNodes()` while calling `addNNDescentConnections()` in loop.

With this change `addNNDescentConnections()` is called in parallel using SIMD FJP.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -465,9 +465,7 @@ public class GraphIndexBuilder<T> {
         }
 
         // repair affected nodes
-        for (var node : affectedLiveNodes) {
-            addNNDescentConnections(node);
-        }
+        simdExecutor.submit(() -> affectedLiveNodes.stream().parallel().forEach(this::addNNDescentConnections)).join();
 
         // reset deleted collection
         deletedNodes.clear();


mdogan:
I did a simple test with 1M siftdata dataset:

```java
        var vectors = SiftLoader.readFvecs("./sift/sift_base.fvecs");
        var ravv = new ListRandomAccessVectorValues(vectors, 128);

        var builder = newIndexBuilder(ravv);

        var start = System.nanoTime();
        builder.build();
        System.out.println("Build took " + TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start) + " ms.");

        start = System.nanoTime();
        for (int i = 0; i < vectors.size() / 5; i++) {
            builder.markNodeDeleted(i);
        }
        System.out.println("Remove took " + TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start) + " ms.");

        start = System.nanoTime();
        builder.cleanup();
        System.out.println("Cleanup took " + TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start) + " ms.");
    }
```

Before:
```
Build took 74131 ms.
Remove took 3 ms.
removeDeletedNodes() took 198853 ms.
Cleanup took 201097 ms.
```

After:
```
Build took 75606 ms.
Remove took 3 ms.
removeDeletedNodes() took 45728 ms.
Cleanup took 48104 ms.
```

jbellis:
Can you also test switching the "remove deleted nodes from neighbor lists" loop to be parallel?

mdogan:
I ran the same test again. That part is already fast when compared to remaining. 

Serial loop took `296 ms`,  parallel one `125 ms`. 

Only change was making `affectedLiveNodes` set thread-safe (`Collections.<Integer>newSetFromMap(new ConcurrentHashMap<>())`) and replacing loop with `Arrays.stream(liveNodes).parallel().forEach(..)`.

Not sure if it's worth it.

jbellis:
Makes sense, thanks for trying it out.

PR #175, opened 2023-12-25
Make GraphIndexBuilder.markNodeDeleted thread-safe

### LINKED ISSUE #174: Shouldn't GraphIndexBuilder.markNodeDeleted be thread-safe?
As far as I understand, `GraphIndexBuilder.markNodeDeleted` uses `GrowableBitSet.set()` in `OnHeapGraphIndex`. `GrowableBitSet` uses `java.util.BitSet` as the backend, which is not thread-safe.

mdogan:
`GraphIndexBuilder.markNodeDeleted` uses `GrowableBitSet` as the backing storage to track deleted nodes. But since `GrowableBitSet` itself is not thread-safe, `markNodeDeleted` is not thread-safe either.

Added a synchronized version of `GrowableBitSet` and used it as the storage for deleted nodes.

_Currently `SynchronizedGrowableBitSet` uses a single lock to guard all access to `BitSet`. If more concurrency needed while deleting nodes, a lock-free & atomic version can be introduced._

Closes #174

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/util/SynchronizedGrowableBitSet.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/util/SynchronizedGrowableBitSet.java
@@ -1,0 +1,169 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/util/SynchronizedGrowableBitSet.java
```diff
@@ -0,0 +1,169 @@
...
+ * does so, it will grow its internal storage multiplicatively, assuming that more growth will be
+ * needed in the future. This is the important difference from FixedBitSet + FBS.ensureCapacity,
+ * which grows the minimum necessary each time.
+ *
+ * @see GrowableBitSet
+ */
+public class SynchronizedGrowableBitSet extends BitSet {
+
+  private final java.util.BitSet bitSet;
+  private final Lock lock = new ReentrantLock();
```
mdogan, line 33: I didn't use `synchronized` blocks, because when virtual threads are used, a `synchronized` block pins virtual thread to platform (OS) thread. Locks are virtual thread friendly.

jbellis: Good thinking!

PR #177, opened 2024-01-02
Fix DenseIntMap size

mdogan:
`DenseIntMap.put()` was incrementing `size` field regardless of the operation is an insertion or an update.

Now it distinguishes inserts and updates by checking the old value.


jbellis:
I don't think we ever modify the contents of a DIM but you're right, at the class level this is a bug.  Good fix.

PR #181, opened 2024-01-08
Handle the case of removing all nodes in the cleanup

mdumandag:
We might mark all nodes of the graph as deleted before calling cleanup.

Doing so causes graph's entry node to become unset(-1) after the `removeDeletedNodes` call in the `cleanup` method.

The rest of the cleanup logic expects a valid entry node for the graph to function correctly. Therefore, I added an early return to handle such a case.


mdumandag:
Sure, I have signed it

jbellis (approved):

PR #182, opened 2024-01-08
Cleanup for 3.0 release

jbellis:


jbellis:
ugh, i mixed in the similarityFloor commit here some how.  fixing

jbellis:
... okay, that should be better now

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/GraphIndexTestCase.java
```diff
@@ -108,7 +108,7 @@ public void testSearchWithAcceptOrds() {
         similarityFunction = VectorSimilarityFunction.DOT_PRODUCT;
         VectorEncoding vectorEncoding = getVectorEncoding();
         GraphIndexBuilder<T> builder =
-                new GraphIndexBuilder<>(vectors, vectorEncoding, similarityFunction, 16, 100, 1.0f, 1.4f);
```
jkni, line 111: @jbellis  I'm not sure it matters, but was it intentional to quadruple this and the following change instead of double?

jbellis: Nope!

PR #185, opened 2024-01-10
Add GraphSearcher::resume

jbellis:


jbellis:
Good point, how about this?

PR #187, opened 2024-01-12
Add HAVERSINE similarity function

michaeljmarshall:
We can natively support geo-search where a geo point is represented as a 2d vector `[latitude, longitude]` and the similarity function is based on the haversine formula. The provided implementation copies the haversine function's implementation from Apache Lucene.

The primary benefits of using haversine instead of euclidean include:
* Correct behavior near poles and near boundaries without any requirement to amplify the search radius to ensure complete results.
* Correct ordering behavior. Because euclidean doesn't account for the map's distortion, it can produce incorrect orderings of locations. For example, given a search point `[45,0]`, the `[46,1]` is closer than `[44,1]` when using haversine, but they are the same when using euclidean.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/util/SloppyMath.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/util/SloppyMath.java
@@ -1,4 +1,12 @@
-/* some code derived from jodk: http://code.google.com/p/jodk/ (apache 2.0)
+/*
+ * All changes to the original code are Copyright DataStax, Inc.
+ *
+ * Please see the included license file for details.
+ */
+
+/*
+ * Original license:
+ * some code derived from jodk: http://code.google.com/p/jodk/ (apache 2.0)
  * asin() derived from fdlibm: http://www.netlib.org/fdlibm/e_asin.c (public domain):
  * =============================================================================
  * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
@@ -11,6 +19,8 @@
  */
 
 package io.github.jbellis.jvector.util;
+
+import io.github.jbellis.jvector.vector.VectorUtil;
 
 /** Math functions that trade off accuracy for speed. */
 public class SloppyMath {
@@ -29,6 +39,8 @@ public class SloppyMath {
      * @return distance in meters.
      */
     public static double haversinMeters(double lat1, double lon1, double lat2, double lon2) {
+        VectorUtil.verifyLatitudeAndLongitudeBoundaries(lat1, lon1);
+        VectorUtil.verifyLatitudeAndLongitudeBoundaries(lat2, lon2);
         return haversinMeters(haversinSortKey(lat1, lon1, lat2, lon2));
     }
 
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/util/SloppyMath.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/util/SloppyMath.java
@@ -1,0 +1,276 @@
+/* some code derived from jodk: http://code.google.com/p/jodk/ (apache 2.0)
+ * asin() derived from fdlibm: http://www.netlib.org/fdlibm/e_asin.c (public domain):
+ * =============================================================================
+ * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+ *
+ * Developed at SunSoft, a Sun Microsystems, Inc. business.
+ * Permission to use, copy, modify, and distribute this
+ * software is freely granted, provided that this notice
+ * is preserved.
+ * =============================================================================
+ */
+
+package io.github.jbellis.jvector.util;
+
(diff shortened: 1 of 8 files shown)


jvector-tests/src/test/java/io/github/jbellis/jvector/graph/Test2DThreshold.java
```diff
@@ -84,19 +84,71 @@ public void testThreshold() throws IOException {
 
     }
 
+    @Test
+    public void testThresholdHaversine() throws IOException {
```
michaeljmarshall, line 91: This test is failing. Not sure why yet.

michaeljmarshall: It was failing because I used the `biasedRandomFloat` method, which generated many vectors near the boundaries.

michaeljmarshall:
That would be interesting. How hard is it to wire up a Bench mark test? In order to do a correct comparison, we'll need to amplify the euclidean search radius because euclidean doesn't directly match. Perhaps we could also do a benchmark on 2D euclidean and harversine similarity functions to get a lower level result.

jvector-base/src/main/java/io/github/jbellis/jvector/util/SloppyMath.java
```diff
@@ -0,0 +1,288 @@
...
+     * <p>Error is at most 4E-1 (40cm) from the actual haversine distance, but is typically much
+     * smaller for reasonable distances: around 1E-5 (0.01mm) for distances less than 1000km.
+     *
+     * @param lat1 Latitude of the first point.
+     * @param lon1 Longitude of the first point.
+     * @param lat2 Latitude of the second point.
+     * @param lon2 Longitude of the second point.
+     * @return distance in meters.
+     */
+    public static double haversinMeters(double lat1, double lon1, double lat2, double lon2) {
```
jkni, line 41: I know we just inherited these, but do we want to update at least the exposed methods from `haversin` to `haversine`? I've never seen the `haversin` variant spelling.

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/Test2DThreshold.java
```diff
@@ -72,31 +72,93 @@ public void testThreshold() throws IOException {
...
                 assert result.getVisitedCount() < vectors.length : "visited all vectors for threshold " + tp.th;
+                assert result.getNodes().length >= 0.9 * tp.exactCount : "returned " + result.getNodes().length
+                        + " nodes for threshold " + tp.th + " but should have returned at least " + tp.exactCount
+                        + " visited " + result.getVisitedCount();
             }
         }
 
     }
 
+    @Test
```
jkni, line 90: I know I suggested this test, but now that it's working, I'm not sure what exactly this tells us. Since the same implementation is used for the creation of test params and the search/graph construction, this is largely the same test as `testThreshold`. Any issues with threshold will be cause by the existing test, and any issues with the similarity function will be obscured by using the same function to generate the ground truth test. Maybe a simple test against known values is more useful?

jbellis:
Should we close this out for now?

michaeljmarshall:
> This might be more worth than its work. For a point of reference, I implemented a quick JMH test to compare Euclidean/Haversine and saw a 35x performance difference. Given the amount of time spent in comparisons during a JVector query, this suggests a ~10x performance penalty for using Haversine over Euclidean. Is this around the ballpark you'd expect? Do you think Haversine still makes sense to support in this case, given the other improved qualities? This is a bit hard to quantify, but I suspect the answer is yes. We could tag it as `@Experimental` to make it clear that the value is not yet demonstrated.

Thanks for running that quick benchmark @jkni, and sorry to have left this PR dangling.

In order to evaluate whether it's worth introducing a haversine similarity function, it will be most meaningful to compare performance against the `GEO_DISTANCE` function in vsearch. Here are some reasons the 10x performance penalty might not be representative for geo:

* We expand the search radius when using euclidean based search to account for map distortion. We use `distanceInMeters / (Math.cos(Math.toRadians(lat)) * DISTANCE_PER_DEGREE_LONGITUDE_AT_EQUATOR)` to amplify the search radius. With haversine distance, we would not need to amplify the radius.
* Because of the amplified search radius, it is normal that we'll read rows from storage that are not in the search radius. (They are filtered because we use haversine distance as the post filter.)
* When a search crosses the anti-meridian, we query a euclidean similarity graph twice because it doesn't do the wrap around natively.

Finally, if we're interested in correct ANN ordering for nearest lat/lon neighbors, we need the graph to use haversine similarity.

> Should we close this out for now?

If I understand correctly, this is not a priority, so let's close it. I opened it because I was concerned that it would be too hard to add this later given how GEO_DISTANCE is implemented in the SAI code base. I took a closer look, and I see a way to retrofit the code later if/when this becomes a priority.

PR #190, opened 2024-01-22
Improve GraphIndexBuilder#cleanup doc for concurrent searches

### LINKED ISSUE #188: Possible race between GraphIndexBuilder.cleanup() and GraphSearcher
Is running `GraphIndexBuilder::cleanup` concurrently with `GraphSearcher::search` allowed when `GraphSearcher` is built with `GraphIndexBuilder::getGraph`? 

So usecase here is, some nodes are marked as deleted and then `GraphIndexBuilder::cleanup` is run to actually remove them, while a few `GraphSearcher`s are used to query the index.  See following code as an example;

```java
public static void main(String[] args) throws Exception {
    var baseVectors = SiftLoader.readFvecs("siftsmall/siftsmall_base.fvecs");
    var ravv = new ListRandomAccessVectorValues(baseVectors, 128);
    var builder = new GraphIndexBuilder<>(ravv, VectorEncoding.FLOAT32, VectorSimilarityFunction.COSINE, 32, 100, 1.5f, 1.5f);
    builder.build();

    // Delete nodes one by one
    Thread thread = new Thread(() -> {
        for (int i = 0; i < baseVectors.size(); i++) {
            builder.markNodeDeleted(i);
            builder.cleanup();
            LockSupport.parkNanos(TimeUnit.MILLISECONDS.toNanos(10));
        }
    });
    thread.setDaemon(true);
    thread.start();

    // Run search in a loop
    var searcher = new Builder<>(builder.getGraph().getView()).withConcurrentUpdates().build();
    while (true) {
        for (float[] queryVector : baseVectors) {
            var scoreFunction = new ExactScoreFunction() {
                @Override
                public float similarityTo(int nodeId) {
                    return VectorSimilarityFunction.COSINE.compare(queryVector, ravv.vectorValue(nodeId));
                }
            };
            searcher.search(scoreFunction, null, 100, Bits.ALL);
        }
    }
}
```

This code fails with `java.lang.NullPointerException: Cannot invoke "io.github.jbellis.jvector.graph.ConcurrentNeighborSet.iterator()" because "neighbors" is null`:

```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "io.github.jbellis.jvector.graph.ConcurrentNeighborSet.iterator()" because "neighbors" is null
	at io.github.jbellis.jvector.graph.OnHeapGraphIndex$ConcurrentGraphIndexView.getNeighborsIterator(OnHeapGraphIndex.java:272)
	at io.github.jbellis.jvector.graph.GraphSearcher.resume(GraphSearcher.java:321)
	at io.github.jbellis.jvector.graph.GraphSearcher.searchInternal(GraphSearcher.java:241)
	at io.github.jbellis.jvector.graph.GraphSearcher.search(GraphSearcher.java:144)
	at io.github.jbellis.jvector.graph.GraphSearcher.search(GraphSearcher.java:168)
	at io.github.jbellis.jvector.graph.GraphSearcher.search(GraphSearcher.java:189)
	at io.github.jbellis.jvector.example.Test.main(Test.java:44)

mdogan:
Added that `GraphIndexBuilder#cleanup` should not be called concurrently with searches on graph.

Closes #188

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -187,7 +187,8 @@ public class GraphIndexBuilder<T> {
      * <p>
      * Must be called before writing to disk.
      * <p>
-     * May be called multiple times, but should not be called during concurrent modifications to the graph.
+     * May be called multiple times, but should not be called during concurrent modifications to the graph
+     * or while executing concurrent searches on the graph.
      */
     public void cleanup() {
         if (graph.size() == 0) {


jbellis:
Thanks!

PR #193, opened 2024-01-25
pq supports off-heap memory to reduce GC and improve performance

xjtushilei:
Hi, I've added a feature about pq in off-heap.

In distributed systems implemented in Java, especially in vector scenarios, the memory is very limited, especially the memory in the Java heap, which may cause some gc problems. Therefore, we hope to put some objects into Java off-heap memory, so we first tried to modify it in PQ, and achieved good results.

As shown in the figure below, testing on the latest Dragonwell 11JDK (with java vector api), we found that the query performance can be greatly improved 26%, while the write performance has almost no change.
image

As shown in the figure below, on OpenJDK version 21.0.1, query performance is still greatly improved 23%.
image

In fact, what we initially focused on was not performance, but memory issues. Under the LSM-Tree architecture, a PQ or Graph object does not live long, but will cause a long GC. GC will cause writing or query glitches, so We want to put PQ outside the heap to reduce glitches. Not only can it be placed in off-heap memory after building, but it can also be placed directly in off-heap memory during the build process, which will greatly reduce the memory management pressure of the JVM.

The current implementation of Graph is relatively complex and does not have the ability to build outside the heap for the time being. It involves a lot of float and bytes conversion overhead. When I have time later, I will also try to test the performance of Graph implemented outside the heap in the Graph query phase.

Thanks for considering this PR. I'm open to any feedback or suggestions for improvements.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
@@ -19,23 +19,44 @@ package io.github.jbellis.jvector.pq;
 import io.github.jbellis.jvector.disk.RandomAccessReader;
 import io.github.jbellis.jvector.graph.NodeSimilarity;
 import io.github.jbellis.jvector.util.RamUsageEstimator;
+import io.github.jbellis.jvector.util.UnsafeUtils;
 import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
 
 import java.io.DataOutput;
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.Arrays;
 import java.util.Objects;
 
 public class PQVectors implements CompressedVectors {
     final ProductQuantization pq;
     private final byte[][] compressedVectors;
+    private final ByteBuffer compressedBuffer;
+    private final long cbOffHeapAddress;
+    private final int vectorsSize;
+    private final int vectorDim;
     private final ThreadLocal<float[]> partialSums; // for dot product, euclidean, and cosine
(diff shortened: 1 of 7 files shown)


jbellis:
Thanks for the patch!

We're working on a fairly large change to JVector including off-heap vectors in https://github.com/jbellis/jvector/pull/191.  This is based on the Java 21 MemorySegment API.  I think that is a better approach for the future than Unsafe.

xjtushilei:
Thank you, I get it.

jbellis:
Hi again,

191 is committed, can you take a look and see if it addresses your use case?

To use the Native (off-heap) vectors, pass `-Djvector.experimental.enable_native_vectorization=true` as a VM argument.

xjtushilei:
> Hi again,
> 
> 191 is committed to main, can you take a look and see if it addresses your use case?
> 
> To use the Native (off-heap) vectors, pass `-Djvector.experimental.enable_native_vectorization=true` as a VM argument.

Thank you for pinging me again. I took a cursory look at the code and it should be able to meet the needs. Unfortunately, our current jdk version does not support Java's FFM. I will try to use your new version after I upgrade the JDK when I have time.

In addition, I would like to confirm again that the data format (mainly write and load related functions) should be compatible between various versions, right? If I upgrade from a lower version of jvector to a higher version, I am worried that I will encounter format compatibility issues where historical data cannot be opened using the higher version of jvector. I can solve the compatibility of Java code interface by myself, but I am mainly worried about the compatibility of data format. If there is no problem, please close this PR.

jbellis:
Thanks.  Yes, the new code can write "fused graphs" in a new format, but is also compatible with the old format.

PR #198, opened 2024-02-01
Fixed name for s3 bucket in README

HanSolo:
Looks like the files are in s3://astra-vector/wikipedia_squad instead of s3://astra-vector/wikipedia_scout


jbellis:
Thanks!  Could you sign the DataStax CLA at https://cla.datastax.com/ ?

PR #200, opened 2024-02-01
Remove overzealous dependency exclusion affecting util-mmap

jkni:
This dependency exclusion was intended to prevent a conflict at some point. Whatever conflicting transitive dependency brought in Guava must have been removed, so util-mmap was failing at runtime due to missing dependencies.

Additional log message will help diagnose fallback causes in the future.


jbellis:
Why does current `main` work for me but not Gerrit?

jbellis (approved):

PR #201, opened 2024-02-01
Anisotropic PQ

jbellis:
Save/load not yet implemented.

On naming, I used "Anisotropic" and "Unweighted" which I don't love because it's two negatives, but I do think it's good for the unusual/complicated path to have an unusual/complicated name, so I prefer this over "Weighted" / "Isotropic."  Entirely possible that there's better options.


jvector-base/src/main/java/io/github/jbellis/jvector/pq/ProductQuantization.java
```diff
@@ -353,14 +551,24 @@ public static ProductQuantization load(RandomAccessReader in) throws IOException
             offset += size;
         }
 
+        float anisotropicThreshold;
+        if (version == 0) {
+            anisotropicThreshold = UNWEIGHTED;
+        } else {
+            // clunky, but doesn't seem worth making everyone writing a readFloat method for this
```
jkni, line 558: I'll come along and add this in a later PR, but we're going to end up wanting readFloat for LVQ

PR #202, opened 2024-02-02
Auto-boxing/unboxing reductions, reduce boxed footprint on heap

jkni:
Agrona data structures here are as good as any, a smaller dependency than some of the other specialized data structure libraries, and already part of the ecosystem of projects using JVector.

Benchmarking doesn't show any drawbacks of these changes.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/disk/GraphCache.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/disk/GraphCache.java
@@ -19,10 +19,9 @@ package io.github.jbellis.jvector.disk;
 import io.github.jbellis.jvector.graph.GraphIndex;
 import io.github.jbellis.jvector.util.Accountable;
 import io.github.jbellis.jvector.util.RamUsageEstimator;
+import org.agrona.collections.Int2ObjectHashMap;
 
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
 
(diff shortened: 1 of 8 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/disk/GraphCache.java
```diff
@@ -65,12 +64,12 @@ public long ramBytesUsed()
...
+        private final Int2ObjectHashMap<CachedNode> cache;
         private long ramBytesUsed = 0;
 
         public HMGraphCache(GraphIndex<float[]> graph, int distance) {
             try (var view = graph.getView()) {
-                HashMap<Integer, CachedNode> tmpCache = new HashMap<>();
+                var tmpCache = new Int2ObjectHashMap<CachedNode>();
                 cacheNeighborsOf(tmpCache, view, view.entryNode(), distance);
                 // Assigning to a final value ensure it is safely published
                 cache = tmpCache;
```
michaeljmarshall, line 75: It's probably a premature optimization, but when I looked at this the other day, I noticed that we could technically rebuild the map at this point with the correct size since we know `tmpCache`'s size, which we only get after walking the graph IIUC. Now that we're using the primitive map, I think it takes up more space to have unused elements, but I'm not sure how much it affects performance.

Probably not worth a change... but I wanted to mention it in case we find it takes up more space than necessary.

PR #204, opened 2024-02-04
Use ExplicitThreadLocal to avoid ThreadLocalMap hanging onto expensive objects forever via FJP

### LINKED ISSUE #189: Is GraphIndexBuilder leaking ThreadLocals ?
`GraphIndexBuilder` creates a few thread-local based `PoolingSupport` instances. And these thread-locals remain live as long as the FJP threads (common or custom) used in `GraphIndexBuilder` are alive. When many `GraphIndexBuilder` come and go during the lifecycle of the application, heap usage always grows. Following artificial code shows that:

```java
public static void main(String[] args) throws Exception {
    var baseVectors = SiftLoader.readFvecs("siftsmall/siftsmall_base.fvecs");
    var ravv = new ListRandomAccessVectorValues(baseVectors, 128);
    printMemory();

    for (int i = 0; i < 100; i++) {
        build(ravv);
        printMemory();
    }
}

private static void build(ListRandomAccessVectorValues ravv) {
    var builder = new GraphIndexBuilder<>(ravv, VectorEncoding.FLOAT32, ...);
    builder.build();
}

private static void printMemory() {
    System.gc();
    Runtime r = Runtime.getRuntime();
    System.out.println((r.totalMemory() - r.freeMemory()) / 1024 / 1024 + " MB");
}
```

Not sure how can we reset & reuse an existing `GraphIndexBuilder`.

jbellis:
fixed #189

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/util/ExplicitThreadLocal.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/util/ExplicitThreadLocal.java
@@ -1,0 +1,36 @@
+package io.github.jbellis.jvector.util;
+
+import org.agrona.collections.Long2ObjectHashMap;
+
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * A replacement for {@link ThreadLocal} that doesn't play clever tricks with the ThreadLocalMap,
+ * which means that we can guarantee that the values are GC'd once the parent instance is no
+ * longer referenced.
+ */
+public abstract class ExplicitThreadLocal<U> {
+    private final Long2ObjectHashMap<U> map = new Long2ObjectHashMap<>();
+
+    // computeIfAbsent wants a callable that takes a parameter, but if we use a lambda
+    // it will be a closure and we'll get a new instance for every call.  So we instantiate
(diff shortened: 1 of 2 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/util/ExplicitThreadLocal.java
```diff
@@ -0,0 +1,36 @@
...
+public abstract class ExplicitThreadLocal<U> {
+    private final Long2ObjectHashMap<U> map = new Long2ObjectHashMap<>();
+
+    // computeIfAbsent wants a callable that takes a parameter, but if we use a lambda
+    // it will be a closure and we'll get a new instance for every call.  So we instantiate
+    // it just once here as a field instead.
+    private final Function<Long, U> initialSupplier = k -> initialValue();
+
+    public U get() {
+        return map.computeIfAbsent(Thread.currentThread().getId(), initialSupplier);
```
mdogan, line 22: Is `Long2ObjectHashMap` safe to be modified by multiple threads?

jbellis:
Closing this in favor of https://github.com/jbellis/jvector/pull/206

PR #206, opened 2024-02-07
Attempt 3 at cleaning up the threadlocal leaks

jbellis:


jvector-base/src/main/java/io/github/jbellis/jvector/util/ExplicitThreadLocal.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * The standard {@link ThreadLocal} appears to be designed to be used with relatively
+ * short-lived Threads.  Specifically, it uses a ThreadLocalMap to store ThreadLocal key/value
+ * Entry objects, and there are no guarantees as to when Entry references are expunged unless
+ * you can explicitly call remove() on the ThreadLocal instance.  This means that objects
```
jkni, line 27: > This means that objects referenced by ThreadLocals will not be able to GC'd for the lifetime of the Thread. 

I don't think this is quite true. The inability to be GCed for the lifetime of the thread is only true if the object containing the ThreadLocal is circularly referenced by the value.

jbellis: My reasoning: TLM has a strong reference to its Entry[] and Entry has a strong reference to value.  Therefore you can't GC value while the Entry exists.

```
        static class Entry extends WeakReference<ThreadLocal<?>> {
            /** The value associated with this ThreadLocal. */
            Object value;

            Entry(ThreadLocal<?> k, Object v) {
                super(k);
                value = v;
            }
        }
```

PR #208, opened 2024-02-08
optimize reconnectOrphanedNodes

jbellis:
Stash the removed neighbors when we trim down to max degree so that we can use those as reconnection candidates if necessary.

(Michael's profile shows that the search is overwhelmingly the most expensive part.)

this also parallelizes the reconnection operation.

Applies on top of the ExplicitThreadLocal changes.


jbellis:
This overxomplicates things, I will update tomorrow morning.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -243,31 +245,57 @@ private void reconnectOrphanedNodes() {
             var gs = graphSearcher.get();
```
jkni: We should get gs/v1/v2 inside the parallel task, right? This doesn't seem right for the ETL

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -243,31 +245,57 @@ private void reconnectOrphanedNodes() {
...
             var v1 = vectors.get();
             var v2 = vectorsCopy.get();
-            var connectionTargets = new IntHashSet();
-            for (int node = 0; node < graph.getIdUpperBound(); node++) {
-                if (!connectedNodes.get(node) && graph.containsNode(node)) {
-                    // search for the closest neighbors
+            var connectionTargets = ConcurrentHashMap.<Integer>newKeySet();
+            parallelExecutor.submit(() -> IntStream.range(0, graph.getIdUpperBound()).parallel().forEach(node -> {
```
jkni: I'd expect this to be on the simdExecutor since the searches will perform significant SIMD work. What's the thinking for putting this on the parallelExecutor?

jkni: This is an example of a case where I'm somewhat concerned about the ETL on FJP. The cleanup method might be run many times over the lifetime of a GIB, and each thread that runs a task on the stream and unregisters from the FJP will leak an entry in the ETL maps. On the SIMD executor, this wouldn't be as big a deal, as we control the types of tasks submitted and know they're not participating in cooperative blocking. On the parallelExecutor default of the common pool, you could have many participating threads over the lifetime of even one IntStream.

jbellis: new approach means most of the time we won't have to do a search.  but i can switch it to simdE

jbellis: I thought FJP threads are
1. Created when we have more tasks than threads [up to the number of cores, on the common pool]
2. Killed when idle for more than a minute

That shouldn't result in "many participating threads over the lifetime of even one IntStream" unless by many you mean "up to the number of cores" in which case this is desired.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jbellis: yes, tx

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jkni: Discussed offline re: managed blockers and changes in idle killing behavior. We agree on behavior here and think this is a reasonable choice for us.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -240,34 +242,61 @@ private void reconnectOrphanedNodes() {
...
+                    toScratchCandidates(result.getNodes(), neighbors);
+                    searchPathNeighbors.put(node, neighbors);
                 }
-            }
+                connectToClosestNeighbor(node, neighbors, connectionTargets);
+            }));
             if (nReconnected.get() == 0) {
                 break;
             }
+            System.out.println("Pass " + i + " reconnected " + nReconnected.get() + " nodes");
```
michaeljmarshall, line 279: You probably want to remove this line.

PR #224, opened 2024-02-20
Improve performance of native vectorization provider

jkni:
* Optimize I/O by copying into backing buffer of MemorySegment and fixing endianness in place.
* Refactor reranking to take array of nodes to rerank, and rerank in native code for dot product/square distance.
* Use native code dot product/Euclidean for longer vectors, based on JMH benchmarking.

This PR further exposes the need for a FloatSequence type so that we can use the type system to highlight cases where we're looking at a float vector compared to any in-memory sequence of floats (which may, for example, be composed of floats representing multiple vectors).

This PR does not show any noticeable regressions in native vectorization performance. Performance for in-memory graphs/building PQ is marginally improved in some cases. The major improvement is in on-disk PQ searches that require reranking.

Baseline results prior to the PR
```text-embedding-3-large_1536_100000: 99744 base and 9232 query vectors created, dimensions 1536
Build M=32 ef=100 in 23.19s with avg degree 32.00 and 0.45 short edges
Uncompressed vectors
  Query (memory) top 100/1 recall 0.9693 in 3.85s after 25,097,256 nodes visited
  Query (disk) top 100/1 recall 0.9693 in 7.74s after 25,097,256 nodes visited
  Query (memory) top 100/2 recall 0.9885 in 7.04s after 42,628,406 nodes visited
  Query (disk) top 100/2 recall 0.9885 in 13.96s after 42,628,406 nodes visited
ProductQuantization(384,256) build in 75.96s,
ProductQuantization(384,256) encoded 99744 vectors [38.03 MB] in 4.59s
  Query top 100/1 recall 0.9123 in 7.01s after 25,284,668 nodes visited
  Query top 100/2 recall 0.9882 in 12.51s after 42,942,174 nodes visited
```

Results post-PR
```text-embedding-3-large_1536_100000: 99744 base and 9232 query vectors created, dimensions 1536
Build M=32 ef=100 in 23.71s with avg degree 32.00 and 0.45 short edges
Uncompressed vectors
  Query (memory) top 100/1 recall 0.9692 in 3.81s after 24,601,258 nodes visited
  Query (disk) top 100/1 recall 0.9692 in 7.69s after 24,601,258 nodes visited
  Query (memory) top 100/2 recall 0.9884 in 6.73s after 41,906,890 nodes visited
  Query (disk) top 100/2 recall 0.9884 in 13.53s after 41,906,890 nodes visited
ProductQuantization(384,256) build in 85.57s,
ProductQuantization(384,256) encoded 99744 vectors [38.03 MB] in 3.92s
  Query top 100/1 recall 0.9118 in 4.52s after 24,763,122 nodes visited
  Query top 100/2 recall 0.9883 in 6.69s after 42,198,490 nodes visited
```

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
@@ -143,7 +143,7 @@ public class NodeQueue {
                                     .filter(m -> decodeScore(m) >= rerankFloor)
                                     .mapToInt(this::decodeNodeId).toArray();
         var scores = vectorTypeSupport.createFloatVector(ids.length);
-        rr.similarityTo(ids, scores);
+        rr.score(ids, scores);
         var nodeScores = new SearchResult.NodeScore[ids.length];
         for (int i = 0; i < ids.length; i++) {
             nodeScores[i] = new SearchResult.NodeScore(ids[i], scores.get(i));
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeSimilarity.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeSimilarity.java
@@ -16,6 +16,7 @@
 
 package io.github.jbellis.jvector.graph;
 
+import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
 import io.github.jbellis.jvector.vector.VectorizationProvider;
 import io.github.jbellis.jvector.vector.types.VectorFloat;
 import io.github.jbellis.jvector.vector.types.VectorTypeSupport;
@@ -81,6 +82,28 @@ public interface NodeSimilarity {
          * @param results the similarity scores for each node, in the same order as the input nodes.
          *                This should be pre-allocated to the same size as nodes.length.
          */
-        void similarityTo(int[] nodes, VectorFloat<?> results);
+        void score(int[] nodes, VectorFloat<?> results);
+
+        /**
+         * Create a Reranker from a VectorSimilarityFunction, a query vector, and a graph view. This is a convenience
+         * method for the common case of scoring a set of nodes based on similarity to a query vector.
+         *
+         * @param queryVector the query vector
+         * @param vsf         the similarity function to use
+         * @param view        the graph view used to retrieve vectors from node ids
+         * @return
+         */
+        static Reranker from(VectorFloat<?> queryVector, VectorSimilarityFunction vsf, GraphIndex.View view) {
+            return (nodes, results) -> {
+                var nodeCount = nodes.length;
+                var dimension = queryVector.length();
+                var packedVectors = vectorTypeSupport.createFloatVector(nodeCount * dimension);
+                for (int i1 = 0; i1 < nodeCount; i1++) {
+                    var node = nodes[i1];
+                    view.getVectorInto(node, packedVectors, i1 * dimension);
+                }
+                vsf.compareMulti(queryVector, packedVectors, results);
+            };
+        }
     }
 }
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorSimilarityFunction.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorSimilarityFunction.java
@@ -41,8 +41,8 @@ public enum VectorSimilarityFunction {
     }
 
     @Override
-    public void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVector, VectorFloat<?> results) {
-      VectorUtil.squareDistanceMultiScore(v1, packedVector, results);
+    public void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVectors, VectorFloat<?> results) {
+      VectorUtil.euclideanMultiScore(v1, packedVectors, results);
     }
   },
 
@@ -60,8 +60,8 @@ public enum VectorSimilarityFunction {
     }
 
     @Override
-    public void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVector, VectorFloat<?> results) {
-      VectorUtil.dotProductMultiScore(v1, packedVector, results);
+    public void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVectors, VectorFloat<?> results) {
+      VectorUtil.dotProductMultiScore(v1, packedVectors, results);
     }
   },
 
@@ -78,8 +78,8 @@ public enum VectorSimilarityFunction {
     }
 
     @Override
-    public void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVector, VectorFloat<?> results) {
-      VectorUtil.cosineMultiScore(v1, packedVector, results);
+    public void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVectors, VectorFloat<?> results) {
+      VectorUtil.cosineMultiScore(v1, packedVectors, results);
     }
   };
 
@@ -98,7 +98,7 @@ public enum VectorSimilarityFunction {
    * similarity scores correspond to closer vectors.
    *
    * @param v1 a vector
-   * @param packedVector N vectors packed into a single vector, of N * v1.length() dimension
+   * @param packedVectors N vectors packed into a single vector, of N * v1.length() dimension
    */
-  public abstract void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVector, VectorFloat<?> results);
+  public abstract void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVectors, VectorFloat<?> results);
 }
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtil.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtil.java
@@ -169,13 +169,13 @@ public final class VectorUtil {
   }
 
   /**
-   * Calculates the square distance similarity scores between v1 and multiple vectors packed into v2.
+   * Calculates the Euclidean similarity scores between v1 and multiple vectors packed into v2.
    * Note that unlike the squareDistance, this method puts similarity scores into results, taking this responsibility from VectorSimilarityFunction.
    * @param v1 the query vector
    * @param v2 multiple vectors to compare against
    * @param results the output vector to store the similarity scores. This should be pre-allocated to the same size as the number of vectors in v2.
    */
-  public static void squareDistanceMultiScore(VectorFloat<?> v1, VectorFloat<?> v2, VectorFloat<?> results) {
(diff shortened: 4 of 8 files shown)


jbellis:
Why is cosine the odd one out?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
```diff
@@ -105,6 +105,14 @@ interface View extends AutoCloseable {
          */
         VectorFloat<?> getVector(int node);
 
+        /**
```
jbellis, line 108: should we remove getVector, or default it into calling getVectorInto?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeSimilarity.java
```diff
@@ -70,6 +74,13 @@ default boolean isExact() {
...
     interface Reranker {
-        float similarityTo(int node2);
+        /**
+         * Rerank the nodes identified by the nodes array, storing the similarities in the results vector.
+         *
+         * @param nodes   the nodes to calculate similarity for
+         * @param results the similarity scores for each node, in the same order as the input nodes.
+         *                This should be pre-allocated to the same size as nodes.length.
+         */
+        void similarityTo(int[] nodes, VectorFloat<?> results);
```
jbellis: let's rename this, the original name implies in my mind that it's returning the similarity

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
```diff
@@ -134,12 +137,18 @@ public int[] nodesCopy() {
...
+    public SearchResult.NodeScore[] nodesCopy(NodeSimilarity.Reranker rr, float rerankFloor) {
+        var ids = IntStream.range(0, size())
+                                    .mapToLong(i -> heap.get(i + 1))
+                                    .filter(m -> decodeScore(m) >= rerankFloor)
+                                    .mapToInt(this::decodeNodeId).toArray();
+        var scores = vectorTypeSupport.createFloatVector(ids.length);
+        rr.similarityTo(ids, scores);
+        var nodeScores = new SearchResult.NodeScore[ids.length];
+        for (int i = 0; i < ids.length; i++) {
+            nodeScores[i] = new SearchResult.NodeScore(ids[i], scores.get(i));
```
jbellis, line 149: I was never happy about reallocating to a NodeScore[] and now it's even more obviously extra work for no benefit

How about this?

SearchResult adds getNode(i), getScore(i) methods.  Internal storage is int[] and vector[]

Move nodesCopy logic and GS.extractScores into an SR static method returning SR

(followup ticket is fine)

jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/VectorTypeSupport.java
```diff
@@ -37,7 +37,31 @@ public interface VectorTypeSupport {
...
+
+    /**
+     * Read a vector from the given RandomAccessReader and store it in the given vector at the specified offset.
+     * @param r the reader to read the vector from.
+     * @param size the size of the vector to read.
+     * @param vector the vector to store the read data in.
+     * @param offset the offset in the vector to store the read data at.
+     * @throws IOException
+     */
+    void readFloatVector(RandomAccessReader r, int size, VectorFloat<?> vector, int offset) throws IOException;
```
jbellis, line 57: How ugly does it make things if we move the read methods into RAR?  I think the corresponding write method used to live in a utility class?

I'd like VTS to be purely about the different vector abstractions, "read bytes into a vector" shouldn't have to live here

(that's what you get adding javadoc so this shows up on my radar :)

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtil.java
```diff
@@ -47,7 +47,7 @@ public static float dotProduct(VectorFloat<?> a, VectorFloat<?> b) {
       throw new IllegalArgumentException("vector dimensions differ: " + a.length() + "!=" + b.length());
     }
     float r = impl.dotProduct(a, b);
-    assert Float.isFinite(r);
+    assert Float.isFinite(r) : String.format("dotProduct(%s, %s) = %s", a, b, r);
```
jbellis, line 50: why doesn't this method just invoke the other with appropriate offset and length?

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtil.java
```diff
@@ -156,4 +156,37 @@ public static int hammingDistance(long[] v1, long[] v2) {
...
+  }
+
+  /**
+   * Calculates the square distance similarity scores between v1 and multiple vectors packed into v2.
+   * Note that unlike the squareDistance, this method puts similarity scores into results, taking this responsibility from VectorSimilarityFunction.
+   * @param v1 the query vector
+   * @param v2 multiple vectors to compare against
+   * @param results the output vector to store the similarity scores. This should be pre-allocated to the same size as the number of vectors in v2.
+   */
+  public static void squareDistanceMultiScore(VectorFloat<?> v1, VectorFloat<?> v2, VectorFloat<?> results) {
```
jbellis: euclideanMultiScore for consistency?

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorSimilarityFunction.java
```diff
@@ -77,4 +92,13 @@ public float compare(VectorFloat<?> v1, VectorFloat<?> v2) {
    * @return the value of the similarity function applied to the two vectors
    */
   public abstract float compare(VectorFloat<?> v1, VectorFloat<?> v2);
```
jbellis, line 94: I think we can improve the parameter names here

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorSimilarityFunction.java
```diff
@@ -77,4 +92,13 @@ public float compare(VectorFloat<?> v1, VectorFloat<?> v2) {
...
+
+  /**
+   * Calculates similarity scores between a query vector and multiple vectors with a specified function. Higher
+   * similarity scores correspond to closer vectors.
+   *
+   * @param v1 a vector
+   * @param packedVector N vectors packed into a single vector, of N * v1.length() dimension
+   */
+  public abstract void compareMulti(VectorFloat<?> v1, VectorFloat<?> packedVector, VectorFloat<?> results);
```
jbellis: packedVectors [plural] ?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/SiftSmall.java
```diff
@@ -101,7 +105,15 @@ private static void testRecallInternal(GraphIndex graph, RandomAccessVectorValue
             }
             else {
                 NodeSimilarity.ApproximateScoreFunction sf = compressedVectors.approximateScoreFunctionFor(queryVector, VectorSimilarityFunction.EUCLIDEAN);
-                NodeSimilarity.Reranker rr = (j) -> VectorSimilarityFunction.EUCLIDEAN.compare(queryVector, view.getVector(j));
+                NodeSimilarity.Reranker rr = (nodes, results) -> {
+                    var nodeCount = nodes.length;
```
jbellis: this is kind of a lot to expect the average user to implement, can we make it reusable somehow?

jbellis:
[end of comments]

jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/VectorTypeSupport.java
jbellis: (if you agree, this would also be fine to move to a followup ticket)

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
jkni: Good question that I also considered. I think removing `getVector` would make sense.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
jkni: That makes sense. It has always bugged me as well. I'll do that in a followup.

jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/VectorTypeSupport.java
jkni: Somewhat; the tricky party is that the right implementation depends on the type of backing storage we have for the vector, so we need a not-as-ugly way to expose that to the RAR. Let me try in a follow up ticket.

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtil.java
jkni: That's a good question. I don't have any answer in particular; this is the pattern we've used for all VectorUtil  methods with offset/non-offset versions. In a few cases, it marginally reduces the number of operations around loop bounds but I'm not sure whether it's worth the maintenance cost.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/SiftSmall.java
jkni: Yes, let me take a pass at it. Believe-it-or-not, this is already the slimmed down version, but I have a couple ideas to make it easier.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
jkni: I think I'd like to defer this to a follow up PR. The remaining dependency on `getVector` is in the caching graphs, which might remove this dependency entirely. If not, there's some work to thread through the vector dimensionality so that all callers (or the View itself) can call `getVectorInto` with an appropriately sized vector.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeSimilarity.java
```diff
@@ -81,6 +82,28 @@ interface Reranker {
...
+        /**
+         * Create a Reranker from a VectorSimilarityFunction, a query vector, and a graph view. This is a convenience
+         * method for the common case of scoring a set of nodes based on similarity to a query vector.
+         *
+         * @param queryVector the query vector
+         * @param vsf         the similarity function to use
+         * @param view        the graph view used to retrieve vectors from node ids
+         * @return
+         */
+        static Reranker from(VectorFloat<?> queryVector, VectorSimilarityFunction vsf, GraphIndex.View view) {
```
jbellis, line 96: perfect

jbellis (approved):
LGTM.

Can you create issues for the followup items so they don't get lost?

PR #229, opened 2024-02-22
Improve quality of bindings with newer jextract version

jkni:
This regenerates our bindings using the latest version of jextract. This shrinks the quantity of generated code and makes it cleaner to interact with/a better fit for the recent improvements to the FFM API.


jbellis (approved):
I'm going to continue to trust that you know what you're doing here.

jvector-native/src/main/c/jextract_vector_simd.sh
```diff
@@ -36,13 +36,9 @@ then
...
+jextract \
   --output ../java \
   -t io.github.jbellis.jvector.vector.cnative \
   -I . \
-  -l jvector \
   --header-class-name NativeSimdOps \
   jvector_simd.h
-
-# Use sed to strip System.loadLibrary("jvector"); from ../java/io/github/jbellis/jvector/vector/cnative/RuntimeHelper.java
-sed -i 's/.*System.loadLibrary("jvector");//' ../java/io/github/jbellis/jvector/vector/cnative/RuntimeHelper.java
```
tjake, line 48: How does this work now?

jkni: We have a LibraryLoader helper class to handle this now. The jextract-generated version doesn't handle unpacking the libraries from a jar when added as a dependency.

PR #240, opened 2024-03-05
Fix Bench timings on paths performing multiple types of queries

jkni:
AFAICT, these timings have been wrong since we first started doing on-heap/on-disk exact searches in the same overquery loop.

Fortunately, the benchmarking of fused graphs was comparing 32-cluster fused against 256-cluster traditional, which were in different compressor settings, so the timings were still accurate. I'm not sure that we made any impactful decisions based on on-disk/on-memory comparison for exact searches.


jbellis:
LGTM, and I think you're right.  Good catch

jbellis (approved):

PR #244, opened 2024-03-08
Build indexes using compressed vectors

jbellis:
Changes to Bench are quick-and-dirty and not intended to be committed as is.

BuildScoreProvider encapsulates build-time similarity APIs.  The main changes to the actual logic are in ConcurrentNeighborSet to perform two-pass diversity checks to reduce the full-precision comparisons required.

Note that
```
     * While GraphIndexBuilder may use approximate scoring to find candidate neighbors, we
     * always rerank them using exact scoring before storing them in the neighbor set.
```

This is important for performance (2x neighbors = 2x slower searches) and it means we don't need to do crazy juggling around "now that i'm writing to disk i'm going to recompute all the scores."


jvector-base/src/main/java/io/github/jbellis/jvector/disk/CachingGraphIndex.java
```diff
@@ -93,6 +94,21 @@ public NodesIterator getNeighborsIterator(int node) {
...
+        }
+
+        @Override
+        public boolean isValueShared() {
+            return false;
+        }
+
+        @Override
+        public RandomAccessVectorValues copy() {
+            return null;
```
jkni: Returning `null` here seems like the most fragile part of ViewWithVectors. Is there anything useful to do with the definition of `copy` to make this behavior make more sense? I'd be surprised for `copy` to return null.

jkni: I also wonder why CachingADCGraphIndex view's copy returns `this`, but this returns `null`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
```diff
@@ -119,10 +125,10 @@ public void replaceDeletedNeighbors(Bits deletedNodes, NodeArray candidates) {
                 }
             }
 
-            // merge the live neighbors with the candidates and prune down to diverse edges
-            var merged = mergeNeighbors(liveNeighbors, candidates);
-            var newNeighbors = copyDiverse(merged, selectDiverse(merged, 0));
-            return new Neighbors(newNeighbors, newNeighbors.size);
+            // merge the remaining neighbors with the candidates
+            NodeArray merged = rescoreAndMerge(liveNeighbors, candidates);
```
jkni, line 129: Aren't liveNeighbors and candidates always exact here, even if the BuildScoreProvider is approximate?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -641,19 +630,22 @@ private static NodeArray toScratchCandidates(SearchResult.NodeScore[] candidates
...
             if (n == newNode) {
                 continue;
             }
-            scratch.insertSorted(n, similarityFunction.compare(values.vectorValue(newNode), valuesCopy.vectorValue(n)));
+            scratch.insertSorted(n, scoreFunction.similarityTo(n));
         }
         return scratch;
     }
 
+    public void setScoreProvider(BuildScoreProvider scoreProvider) {
```
jkni: This will need some work when actually adopted - the neighbor factory captures the original score provider.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -600,36 +574,51 @@ private int approximateMedioid() {
...
-            var centroid = vectorTypeSupport.createFloatVector(dimension);
-            for (var it = graph.getNodes(); it.hasNext(); ) {
-                var node = it.nextInt();
-                VectorUtil.addInPlace(centroid, vc.vectorValue(node));
+        var centroid = scoreProvider.approximateCentroid();
+        // if the centroid is the zero vector, pick a random node
+        // (this is not scenario likely to arise outside of small, contrived tests)
+        if (dotProduct(centroid, centroid) < 1E-6) {
+            while (true) {
+                int randomNode = ThreadLocalRandom.current().nextInt(graph.getIdUpperBound());
```
jkni: Technically this could be an infinite loop because we aren't guaranteed to have a non-deleted node in the graph at this point

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @return true if the primary score functions used for construction are exact.  This
+     * is modestly redundant, but it saves having to allocate new Search/Diversity provider
+     * objects in some hot construction loops.
+     */
+    boolean isExact();
+
+    /**
+     * @return the approximate centroid of the known nodes.  This is called every time the graph
```
jkni: This isn't a quite complete description of when this can be called, since it will also be called if the entry point is deleted

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
```diff
@@ -0,0 +1,214 @@
...
+     * separately by the ConcurrentNeighborSet diversity code.)
+     * <p>
+     * @param node the graph node to provide similarity scores against
+     */
+    SearchScoreProvider searchProviderFor(int node);
+
+    /**
+     * Create a diversity score provider to use internally during construction.
+     * <p>
+     * The difference between the diversity provider and the search provider is
```
jkni, line 75: Should we make it clear that the diversity and search providers should use the same approximation (to be able to merge)?

jvector-base/src/main/java/io/github/jbellis/jvector/disk/CachingGraphIndex.java
jbellis: I think something in my refactoring resulted in IntelliJ generating defaults there.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
jbellis: yes, I went ahead and changed this code in advance of implementing

                // TODO support approximate scoring here

in removeDeletedNodes, because I thought it was likely I'd forget

jbellis: ... actually now that we are using the FreshDiskANN algorithm I don't think it makes sense to use approximate scoring at all.  will update

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jbellis: oh man debugging this would have driven me crazy, nice catch

jbellis: the fix is ugly, we don't need this (changing the provider mid-build) for SAI, and we might not for Mimir.  i'll take it out

jkni: I was wondering about the CNS changes too and assumed you already had a plan there! WFM.

jbellis:
Oops, I thought I did pull, sorry.

PR #246, opened 2024-03-12
Use on-heap MemorySegments for native vectors/sequences

jkni:
As discussed, the current Arena.ofAuto usage for off-heap memory segments is not sufficient for production workloads. While it may be worthwhile to continue to invest in off-heap options, this switch to on-heap memory segments increases stability at very little cost. It continues to allow the use of native code, as long as that code doesn't upcall back into the JVM.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/disk/RandomAccessReader.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/disk/RandomAccessReader.java
@@ -32,11 +32,15 @@ public interface RandomAccessReader extends AutoCloseable {
 
     void readFully(ByteBuffer buffer) throws IOException;
 
-    void readFully(float[] floats) throws IOException;
+    default void readFully(float[] floats) throws IOException {
+        read(floats, 0, floats.length);
+    }
 
     void readFully(long[] vector) throws IOException;
 
     void read(int[] ints, int offset, int count) throws IOException;
+
+    void read(float[] floats, int offset, int count) throws IOException;
 
     void close() throws IOException;
 }
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorProvider.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorProvider.java
@@ -54,9 +54,7 @@ final class ArrayVectorProvider implements VectorTypeSupport
 
     @Override
     public void readFloatVector(RandomAccessReader r, int size, VectorFloat<?> vector, int offset) throws IOException {
-        float[] v = new float[size];
-        r.readFully(v);
-        System.arraycopy(v, 0, ((ArrayVectorFloat)vector).get(), offset, size);
+        r.read(((ArrayVectorFloat) vector).get(), offset, size);
     }
 
     @Override
--- a/jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentByteSequence.java
+++ b/jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentByteSequence.java
@@ -18,32 +18,31 @@ package io.github.jbellis.jvector.vector;
 
 import io.github.jbellis.jvector.vector.types.ByteSequence;
 
-import java.lang.foreign.Arena;
 import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySegment;
 import java.lang.foreign.ValueLayout;
 import java.nio.Buffer;
 
 /**
- * ByteSequence implementation backed by an off-heap MemorySegment.
+ * ByteSequence implementation backed by an on-heap MemorySegment.
  */
-public class OffHeapByteSequence implements ByteSequence<MemorySegment> {
+public class MemorySegmentByteSequence implements ByteSequence<MemorySegment> {
     private final MemorySegment segment;
     private final int length;
 
-    OffHeapByteSequence(int length) {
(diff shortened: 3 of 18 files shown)


jbellis:
Nit: OffHeapVectorFloat name is no longer accurate

More substantive: can we give people a way to manually construct off-heap vectors as in https://github.com/jbellis/jvector/pull/193 ?

jbellis:
Could we make it so you could instantiate your preferred VTS manually?  Then you wouldn't need to clutter the master api with extra parameters

jbellis:
I was suggesting the former because I didn't know better, the latter SGTM :)

jvector-base/src/main/java/io/github/jbellis/jvector/disk/RandomAccessReader.java
```diff
@@ -38,5 +38,7 @@ public interface RandomAccessReader extends AutoCloseable {
 
     void read(int[] ints, int offset, int count) throws IOException;
 
+    void read(float[] floats, int offset, int count) throws IOException;
```
jbellis, line 43: make the existing reafFully(float[]) default to calling this?

jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorProvider.java
```diff
@@ -38,6 +38,11 @@ public VectorFloat<?> createFloatVector(Object data)
         return new ArrayVectorFloat((float[]) data);
     }
 
+    /**
```
jbellis: why copy the super javadoc in this one case?

jvector-base/src/main/java/io/github/jbellis/jvector/vector/DefaultVectorizationProvider.java
```diff
@@ -43,8 +44,13 @@ public VectorUtilSupport getVectorUtilSupport() {
     return vectorUtilSupport;
   }
 
+  /**
+   * Returns the default vector type support. Hint is not respected.
```
jbellis: "Hint is not respected, because the Default provider only supports on-heap allocations."

jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentByteSequence.java
```diff
@@ -27,23 +28,26 @@
 /**
  * ByteSequence implementation backed by an off-heap MemorySegment.
```
jbellis: "backed by an on- or off-heap MemorySegment"

jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentVectorFloat.java
```diff
@@ -16,41 +16,41 @@
...
 import io.github.jbellis.jvector.vector.types.VectorFloat;
 
 import java.lang.foreign.Arena;
-import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySegment;
 import java.lang.foreign.ValueLayout;
 import java.nio.Buffer;
 
 /**
  * VectorFloat implementation backed by an off-heap MemorySegment.
```
jbellis: on- or off-

jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentVectorProvider.java
```diff
@@ -0,0 +1,147 @@
...
+    @Override
+    public VectorFloat<?> createFloatVector(Object data)
+    {
+        if (data instanceof Buffer)
+            return new MemorySegmentVectorFloat((Buffer) data, hint);
+
+        return new MemorySegmentVectorFloat((float[]) data);
+    }
+
+    /**
```
jbellis: gratuitous javadoc again

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorizationProvider.java
```diff
@@ -42,8 +43,13 @@ public VectorUtilSupport getVectorUtilSupport() {
         return vectorUtilSupport;
     }
 
+    /**
```
jbellis: "Hint is ignored, because only on-heap vectors are supported."

jbellis (approved):
LGTM

jbellis:
... do we expect people to want to mix on- and off-heap vectors?

The core problem is that the AUTO MemorySegment allocation is not production quality, so the problem we need to solve is allowing someone to wire up an off-heap VectorProvider that does something smarter.  I'm not sure how StorageHint addresses that.

jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorProvider.java
jkni: Good catch. In an earlier version of the patch, the Javadoc was distinct from super. When I rearranged, that was no longer the case, but I only removed the changes rather than the whole Javadoc.

jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentVectorProvider.java
jkni: Same reasoning as above. Thanks for the catch here.

jbellis:
SGTM

PR #248, opened 2024-03-13
Concurrent deletes

jbellis:


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -435,93 +448,125 @@ public void markNodeDeleted(int node) {
...
+        // but if the entry point was deleted then we have no choice
+        if (toDelete.get(graph.entry())) {
+            updateEntryPoint();
         }
 
-        // repair affected nodes
-        simdExecutor.submit(() -> affectedLiveNodes.stream().parallel().forEach(this::addNNDescentConnections)).join();
+        // wait for in-progress inserts to complete before removing the deleted nodes,
+        // so that we don't invalidate the candidate queues of their searches
+        busyWaitForInProgressInserts();
```
jkni: I don't think this is safe enough for concurrent searches that aren't driven by inserts. I modified testConcurrentDeletes to sometimes perform searches, and it will occasionally fail on invalid edges.

PR #257, opened 2024-03-20
8-bit Turbo LVQ for reranking

jkni:
Implement single-level 8-bit Turbo LVQ prototype. This includes on-disk format and dot product support. This needs more test coverage, support for Euclidean/cosine, some refactoring of graphs to make the repetition less burdensome, and reworking lvqDot to not assume AVX-512.

No need for review as-is. I will push to this branch as it progresses.


jbellis:
Depending on the level of difficulty of implementing alternatives, I'm okay with specializing to avx512.

PR #266, opened 2024-04-03
make Test2DThreshold less fragile

jbellis:


jbellis:
the most meaningful change is the maxDegree of the graphs being constructed

I noticed that sometimes the graphs were not fully connected so I increased the reconnect tries to 5

then I reduced the threshold for success to 85% for good measure.  this was not necessary to pass 1k iterations for me but for some reason (possibly more cores = more concurrent candidates being considered) it is less fragile for me than for CI.

PR #268, opened 2024-04-04
Fix packing of LVQ vector dimensionalities not divisible by 64. Introduce test for LVQ similarity.

jkni:
Bug in LVQ packing was introduced during an earlier refactor. 50000 runs of the LVQ test passed with this tolerance, so I don't think it should be noisy.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/LocallyAdaptiveVectorQuantization.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/LocallyAdaptiveVectorQuantization.java
@@ -17,8 +17,8 @@
 package io.github.jbellis.jvector.pq;
 
 import io.github.jbellis.jvector.disk.RandomAccessReader;
-import io.github.jbellis.jvector.graph.disk.LVQView;
 import io.github.jbellis.jvector.graph.RandomAccessVectorValues;
+import io.github.jbellis.jvector.graph.disk.LVQView;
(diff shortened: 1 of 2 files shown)


jbellis:
LGTM

suggested commenting:
```
            // write the min and max
            out.writeFloat(bias);
            out.writeFloat(scale);

            // write the main part of the encoded vector that is evenly divisible into 64-byte blocks
            int mainBlockCount = bytes.length() / 64;
            int i;
            for (i = 0; i < mainBlockCount; i++) {
                var startIndex = i * 64;
                for (int j = startIndex; j < startIndex + 16; j++) {
                    out.writeByte(bytes.get(j));
                    out.writeByte(bytes.get(j + 16));
                    out.writeByte(bytes.get(j + 32));
                    out.writeByte(bytes.get(j + 48));
                }
            }

            // write the "tail" bytes from the last partial block
            var startIndex = i * 64;
            if (startIndex < bytes.length()) {
                var endIndex = Math.min(startIndex + 16, bytes.length());
                int j = startIndex;
                for (; j < endIndex; j++) {
                    writeByteSafely(out, bytes, j);
                    writeByteSafely(out, bytes, j + 16);
                    writeByteSafely(out, bytes,  j + 32);
                    writeByteSafely(out, bytes, j + 48);
                }
                // pad the rest of the block with zeros
                for (; j < startIndex + 16; j++) {
                    out.writeInt(0);
                }
            }
```

PR #270, opened 2024-04-08
Reduce allocation by pooling GraphSearcher objects

jbellis:
Build and query allocations both reduced about 75% (16 GB -> 4GB, 6 GB -> 1.5 GB)

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
@@ -169,6 +169,9 @@ public class NodeQueue {
         heap.clear();
     }
 
+    /**
+     * Set the max size of the underlying heap.  Only valid when NodeQueue was created with BoundedLongHeap.
+     */
     public void setMaxSize(int maxSize) {
         ((BoundedLongHeap) heap).setMaxSize(maxSize);
     }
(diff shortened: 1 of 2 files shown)


jbellis:
Wall clock time doesn't change for Bench but I imagine it would show up if we stressed GC more in a long-running process like C*.

jvector-base/src/main/java/io/github/jbellis/jvector/util/ExplicitThreadLocal.java
```diff
@@ -33,10 +33,11 @@
  * <p>
  * Because ExplicitThreadLocal doesn't hook into Thread internals, any referenced values
  * can be GC'd as expected as soon as the ETL instance itself is no longer referenced.
+ * ExplicitThreadLocal also provides a `closeAll` method to cleanup non-GC'd resources.
```
jkni: Method isn't called `closeAll`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
```diff
@@ -168,6 +169,10 @@ public void clear() {
         heap.clear();
     }
 
+    public void setMaxSize(int maxSize) {
```
jkni, line 175: We should probably Javadoc that this depends on NodeQueue being initialized with a BoundedLongHeap

PR #271, opened 2024-04-09
V2 ODGI extensibility (mix-and-match features)

jkni:
Implement extensible OnDiskGraphIndex that absorbs functionality of all existing on-disk indices, including LVQ/QuickADC. Remove node contents from graph caches, such that the cache only reflects graph structure. Add extensibility/versioning to ODGI structure.


jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FeatureId.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Set;
+import java.util.function.BiFunction;
+
+/**
+ * An enum representing the features that can be stored in an on-disk graph index.
+ * Order of this enum is critical! It will be serialized on-disk in enum order.
+ * These are typically mapped to a Feature or FeatureWriter.
+ */
+public enum FeatureId {
+    INLINE_VECTORS(InlineVectors::load),
```
jbellis, line 32: #collaborating

it might be cleaner to have it take (int bitflag, loader) to make the flags explicit instead of implicit wrt the order of the enum 

means you can always get the flag from the enum itself without having to count through values()

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FeatureWriter.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.io.DataOutput;
+import java.io.IOException;
+
+/**
+ * A feature of an on-disk graph index. Information to use a feature is stored in the header on-disk.
+ * This interface extends {@link Feature} to add the ability to write the feature inline, which
+ * may require more context than is stored on-disk.
+ */
+interface FeatureWriter extends Feature {
```
jbellis, line 27: Is this premature generalization?  It looks to me like all of our Features today require writeInline

and if we had a future inline-free Feature we could just make the method a no-op

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FusedADCNeighbors.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.jbellis.jvector.graph.disk;
+
+import io.github.jbellis.jvector.vector.types.ByteSequence;
+
+public interface FusedADCNeighbors {
```
jbellis, line 21: there is only one implementation, do we need the extra interface?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/Header.java
```diff
@@ -0,0 +1,80 @@
...
+
+    }
+
+    static Header load(RandomAccessReader reader, long offset) throws IOException {
+        reader.seek(offset);
+        int maybeMagic = reader.readInt();
+        int version;
+        EnumSet<FeatureId> featureIds;
+        EnumMap<FeatureId, Feature> features = new EnumMap<>(FeatureId.class);
+        int size;
```
jbellis, line 54: these can move into the old-format block

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -140,86 +116,154 @@ public void close() throws IOException {
...
-                                Consumer<DataOutput> headerWriter,
-                                InlineWriter inlineWriter,
-                                Map<Integer, Integer> oldToNewOrdinals)
-            throws IOException
-    {
-        if (graph instanceof OnHeapGraphIndex) {
-            var ohgi = (OnHeapGraphIndex) graph;
-            if (ohgi.getDeletedNodes().cardinality() > 0) {
-                throw new IllegalArgumentException("Run builder.cleanup() before writing the graph");
+    public class View implements GraphIndex.View, ScoringView, RandomAccessVectorValues {
```
jbellis, line 129: GI.View is redundant

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FusedADCNeighbors.java
jbellis: looks like this is for consistency w/ some of the others in which case I'm agnostic

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -140,86 +116,154 @@ public void close() throws IOException {
...
+        public Bits liveNodes() {
+            return Bits.ALL;
+        }
+
+        @Override
+        public void close() throws IOException {
+            reader.close();
+        }
+
+        public ScoreFunction.ExactScoreFunction rerankerFor(VectorFloat<?> queryVector, VectorSimilarityFunction vsf, Set<FeatureId> permissibleFeatures) {
```
jbellis, line 241: is it worth illustrating how to use in Grid?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -0,0 +1,210 @@
...
+        public Builder(GraphIndex graphIndex) {
+            this(graphIndex, getSequentialRenumbering(graphIndex));
+        }
+
+        public Builder(GraphIndex graphIndex, Map<Integer, Integer> oldToNewOrdinals) {
+            this.graphIndex = graphIndex;
+            this.oldToNewOrdinals = oldToNewOrdinals;
+        }
+
+        public Builder withFusedADC(PQVectors pqVectors) {
```
jbellis, line 187: WDYT about having instead a with(Feature) method so we don't have to hardcode different pieces of state in the Builder and Writer?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -265,7 +270,6 @@ private static ResultSummary performQueries(ConfiguredSystem cs, int topK, int e
                 var searcher = cs.getSearcher();
                 var sf = cs.scoreProviderFor(queryVector, searcher.getView());
                 sr = searcher.search(sf, efSearch, Bits.ALL);
-
                 var gt = cs.ds.groundTruth.get(i);
```
jbellis, line 273: the newline separated "run the search" from "process the results" but explicit comments would be better

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FeatureWriter.java
jbellis: it looks like the distinction is we only need Feature at read time, while the Writers need additional state

so it probably makes more sense to push writeHeader into FeatureWriter instead of trying to eliminate FeatureWriter

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
jkni: will add some, this change was accidental during rebase on top of latest changes

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
jkni: Yes, that was the original plan but didn't work due to some early design choices that have since changed. I'll give that another shot as I don't think those limitations apply any more. Most of the refactoring in this PR was to reduce the amount of places the same structure was hardcoded in builder/writer/reader, so I'm fully on-board with that goal.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FusedADCNeighbors.java
jkni: Yes, this is for consistency/some mocking in tests.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FeatureWriter.java
jkni: Yes, that's the source of the distinction. I also originally pushed writeHeader into FeatureWriter, but that caused some complications for Header, which can currently write Feature headers without the additional state needed for Writers. Fundamentally, this distinction exists only because the state needed to consume features is different than the state needed to write features. I'll try pulling writeHeader into FeatureWriter and see if I can resolve the Header issue I had on the first pass.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FeatureId.java
jkni: That makes sense. I'll give this a shot.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FeatureWriter.java
jbellis: Looks like the only code path that calls writeHeader has a FeatureWriter already, so I went ahead and did this in my branch (with a cast).

PR #296, opened 2024-04-16
Add a memory-mapped RandomAccessReader using MemorySegment api

### LINKED ISSUE #278: add a RandomAccessReader implementation for jdk 20+ using modern MMap or MemorySegment
SimpleMappedReader being limited to 2GB is very limiting

mdogan:
I'd prefer this to be in `jvector-twenty` module. But `--enable-preview` flag is only allowed for the Java release version used to compile the code. When building with Java 22, `--enable-preview` is not allowed on `twenty` module because it builds for Java 20.

Closes #278


jbellis:
Outstanding!

Can you wire this into ReaderSupplierFactory (replacing MMapReaderSupplier)?  Not sure if reflection is needed given the module configuration.

mdogan:
> Outstanding!
> 
> Can you wire this into ReaderSupplierFactory (replacing MMapReaderSupplier)? Not sure if reflection is needed given the module configuration.

Done.

mdogan:
> Thank you @mdogan! I feel your pain here regarding module placement. We could adopt Maven Toolchains so that we could call out to a Java 20 compiler in `jvector-twenty`, allowing us to use preview features there. Any strong feelings about whether that's worth it? It only introduces marginal additional complexity in the build.

Actually I don't about Java 20 but maybe it would be useful for Java 21 users (since it's LTS). But your new native extensions are targeting 22, so this can remain in Java 22 too.

jbellis:
LGTM. Ship it!

jbellis:
(pushing the button since i want to put this in beta3)

PR #300, opened 2024-04-19
Flesh out support for reranking from inline full/lvq vectors while building graph using pq

jbellis:


jvector-base/src/main/java/io/github/jbellis/jvector/graph/CachingVectorValues.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+package io.github.jbellis.jvector.graph;
+
+import io.github.jbellis.jvector.pq.PQVectors;
+import io.github.jbellis.jvector.vector.types.VectorFloat;
+import org.agrona.collections.Int2ObjectHashMap;
+
+/**
+ * This is NOT a general "make vectors go faster" class.  It is used specificly by diversity computations
```
jkni:
```suggestion
 * This is NOT a general "make vectors go faster" class.  It is used specifically by diversity computations
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -17,20 +17,31 @@
...
+import java.nio.file.Path;
+import java.util.Collection;
 import java.util.EnumMap;
 import java.util.Map;
 import java.util.function.IntFunction;
 
 /**
  * Write a graph index to disk, for later loading as an OnDiskGraphIndex.
+ * <p>
+ * Implements FeatureReader to allow incremental construction of a larger-than-memory graph
```
jkni: FeatureReader an outdated name? I think you're referring to it working as a feature source

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -98,116 +102,202 @@ static void runAll(DataSet ds,
...
+                                                                 int efConstruction,
+                                                                 DataSet ds,
+                                                                 Path testDirectory)
+            throws IOException
+    {
+        var floatVectors = ds.getBaseRavv();
+        Map<Set<FeatureId>, GraphIndex> indexes = new HashMap<>();
+        long start;
+        var bsp = BuildScoreProvider.randomAccessScoreProvider(floatVectors, ds.similarityFunction);
+        GraphIndexBuilder builder = new GraphIndexBuilder(bsp, floatVectors.dimension(), M, efConstruction, 1.2f, 1.2f,
```
jkni, line 276: intentional change from 1.5f, 1.2f here? (compared to building on disk and building in memory prior to change)

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -98,116 +102,202 @@ static void runAll(DataSet ds,
...
+            var writer = bws.builder.build();
+            writers.put(features, writer);
+            suppliers.put(features, bws.suppliers);
+            if (features.equals(EnumSet.of(FeatureId.LVQ))) {
+                scoringWriter = writer;
+                lvq = (LVQ) bws.builder.getFeature(FeatureId.LVQ);
+            }
+        }
+        if (scoringWriter == null) {
+            throw new IllegalStateException("Incremental compressed builds are only supported by Bench when one of the feature sets is {LVQ}");
```
jkni: out of curiosity, why? Couldn't we do this with full inline vectors?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
```diff
@@ -137,13 +138,14 @@ public SearchScoreProvider.Factory diversityProvider() {
 
     /**
      * Returns a BSP that performs approximate score comparisons using the given PQVectors,
-     * with reranking performed using full resolutions vectors read from the reader
+     * with reranking performed using LVQ PackedVectors
```
jkni: Couldn't this all be done with full vectors?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
jbellis: yes, will fix

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
jbellis: i think it's correct as is now, it's supposed to be (1.5, 1.2) for on disk and (1.2, 1.2) for in memory

(retainDiverse is more expensive when we have to fetch from disk, so we allow it to overflow more to call that less frequently)

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
jbellis: yes, this is purely a Bench limitation not a general one, everything works if you switch LVQ to INLINE_VECTORS

jbellis: I didn't want to try to figure out the right thing to do if user specifies both IV and LVQ

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
jbellis: yes, comment is outdated, will update

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
jbellis: clarified the message

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
jkni: That makes sense.

PR #304, opened 2024-04-29
Switch Fused ADC from 32-cluster to 256-cluster PQ, maxDegree 32 graphs

jkni:
This PR implements support for 256 cluster fused ADC in the style of Quicker ADC. As before, transposed quantized neighbors are fused into the graph structure. To reduce the number of shuffles/blends when doing bulk shuffle edge similarity, partial sum tables are quantized to 16-bit ints and summed with saturating arithmetic.

Several interesting bits of the PR:

- Quantization of the partial sums to 8-bit ints was attempted. This had too significant an adverse effect of recall.
- Quantization bounds are decided based on worst distances encountered during the search. As such, there is a tradeoff between finding the worst distance required and the number of slower evaluations before the quantization of partial sums can be performed. This PR currently evaluates maxDegree slow similarities before quantizing with no noticeable impact on recall.
- Earlier versions of this PR used LVQ for the slower evaluations prior to quantization. This was accelerated by implementing a specialized assemble/sum of the transposed neighbors which has not yet been SIMD-ified. This is still a performance improvement. During review, I will attempt to SIMD-ify this and PR a follow up if this has a noticeable performance improvement.
- LVQ/Fused ADC shows no noticeable recall drop over LVQ in Bench.
- Supporting varying max degrees/varying cluster counts is possible, but it will require additional complexity in the native implementations. I've deferred this in support of evaluating quality in the overall case. If this is valuable, we can do it as a follow up.
- Unsigned shorts are being read/written to the ByteSequence. This incurs some additional mental overhead at call sites. It might be worth adding a ShortSequence interface.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -20,9 +20,11 @@ If you only read one thing, read this!
 - Experimental support for native code acceleration has been added. This currently only supports Linux x86-64 
   with certain AVX-512 extensions. This is opt-in and requires the use of MemorySegment `VectorFloat`/`ByteSequence`
   representations.
-- Experimental support for fused graph indexes has been added. These work best in concert with native code acceleration.
+- Experimental support for fused ADC graph indexes has been added. These work best in concert with native code acceleration.
+  Without the NativeVectorizationProvider, results using fused ADC will be valid but performance will degrade.
   This explores a design space allowing for packed representations of vectors fused into the graph in shapes optimal
-  for approximate score calculation. This is a new type of graph index and is opt-in.
+  for approximate score calculation. This is a new feature of graph indexes and is opt-in. At this time, only graphs with
+  maxNeighbors=32 and 256-cluster ProductQuantization can use fused ADC.
 - Support for larger-than-memory graph construction by using quantized vectors + rerank for the searches
   performed during construction.
 - Support for Anisotropic Product Quantization as described in "Accelerating Large-Scale Inference with Anisotropic Vector Quantization"


jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FusedADCNeighbors.java
```diff
@@ -25,4 +25,10 @@ public interface FusedADCNeighbors {
      * @return the quantized, transposed, and packed vectors of the given node's neighbors.
      */
     ByteSequence<?> getPackedNeighbors(int node);
+
+    /**
+     * Get the maximum degree of the graph for which this FusedADCNeighbors instance is used.
+     * @return the maximum degree of the graph.
+     */
+    int maxDegree();
```
jbellis: Are you planning to add more implementations of FADCN?  If not I'd prefer to just inline it into the implementation

jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/ByteSequence.java
```diff
@@ -31,6 +31,12 @@ public interface ByteSequence<T> extends Accountable
 
     void set(int i, byte value);
 
+    /**
+     * @param byteIndex byte index inside the sequence to start setting short value
+     * @param value short value to set
+     */
+    void setLittleEndianShort(int byteIndex, short value);
```
jbellis: if we're not adding a ShortSequence, maybe indexing shorts in shortIndex instead of bytes would be a minor improvement

jbellis:
LGTM pending your review of CachingDecoder changes and the potential TODO note there

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/FusedADCNeighbors.java
jkni: I'll inline it

jbellis:
ship it!

PR #306, opened 2024-05-03
Fix usage of Float.MIN_VALUE

jkni:
Float.MIN_VALUE is not what I thought.

Incidentally, I noticed that the VectorSimdOps min/max implementation is less efficient than SimdOps. Fixing those while I'm here.


jbellis (approved):
LGTM

PR #307, opened 2024-05-03
Check for GCC 11+ instead of failing compile

jkni:
With Fused ADC merged, we use GCC 11+ intrinsics.


jbellis (approved):
confirmed that this lets me build on gcc 9.4

PR #315, opened 2024-05-09
add getFeatureSet methods to ODGI and ODGIW

jbellis:
apply this to not relying on exceptions for SSP construction in Grid

(C* will use it too)


jbellis:
I removed the `permissibleFeatures` overload of `approximateScoreFunctionFor`, not completely sure what you had in mind there but istm we can apply the YAGNI principle for now

PR #316, opened 2024-05-10
improve memory usage during construction

jbellis:
commit 2 does most of the work, the key is that when insert() or insertDiverse() is called, we copy the source nodes into a size-limited NodeArray instead of assuming that the source array is correctly sized.  commit 2 also inlines and removes removeAllNonDiverse to improve clarity, that method was the only one related to diversity that did internal copies which was a bit confusing.

commit 3 does some fairly mechanical improvements of ramBytesUsed; most methods were not including object headers which for the ones where we have millions of instances (CNS and ArrayByteSequence) becomes material

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
@@ -16,6 +16,7 @@
 
 package io.github.jbellis.jvector.graph;
 
+import io.github.jbellis.jvector.annotations.VisibleForTesting;
 import io.github.jbellis.jvector.graph.similarity.BuildScoreProvider;
 import io.github.jbellis.jvector.graph.similarity.ScoreFunction;
 import io.github.jbellis.jvector.util.BitSet;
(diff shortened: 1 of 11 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
```diff
@@ -179,10 +190,13 @@ public void insertDiverse(NodeArray toMerge) {
                 merged = toMerge.copy(); // still need to copy in case we lose the race
                 retainDiverse(merged, 0, scoreProvider.isExact());
             }
-            return new Neighbors(merged, merged.size);
+            // insertDiverse usually gets called with a LOT of candidates, so trim down the resulting NodeArray
+            var nextNodes = merged.node.length <= nodeArrayLength() ? merged : merged.copy(nodeArrayLength());
+            return new Neighbors(nextNodes, merged.size);
```
jkni: I don't think it has adverse consequences right now, but it seems weird to set diverse before based on merged.size when nextNodes can be smaller than merged.

jbellis: nextNodes and merged have the same *size* (we don't throw away any of the contents) but different *length*

jbellis: but you're right that it's probably more clear to write `nextNodes.size`, even though they are equal

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborSet.java
```diff
@@ -51,28 +53,33 @@ public class ConcurrentNeighborSet {
...
     private final BuildScoreProvider scoreProvider;
 
-    /** the maximum number of neighbors we can store */
-    private final int maxConnections;
+    /** the maximum number of neighbors we will end up with */
+    private final int maxDegree;
+    /** the maximum number of neighbors we can have temporarily during construction */
+    private final int maxOverflowDegree;
 
     /** the proportion of edges that are diverse at alpha=1.0.  updated by removeAllNonDiverse */
```
jkni: tiny nit, but removeAllNonDiverse doesn't exist any more

PR #325, opened 2024-05-22
Reduce the per-node overhead of edge lists in OnHeapGraphIndex

jbellis:
Sizes for building with Bench (32 degree * 1.2 overflow):

Raw data (NodeArray nodes, scores, and size fields but not the NodeArray class header or array references): 436 bytes

Before: 
436 + 44 [NodeArray overhead] + 24 [AtomicReference overhead] + 72 [CNS overhead] = 576 bytes

After: 436 + 44 [NodeArray overhead] + 40 [CNM.Neighbor overhead] = 520 bytes

I also tried making CNM.Neighbor static, which I expected would reduce the overhead of those instances by eliminating the implicit reference to CNM.this, but it did not actually change the observed size in the heap dump, so I reverted the change.  (I'm still not sure why this did not make a difference.)


jbellis:
Round 2:
- Neighbors is static again, this time for keeps
- Neighbors extends NodeArray to save the object reference (in Neighbors) and the object header (in the NodeArray instance)
- Delete shortEdges field and compute it during cleanup instead

This brings us to
436 + 60 [Neighbor overhead] = 496 bytes

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborMap.java
```diff
@@ -0,0 +1,506 @@
...
+
+        NodeArray getCurrent() {
+            return this;
+        }
+
+        /**
+         * Insert a new neighbor, maintaining our size cap by removing the least diverse neighbor if
+         * necessary. "Overflow" is the factor by which to allow going over the size cap temporarily.
+         */
+        public Neighbors insert(int neighborId, float score, float overflow, ConcurrentNeighborMap map) {
```
jkni: Make this private like the other mutating methods?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborMap.java
```diff
@@ -0,0 +1,506 @@
...
+     * object header.
+     */
+    public static class Neighbors extends NodeArray {
+        /** the node id whose neighbors we are storing */
+        private final int nodeId;
+
+        /** entries in `nodes` before this index are diverse and don't need to be checked again */
+        private int diverseBefore;
+
+        Neighbors(int nodeId, int intialSize, int diverseBefore)
```
jkni: nit: s/intialSize/initialSize

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborMap.java
```diff
@@ -391,7 +419,7 @@ private boolean isDiverse(int node, float score, NodeArray others, ScoreFunction
         }
 
         NodeArray getCurrent() {
-            return nodes;
+            return this;
```
jkni: With this PR now changing where the AtomicReferences are maintained and Neighbors extending NodeArray, this is almost entirely vestigial (except as a way to change the type bounds). Maybe we should just remove it entirely?

PR #326, opened 2024-05-23
avoid making a copy in insert() if the new neighbor already exists or we're at capacity and it's worse than our worst edge

jbellis:


jbellis:
test failure is reproducible and does not appear in `main`.  investigating...

jbellis:
so in fact this is an incorrect optimization -- we should not throw away potential edges, even "bad" ones, before checking diversity.  Doing so is causing the graph in the test to be disconnected.  I will close the PR and edit out the TODO.

PR #327, opened 2024-05-23
Remove on-disk reranking

jbellis:
On-disk reranking looks fast enough in Bench but that is because the datasets are small enough that the index under construction stays in the page cache, so reranking "from disk" is really reranking from page cache.

Performance drops off rapidly as the index outgrows the page cache.

Testing shows that the difference in accuracy between an index constructed with reranking and one constructed only with PQ is small and can be made up with minimal (~10%) more overquery.

Thus, it looks like rerank-from-disk during construction is not in fact a useful option and we are better off simplifying the code around LTM construction without it.


jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborMap.java
```diff
@@ -289,31 +285,11 @@ private Neighbors insertDiverse(NodeArray toMerge, ConcurrentNeighborMap map) {
          * a new NodeArray; does not modify `old` or `toMerge`
          */
         private NodeArray rescoreAndRetainDiverse(NodeArray old, NodeArray toMerge, ConcurrentNeighborMap map) {
```
jkni: Is this worth renaming now that it rescoring isn't a thing?

jbellis: good idea.  inlined it instead

PR #328, opened 2024-05-24
Ecapsulate NodeArray internals

jbellis:
Goal is to make it easier to audit ConcurrentNeighborMap to make sure we're not mutating the edge lists without copying first.

Did not see any violations of this rule but did fix a minor bug with insertNotDiverse as part of the first (size-related) commit, see the new assert.

(Builds on top of the rerank PR.)


jbellis:
(I was thinking of also adding a `boolean sealed` to NodeArray and make mutation methods throw if it's set, but that conflicts with the goal of minimizing the footprint of NodeArray/CNM, so I left it out.)

PR #332, opened 2024-05-28
add writeHeader and getPath methods to OnDiskGraphIndexWriter

jbellis:


jbellis:
I think we should flush it.

PR #333, opened 2024-05-28
add ForDatabases annotation

jbellis:


jbellis:
Good idea?  Is there a better name?  `@AdvancedAPI`

TBH I'm mostly trying to guard against my own urges to delete unused code, maybe this isn't really necessary at all

PR #335, opened 2024-05-30
Reduce tendency of reconnectOrphanedNodes to leave orphaned nodes

jkni:
This PR proposes three changes, each addressing a different source of disconnected graphs.

1. Update the entry node before reconnecting orphaned nodes. If the entry node is updated after, orphaned nodes that were reconnected to the new entry node tend to be disconnected by improvements to the new entry node.
2. Preserve connection targets map across loop iterations. When reset on each iteration of the loop, it's not exceedingly rare to see disconnection cycles where on run 0, node A is reconnected (disconnecting B) and node C is reconnected to node D in its neighbor list. Then, if D is the best neighbor of B, C/B will alternate being reconnected across future iterations.
3. Exclude connectionTargets when searching for new connection points. If all of a node's existing neighbors have been considered and a search is performed, it is not unusual to see that the existing neighbors are many of the search results. This is particularly important if approach 2 is used, as connection targets will grow across loop iterations.

Testing on real data sets looks positive. This approach has also been successful at reconnecting graphs build from random vectors.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -236,9 +236,11 @@ public class GraphIndexBuilder implements Closeable {
     private void reconnectOrphanedNodes() {
         var searchPathNeighbors = new ConcurrentHashMap<Integer, NodeArray>();
         // It's possible that reconnecting one node will result in disconnecting another, since we are maintaining
-        // the maxConnections invariant. So, we do a best effort of 5 loops.
+        // the maxConnections invariant. So, we do a best effort of 3 loops. We claim the entry node as an
+        // already used connectionTarget so that we don't clutter its edge list.
         var connectionTargets = ConcurrentHashMap.<Integer>newKeySet();
-        for (int i = 0; i < 5; i++) {
+        connectionTargets.add(graph.entry());
+        for (int i = 0; i < 3; i++) {
             // find all nodes reachable from the entry node
             var connectedNodes = new AtomicFixedBitSet(graph.getIdUpperBound());
             connectedNodes.set(graph.entry());


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -225,18 +225,19 @@ public void cleanup() {
                     .average()
         ).join().orElse(Double.NaN);
 
+        // optimize entry node -- we do this before reconnecting, as otherwise, improving the entry node's
+        // connections will tend to disconnect any orphaned nodes reconnected to the entry node
+        updateEntryPoint();
```
jbellis, line 230: should we special case orphan reconnect to ignore entry node?  seems like it's important to not clutter that with poor quality edges

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -225,18 +225,19 @@ public void cleanup() {
...
-        // optimize entry node
-        updateEntryPoint();
     }
 
     private void reconnectOrphanedNodes() {
         var searchPathNeighbors = new ConcurrentHashMap<Integer, NodeArray>();
         // It's possible that reconnecting one node will result in disconnecting another, since we are maintaining
-        // the maxConnections invariant.  In an extreme case, reconnecting node X disconnects Y, and reconnecting
-        // Y disconnects X again.  So we do a best effort of 3 loops.
+        // the maxConnections invariant. So, we do a best effort of 5 loops.
```
jbellis: can we switch back to 3 with this change?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -263,13 +263,14 @@ private void reconnectOrphanedNodes() {
 
                 // no unused candidate found -- search for more neighbors and try again
                 neighbors = searchPathNeighbors.get(node);
-                if (neighbors == null) {
+                // run search again if neighbors is empty or if every neighbor is already in connection targets
+                if (neighbors == null || isSubset(neighbors, connectionTargets)) {
```
jbellis, line 269: half-baked thought: would it be cleaner to mutate the cached neighbors as we add connections?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jkni: Yes, that's a good idea.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jkni: I think so - I've never seen it take more than two rounds.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jkni: Maybe? I think it does more work in the common case, as our most common case is to never need to search for neighbors, the second most common (but much less likely) case is to search for neighbors once and find a good result, and the third most common case (even more unlikely) is to have to search again because every search-discovered neighbor is a connection target. Since we rarely hit this case, mutating the neighbors on a connection addition is probably mutating a data structure that will never be accessed again. But, that's also not so much work that we should avoid it if it's significantly cleaner. I could go either way.

jbellis: let's leave it alone, then

jbellis (approved):

PR #337, opened 2024-06-07
Clear scratch structures if search terminates exceptionally

### LINKED ISSUE #336: JVector should clear scratch search structures when a search terminates exceptionally
A GraphSearcher maintains scratch search data structures and asserts that these are empty when resuming, as the previous search should have emptied their contents. If a query terminates exceptionally, these should be cleared in an exception handler, to ensure they're in a good state for the next search.

jkni:
Code asserts that previous runs emptied the scratch structures, which happens at the end of the query.

Closes #336

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
@@ -364,11 +364,11 @@ public class GraphSearcher implements Closeable {
             assert popFromQueue.size() == 0;
 
             return new SearchResult(nodes, numVisited, worstApproximateInTopK);
-        } catch (Exception e) {
-            // clear scratch structures if terminated via exception, as they may not have been drained
+        } catch (Throwable t) {
+            // clear scratch structures if terminated via throwable, as they may not have been drained
             approximateResults.clear();
             rerankedResults.clear();
-            throw e;
+            throw t;
         }
     }
 


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -246,123 +246,130 @@ SearchResult searchInternal(SearchScoreProvider scoreProvider,
...
             }
+            // that should be everything
+            assert popFromQueue.size() == 0;
 
-            worstApproximateInTopK = Float.POSITIVE_INFINITY;
-            popFromQueue = approximateResults;
-        } else {
-            worstApproximateInTopK = approximateResults.rerank(topK, scoreProvider.reranker(), rerankFloor, rerankedResults, evictedResults);
+            return new SearchResult(nodes, numVisited, worstApproximateInTopK);
+        } catch (Exception e) {
```
eolivelli: Dummy question: why Exception and not Throwable ?

eolivelli (approved):
+1

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
jkni: I think you're right and we should catch throwable here. In the event the throwable is recovered from by calling code, we should still correct the state as possible.

PR #340, opened 2024-06-21
extract RandomAccessWriter interface from BRAW

jbellis:
ODGIW.path is removed

commentary on path removal:

OnDiskGraphIndexWriter wanted to own a raw Path to create its writer from because back when we thought we needed to rerank from disk during LTM construction, we'd use the Path to create a per-thread RandomAccessFile backing a custom FeatureSource.

Since then I think we came up with a cleaner solution, which is to call ODGI.writeHeader, and then open up a normal ODGI and use its read methods.

Besides being cleaner, this has the advantage of making during-construction reads use the same well-optimized path as other reads instead of being a special snowflake.

Finally, if this is wrong and we do in fact find a need to open additional handles against the output file in ODGIW, we could add a getPath method to the RandomAccessWriter interface and we still wouldn't need to change the ODGIW api again.


eolivelli (approved):
+1

PR #349, opened 2024-07-31
add support for non-sequential remapped ordinals

jbellis:


jvector-base/src/main/java/io/github/jbellis/jvector/util/DenseIntMap.java
```diff
@@ -102,10 +102,15 @@ public int size() {
      * @return the value of the key, or null if not set
      */
     public T get(int key) {
+        if (key < 0) {
```
jkni: Is this change load-bearing? I don't immediately see when we'd call this with OMITTED.

jbellis: one of the tests (i think it's the C* range test that starts with -100 as the start key) ended up with a negative key here

i can probably make the test behave better instead

jkni: I'm happy either way. My main goal was to make sure I understood the change.

PR #355, opened 2024-09-09
Handle unreachable nodes from the entry point better while removing deleted nodes

mdumandag:
It is unlikely, but possible to have no path to some nodes from the entry point of the graph. It might happen for example when you have added a group of nodes that are very close to each other, and then added another group of nodes that are very far away from the first group such that there are no backlinks to these new nodes from the initial group.

In that case, when you delete every node from the initial group, you end up with some live nodes that are unreachable. So, we set the entry point to -1 in removeDeletedNodes, although we have some live nodes.

That breaks the logic in cleanup method, as we expect a valid entry point when the graph is not empty. We recently hit a production issue in a scenario what we believe was similar to this(we can't confirm it fully as we can't access the customer data, but deploying a new version with this fix solved the issue).

To handle this case, when the entry point is deleted in removeDeletedNodes, we try to update the entry point as usual. If we can't do that although there are some live nodes, we then assign a random entry point to the graph.


mdumandag:
The unlikely scenario I mentioned can be reproduced when you add 1000 nodes with the same values(say with the values (0.01, 0.0)) and then add another 100 nodes with again same values but very far away from the initial value (say with (1.0, 1.0))

jbellis:
Graph indexes are not designed to tolerate duplicate nodes being added.  Does the problem occur outside that scenario?

mdumandag:
Here is some a reproducer 
```
    @Test
    public void testNoPathToLiveNodesWhenRemovingDeletedNodes2() throws IOException {
        var vectorTypeSupport = VectorizationProvider.getInstance().getVectorTypeSupport();
        var random = getRandom();
        var ravv = MockVectorValues.fromValues(
                IntStream.range(0, 1100).mapToObj(i -> {
                    if (i < 1000) {
                        return vectorTypeSupport.createFloatVector(new float[]{0.01f + 100 * random.nextFloat(), 0.01f + 100 * random.nextFloat()});
                    } else {
                        return vectorTypeSupport.createFloatVector(new float[]{10_000.0f + 10 * random.nextFloat(), 10_000.0f + 100 * random.nextFloat()});
                    }
                }).toArray(VectorFloat<?>[]::new)
        );

        try(var builder = new GraphIndexBuilder(ravv, VectorSimilarityFunction.COSINE, 2, 10, 1.0f, 1.0f)) {
            for (int i = 0; i < 1100; i++) {
                builder.addGraphNode(i, ravv.getVector(i));
            }

            for (int i = 0; i < 1000; i++) {
                builder.markNodeDeleted(i);
            }

            builder.cleanup();
        }
    }
```

mdumandag:
The actual problem is that there are no backlinks to the second group (nodes 1000 to 1100) from the first one (nodes 0 to 1000). The last 100 nodes would also be not reachable in the searches etc., but I don't now what we can do about them. The heuristic of choosing edges between more similar nodes while cleaning up the overflowed ones seems correct to me, but it leads to scenario like this

jbellis:
understood, the new test is a good one.

the core problem is that approximateMediod is returning NO_ENTRY_POINT when the graph is not empty, which is a violation of its contract.  So we should fix that instead of adding a separate repair step to cleanup.

but the reason approximateMediod doesn't work is that the graph is disconnected before it is called.  it's difficult to reason about graphs that aren't connected and I'm not sure we can do better than picking a random node in the general case.

so I think what I would prefer is a check in approximateMediod to handle the disconnected case.

I will open an alternative PR with that approach.

(Side note: if you're calling removeDeletedNodes() outside cleanup() then you should probably also call reconnectOrphanedNodes(), but I don't want to move the one inside the other.)

mdumandag:
Ok, sounds good. I am going to close this one then.

(No, we were only calling markNodeDeleted + cleanup, but the problematic entry point was coming from the removeDeletedNodes inside cleanup, hence I added the test that way)

jbellis:
please review https://github.com/jbellis/jvector/pull/356

PR #356, opened 2024-09-09
approximateMediod returns a random node when the graph is too disconnected to search for the centroid

jbellis:
alternative solution to the problem described in https://github.com/jbellis/jvector/pull/355


mdumandag (approved):
LGTM, thanks

PR #359, opened 2024-09-19
Improve performance of reconnectOrphanedNodes

jkni:
reconnectOrphanedNodes's ability to produce connected graphs was improved by #335. One change in particular produces a performance regression when reconnecting large (1M+ node graphs) with a sizeable partitioning (e.g., only ~600K nodes are reachable from the entry point, meaning 400k nodes need to be reconnected).

The change is as follows:

```
Exclude connectionTargets when searching for new connection points. If all of a node's existing neighbors have been considered and a search is performed, it is not unusual to see that the existing neighbors are many of the search results. This is particularly important if approach 2 is used, as connection targets will grow across loop iterations.
```

This works very well when a small number of nodes need to be reconnected, as the set of connection targets is small. When it's large, like when reconnecting 400k nodes, the performance of the searches is extremely poor.

This PR proposes several changes:

1. Limit neighbors used as connection targets to nodes connected at the start of the loop iteration. Testing the current reconnection behavior for graphs with large numbers of disconnected nodes showed that the neighbor reconnection has a tendency to connect to nodes already in the partition. This change improves the effectiveness of neighbor reconnection.
2. Instead of using excludeBits, use resumable searches and post-filter results based on connection targets. This removes the performance hit of connection targets. The threshold of 50 resumes was decided by a basic histogram analysis of the number of resumes needed. Past this point, we see diminishing returns.
3. Increase reconnect loops. The previous decrease from 5 to 3 is no longer appropriate to see highly effective restoration of connectivity on large, partitioned graphs.
4. Introduce slf4j-api for basic debug logging. java.util.logging bridged to other logging frameworks via slf4j has some performance concerns. This seemed like the appropriate way to bridge logging to the broader ecosystem.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -289,14 +289,14 @@ public class GraphIndexBuilder implements Closeable {
                     var j = 0;
                     // no need to filter to connected nodes here, as they're connected by virtue of being reachable via
                     // search
-                    var reconnectedTo = connectToClosestNeighbor(node, neighbors, globalConnectionTargets);
+                    var reconnectedTo = connectToClosestNeighbor(node, neighbors, Bits.ALL, globalConnectionTargets);
                     // if we can't find a valid connectionTarget within 2*degree of the search destination, give up
                     while (reconnectedTo == null && j < 2 * graph.maxDegree) {
                         j++;
                         nResumesRun.incrementAndGet();
                         result = gs.resume(beamWidth, beamWidth);
                         toScratchCandidates(result.getNodes(), neighbors);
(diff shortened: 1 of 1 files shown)


jbellis:
Looks good overall.

I'm not sure about switching from excludebits to search/resume, especially since we end up with code that uses both.  Here's my attempt to make excludebits less painful.

jbellis:
Try without the backlinking?

jbellis:
Okay, I added back the split search.  It retains most of the simplicity I think.

If it's still slow then again I blame backlink. :)

jbellis:
(I changed the `50` to 2*degree so that it doesn't break down in the corner case with larger degree, since I left out the excludebits for neighbors as well.)

jbellis:
okay, what's your preferred version at this point?

jbellis:
LGTM

nit: can we combine the two connectToClosestNeighbors by passing Bits.ALL as connectedNodes to the four-parameter version?

PR #362, opened 2024-10-07
Remove max JDK version check

jkni:
Previously, the Panama/Native vectorization providers have a hardcoded max supported runtime version. This was a reasonable precaution when we were using features under heavy modification in Java 20/21/22. Now, these are becoming somewhat more stable, and this may be causing more pain than it's worth.

As of Java 22+, the sole preview feature we were using was the String template in library loader (and string templates have been removed in Java 23). We previously needed this for FFM, but that became stable in Java 22. The vector support is incubating, not preview, so we don't need the enable-preview flag.

We continue to build jvector-native using release version 22, which ensures that JVector built on new versions will still run on Java 22. We don't know when the vector API will leave incubating, but it will likely depend on the release of Valhalla, so we can't set the max version based on that change.


jbellis (approved):
Thanks!

PR #363, opened 2024-10-09
Use fma in SimdOps.cosineSimilarity sum vector

jkni:
This better matches the structure we use in other places, and I suspect it may make work slightly easier for C2 in some cases that I'm debugging.


AbhijitKulkarni1:
Hi @jkni, to ensure code consistency,  suggest to apply the same code change to the cosineSimilarity method in VectorSimdOps.java (io.github.jbellis.jvector.vector.VectorSimdOps#cosineSimilarity) as well.

PR #366, opened 2024-11-06
Use fma in VectorSimdOps.cosineSimilarity

jkni:
Same rationale as https://github.com/jbellis/jvector/pull/363, but this is for the NativeVectorizationProvider rather than the PanamaVectorizationProvider.


marianotepper (approved):
LGTM

PR #368, opened 2024-11-14
Reenable SimdOps.assembleAndSum; implement Panama/Native equivalent for CosineDecoder acceleration

michaeljmarshall:
Fixes: https://github.com/jbellis/jvector/issues/367


marianotepper (commented):
It looks good to me.

My only suggestion would be to use a more descriptive name than decodedCosineSimilarity. Given that we are light on documentation, I would at least use a name like pqDecodedCosineSimilarity to give it some context.

michaeljmarshall:
@marianotepper thanks for your review. I prefixed (and fixed casing) on all `decodedCosineSimilarity` and `decoded_cosine_similarity` method names with `pq` and `pq_` respectively.

marianotepper:
LGTM

PR #369, opened 2024-11-21
WIP: store raw representation of ByteSeqeunce in PQVectors

michaeljmarshall:
I plan to rebase this after #368 is merged.

PR #370, opened 2024-11-23
Store compressed vectors in dense ByteSequence for PQVectors

michaeljmarshall:
The current design stores the compressed vectors using a list of `ByteSequence` objects where each object represents a single vector. That costs us an overhead of 16 bytes for the `ByteSequence` wrapper class and 16 more bytes for the array in the `ArrayByteSequence`. In the case of a 128 dimension vector, we compress PQ down to 64 bytes, so the cost of the wrapper/array is 32 bytes out of the total 96 bytes, which is 33% of the total size. In cases where we have many of these vectors in memory, an extra 33% is quite large.

This PR proposes removing the per-vector wrapper and opts for a more dense representation by

1. Storing the vectors in an array of ByteSequence references where the `ByteSequence` has as many vectors as it can possibly hold.
2. Introducing a `ByteSequence#slice` method to allow for getting a slice of the whole `ByteSequence`. This method is intended to be light weight.
3. Introduce `ArraySliceByteSequence` to store the offset and length information to simplify element retrieval.
4. Introducing a `ByteSequence#offset` method to be used in conjunction with the `get` method to ensure that the Panama and SimdOps methods do not need to perform an array copy when passing the array to Panama.
5. Added assertions in the native code when we call `get` on a `MemorySegmentByteSequence` since we're technically not passing the offset. FWIW, we could consider updating the c to take a `long` offset and a `long` length since those MemorySegments can exceed int length.

We also make some minor breaking changes to the `VectorCompression` interface to allow implementations to construct the `CompressedVectors` object directly. This change simplifies most usages of the `encodeAll` method.

This PR is a more invasive alternative to #369, which only removed one of the two wrappers. 

## Rejected alternatives

Before implementing the PR this way, I tried a few alternatives.

1. Use MemorySegment in the PQVectors to remove the chunking required because we can have more PQVector bytes than int max, which pushes us to store the `compressedVectors` field as a `ByteSequence[]` instead of a constant `ByteSequence`. Further, a MemorySegment would have allowed for a simpler API because of its support in the Panama API. However, this solution proved challenging due to compilation requirements and concerns about using a preview api (MemorySegment) in java 20.
2. Attempt to use a ByteBuffer instead of `ArraySliceByteSequence`. This is plausible, but really just comes down to API design preference. It felt a bit clunky to add it in. The main advantage is that if we add this, we can avoid the `offset` method on the `ByteSequence` interface, which might be convenient. One way we could add it is by making PQVectors know how to create a ByteBuffer wrapping a slice of the `compressedVectors` data.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/BinaryQuantization.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/BinaryQuantization.java
@@ -63,12 +63,13 @@ public class BinaryQuantization implements VectorCompressor<long[]> {
     }
 
     @Override
-    public long[][] encodeAll(RandomAccessVectorValues ravv, ForkJoinPool simdExecutor) {
-        return simdExecutor.submit(() -> IntStream.range(0, ravv.size())
+    public CompressedVectors encodeAll(RandomAccessVectorValues ravv, ForkJoinPool simdExecutor) {
+        var cv = simdExecutor.submit(() -> IntStream.range(0, ravv.size())
                 .parallel()
                 .mapToObj(i -> encode(ravv.getVector(i)))
                 .toArray(long[][]::new))
                 .join();
+        return new BQVectors(this, cv);
     }
 
     /**
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/BinaryQuantization.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/BinaryQuantization.java
@@ -66,7 +66,12 @@ public class BinaryQuantization implements VectorCompressor<long[]> {
     public CompressedVectors encodeAll(RandomAccessVectorValues ravv, ForkJoinPool simdExecutor) {
         var cv = simdExecutor.submit(() -> IntStream.range(0, ravv.size())
                 .parallel()
-                .mapToObj(i -> encode(ravv.getVector(i)))
+                .mapToObj(i -> {
+                    var vector = ravv.getVector(i);
+                    return vector == null
+                            ? new long[compressedVectorSize() / Long.BYTES]
+                            : encode(vector);
+                })
                 .toArray(long[][]::new))
                 .join();
         return new BQVectors(this, cv);
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/BinaryQuantization.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/BinaryQuantization.java
@@ -81,7 +81,13 @@ public class BinaryQuantization implements VectorCompressor<long[]> {
     public long[] encode(VectorFloat<?> v) {
         int M = (int) Math.ceil(v.length() / 64.0);
         long[] encoded = new long[M];
-        for (int i = 0; i < M; i++) {
+        encodeTo(v, encoded);
+        return encoded;
+    }
+
+    @Override
+    public void encodeTo(VectorFloat<?> v, long[] dest) {
+        for (int i = 0; i < dest.length; i++) {
             long bits = 0;
             for (int j = 0; j < 64; j++) {
                 int idx = i * 64 + j;
@@ -92,9 +98,8 @@ public class BinaryQuantization implements VectorCompressor<long[]> {
                     bits |= 1L << j;
                 }
             }
-            encoded[i] = bits;
+            dest[i] = bits;
         }
-        return encoded;
     }
 
     @Override
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
@@ -256,6 +256,8 @@ public class PQVectors implements CompressedVectors {
     }
 
     public ByteSequence<?> get(int ordinal) {
+        if (ordinal < 0 || ordinal >= vectorCount)
+            throw new IndexOutOfBoundsException("Ordinal " + ordinal + " out of bounds for vector count " + vectorCount);
         return get(compressedDataChunks, ordinal, vectorsPerChunk, pq.getSubspaceCount());
     }
 
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
@@ -17,6 +17,7 @@
 package io.github.jbellis.jvector.pq;
 
 import io.github.jbellis.jvector.disk.RandomAccessReader;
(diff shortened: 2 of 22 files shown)


michaeljmarshall:
@jbellis @jkni @marianotepper - this is ready for review. The other alternative, described above, where we use a ByteBuffer instead of creating an `ArraySliceByteSource` seems viable, but comes down to preference. Let me know what you think.

jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
```diff
@@ -77,44 +77,76 @@ public static PQVectors load(RandomAccessReader in) throws IOException {
...
+        }
+        return true;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(pq, compressedVectors);
+        // We don't use the array structure in the hash code calculation because we allow for different chunking
+        // strategies. Instead, we use the first entry in the first chunk to provide a stable hash code.
+        return Objects.hash(pq, count(), count() > 0 ? get(0).get(0) : 0);
```
jbellis: If you're worried about the performance of hashing millions of vectors, I'd prefer adding the first codepoint from each vector, to using all the first vector's codepoints

jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
```diff
@@ -77,44 +77,76 @@ public static PQVectors load(RandomAccessReader in) throws IOException {
...
         if (o == null || getClass() != o.getClass()) return false;
 
         PQVectors that = (PQVectors) o;
         if (!Objects.equals(pq, that.pq)) return false;
-        return Objects.equals(compressedVectors, that.compressedVectors);
+        if (this.count() != that.count()) return false;
+        // TODO how do we want to determine equality? With the current change, we are willing to write one
+        // thing and materialize another. It seems like the real concern should be whether the compressedVectors have
+        // the same data, not whether they are in a MemorySegment or a byte[] and not whether there is one chunk of many
+        // vectors or many chunks of one vector. This technically goes against the implementation of each of the
```
jbellis: I'm fine with changing the ByteSequence equals implementations to match.  You get the class comparison by default when intellij generates it for you and I think we just ran with that.

jbellis:
I like where this is going, I'm just concerned that it doesn't break compatibility enough. :)

That is, the PQVectors API is hard to use "correctly" now since the only method that gives you One Big Array is load().  (Technically the constructor used by load is public but everyone else just calls it with an array of one vector per chunk.)

I think we should push the changes into VectorCompressor to solve this.  Instead of introducing a new class to abstract long[] vs ByteSequence, encodeAll should return CompressedVectors directly.  (By far the most common way encodeAll is used is to immediately wrap the result in CV anyway.)  I think there's a code path or two where you don't need the full CV but it's not a big deal to construct one for those cases.

CompressedVectors is not generic so VC wouldn't have to be either, which makes me happy.

jbellis:
I'm 80% sure that it's correct to go ahead and change the API and release 3.1 instead of doing unnatural contortions to maintain compatibility.

jbellis:
Also I do not insist that BQ change to the chunking approach.  If we can conform to the "right" API without doing that, it's fine to leave optimizing BQ until somebody needs it.

michaeljmarshall:
@jbellis - how would this work in the `CompactionGraph`? I left the hybrid creation method in place because it seemed like we needed it for that class, but perhaps we just need to add an `add` method to the `PQVectors` instead of letting them mutate the ArrayList (the current design).

jbellis:
Yes, I think an add() or a set() method would be straightforward since CG knows up front how many it's going to generate.  This is a better design than leaking the underlying List to be mutated, but that was the path of least resistance at one point!

jbellis:
Is it possible to unify encode/set to `void encodeTo(VectorFloat, CompressedVectors, int i)` to avoid the copy?

michaeljmarshall:
I had started to look at that, but stopped because of our logic to for `encodedOmittedVector`. Turns out that is just setting a zero vector, so as long as we have an on-heap data structure, we're good to go, but we might not want to assume that for long since an off heap MemorySegment looks like it might have undefined values for regions that are not explicitly set.

Getting the change written up now.

jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
michaeljmarshall: I implemented this, but it's not as elegant as I wanted. I could have made the ByteSequence interface an abstract class and overridden equals/hashCode, but that seemed like it went too far, so I added default methods that implementations of the interface can call. In practice, I doubt these objects are compared for strict equality  frequently.

jvector-base/src/main/java/io/github/jbellis/jvector/pq/PQVectors.java
michaeljmarshall: I ended up just hashing based on every value since that is the norm. If it becomes a performance bottleneck, we can address it then--I doubt we compute the hash code frequently.

michaeljmarshall:
Note: I'll update the PR description tomorrow to align with recent commits.

jbellis:
LGTM

do you mind updating UPGRADING and README?

jvector-base/src/main/java/io/github/jbellis/jvector/pq/BinaryQuantization.java
```diff
@@ -63,12 +63,18 @@ public CompressedVectors createCompressedVectors(Object[] compressedVectors) {
     }
 
     @Override
-    public long[][] encodeAll(RandomAccessVectorValues ravv, ForkJoinPool simdExecutor) {
-        return simdExecutor.submit(() -> IntStream.range(0, ravv.size())
+    public CompressedVectors encodeAll(RandomAccessVectorValues ravv, ForkJoinPool simdExecutor) {
+        var cv = simdExecutor.submit(() -> IntStream.range(0, ravv.size())
                 .parallel()
```
eolivelli, line 68: out of curiosity:
does it really help to use "parallel" here ?

jbellis: Massively.

jbellis:
LGTM, merged

Thoughts for if/when we revisit this:
1. I don't love PQVectors being mutable or not based on which constructor was called.  Would factory methods improve the clarity?  Separate classes?
2. Could we eliminate ArrayByteSequence and just make everything an ArraySliceBS?  I don't think there's a path left where we care about the overhead from the extra fields and I'd rather have a single implementation instead of having to make fixes like setLittleEndianShort in two places.

PR #372, opened 2024-12-03
Don't use segment hashCode in MemorySegmentVectorFloat

jkni:
This hashcode depends on segment base/offset in the heap rather than contents. This breaks testing around PQVectors hashcodes added in #370. Rather than this being a regression uncovered in #370, this is due to increased test assertions added in #370 that only fail when using the native provider. We adopt the same approach used by ByteSequence in #370.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorFloat.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorFloat.java
@@ -114,7 +114,7 @@ final public class ArrayVectorFloat implements VectorFloat<float[]>
     @Override
     public int hashCode()
     {
-        return Arrays.hashCode(data);
+        return this.getHashCode();
     }
 }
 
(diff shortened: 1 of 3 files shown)


michaeljmarshall:
I've been thinking of MemorySegment as a more flexible ByteBuffer, but it looks like the hashcode implementations diverge. Here is ByteBuffer's:

```java
    /**
     * Returns the current hash code of this buffer.
     *
     * <p> The hash code of a byte buffer depends only upon its remaining
     * elements; that is, upon the elements from {@code position()} up to, and
     * including, the element at {@code limit()}&nbsp;-&nbsp;{@code 1}.
     *
     * <p> Because buffer hash codes are content-dependent, it is inadvisable
     * to use buffers as keys in hash maps or similar data structures unless it
     * is known that their contents will not change.  </p>
     *
     * @return  The current hash code of this buffer
     */
    public int hashCode() {
        int h = 1;
        int p = position();
        for (int i = limit() - 1; i >= p; i--)



            h = 31 * h + (int)get(i);

        return h;
    }
```

PR #374, opened 2024-12-18
Non-uniform vector quantization

marianotepper:
This PR adds a new feature that uses quantized vectors for re-ranking. The quantization is computed on subvectors of each vector in the index, using a non-uniform quantizer.


jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NVQ.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    ScoreFunction.ExactScoreFunction rerankerFor(VectorFloat<?> queryVector,
+                                                 VectorSimilarityFunction vsf,
+                                                 FeatureSource source) {
+        var function = scorer.scoreFunctionFor(queryVector, vsf);
+
+        return new ScoreFunction.ExactScoreFunction() {
+            private final QuantizedVector scratch = NVQuantization.QuantizedVector.createEmpty(nvq.subvectorSizesAndOffsets, nvq.bitsPerDimension);
```
jkni: In most other places like this, we use a reusable thread-local scratch. I'm not sure if it's worth it, so consider this a possible deferred optimization.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -324,6 +324,8 @@ public OnDiskGraphIndexWriter build() throws IOException {
             int dimension;
             if (features.containsKey(FeatureId.INLINE_VECTORS)) {
                 dimension = ((InlineVectors) features.get(FeatureId.INLINE_VECTORS)).dimension();
+            } else if (features.containsKey(FeatureId.NVQ_VECTORS)) {
+                dimension = ((NVQ) features.get(FeatureId.NVQ_VECTORS)).dimension();
             } else {
                 throw new IllegalArgumentException("Inline vectors must be provided.");
```
jkni: We should update this error message to indicate that either inline vectors or NVQ vectors much be provided.

jvector-base/src/main/java/io/github/jbellis/jvector/optimization/LossFunction.java
```diff
@@ -0,0 +1,137 @@
...
+    private float[] minBounds;
+    private float[] maxBounds;
+
+    /**
+     * Constructs a LossFunction acting on vectors of the specified number of dimensions.
+     * @param nDims the number of dimensions
+     */
+    public LossFunction(int nDims) {
+        if (nDims <= 0) {
+            throw new IllegalArgumentException("The standard deviation initSigma must be positive");
```
jkni: bad error message copy/paste

jvector-base/src/main/java/io/github/jbellis/jvector/optimization/LossFunction.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+        minBounds = new float[nDims];
+        maxBounds = new float[nDims];
+        for (int d = 0; d < nDims; d++) {
+            minBounds[d] = Float.NEGATIVE_INFINITY;
```
jkni: Could use Arrays.fill here

jvector-base/src/main/java/io/github/jbellis/jvector/optimization/NESOptimizer.java
```diff
@@ -0,0 +1,323 @@
...
+ * the feasible set.
+ * <p>
+ * [1]
+ * Wierstra, Schaul, Glasmachers, Sun, Peters, and Schmidhuber
+ * Natural Evolution Strategies
+ * <a href="https://www.jmlr.org/papers/volume15/wierstra14a/wierstra14a.pdf">Link</a>
+ */
+public class NESOptimizer {
+    public enum Distribution {
+        MULTINORMAL, // This is for adding future support for the multinormal case (Algorithm 5 in [1])
```
jkni: Do we think we'll get back to MULTINORMAL relatively soon? if not, is this worth cleaning up for now?

jvector-base/src/main/java/io/github/jbellis/jvector/optimization/NESOptimizer.java
```diff
@@ -0,0 +1,323 @@
...
+    public void setMaxIterations(int maxIterations) {
+        this.maxIterations = maxIterations;
+    }
+
+    /**
+     * NES utilizes rank-based fitness shaping in order to render the algorithm invariant under
+     * monotonically increasing (i.e., rank preserving) transformations of the loss function.
+     * For this, the samples losses are transformed into a set of utility values using the method in Section 3.1 of [1].
+     *
+     * @param nSamples the loss function values for each sample
```
jkni: bad comment -- copy/paste from funValues below

jvector-base/src/main/java/io/github/jbellis/jvector/optimization/NESOptimizer.java
```diff
@@ -0,0 +1,323 @@
...
+     *
+     * @param maxIterations the maximum number of iterations
+     */
+    public void setMaxIterations(int maxIterations) {
+        this.maxIterations = maxIterations;
+    }
+
+    /**
+     * NES utilizes rank-based fitness shaping in order to render the algorithm invariant under
+     * monotonically increasing (i.e., rank preserving) transformations of the loss function.
```
jkni: also a copy-paste from below -- could be clearer that this computes the set of utilities

jvector-base/src/main/java/io/github/jbellis/jvector/optimization/NESOptimizer.java
```diff
@@ -0,0 +1,323 @@
...
+
+        var nSamples = computeNSamples(nDims);
+        var lrSigma = computeLrSigma(nDims);
+
+        // Initialize mu and sigma
+        var sample = new float[nDims];
+        var mu = new float[nDims];
+        var sigma = new float[nDims];
+        for (int d = 0; d < nDims; d++) {
+            mu[d] = initialSolution[d];
```
jkni: could use Arrays.copyOf/Arrays.fill here again

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQScorer.java
```diff
@@ -0,0 +1,146 @@
...
+
+import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
+import io.github.jbellis.jvector.vector.VectorUtil;
+import io.github.jbellis.jvector.vector.types.VectorFloat;
+
+public class NVQScorer {
+    final NVQuantization nvq;
+
+    /**
+     * Initialize the NVQ Vectors with an initial List of vectors.  This list may be
```
jkni: This comment seems wrong. Maybe an early version of the NVQVectors comment that was copy/pasted?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+            }
+        },
+        FOUR {
+            @Override
+            public int getInt() {
+                return 4;
+            }
+
+            @Override
+            public void write(DataOutput out) throws IOException {
```
jkni: don't think we need this override

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+    public final VectorFloat<?> globalMean;
+
+    // The number of dimensions of the original (uncompressed) vectors:
+    public final int originalDimension;
+
+    // A matrix that stores the size and starting point of each subvector:
+    public final int[][] subvectorSizesAndOffsets;
+
+    // Whether we want to skip the optimization of the NVQ parameters. Here for debug purposes only.
+    public boolean learn = true;
```
jkni, line 120: Let's use the @VisibleForTesting annotation here so that IDEs surface the reason for public visibility

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+     * Computes the global mean vector and the data structures used to divide each vector into subvectors.
+     *
+     * @param ravv the vectors to quantize
+     * @param nSubVectors number of subvectors
+     */
+    public static NVQuantization compute(RandomAccessVectorValues ravv, int nSubVectors) {
+        var subvectorSizesAndOffsets = getSubvectorSizesAndOffsets(ravv.dimension(), nSubVectors);
+
+        var ravvCopy = ravv.threadLocalSupplier().get();
+        // convert ravvCopy to list
```
jkni: stale comment?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+        var dim = ravvCopy.getVector(0).length();
+        var globalMean = vectorTypeSupport.createFloatVector(dim);
+        globalMean.zero();
```
jkni: We shouldn't need a zero-ing here. `createFloatVector(dim)` returns a zero-vector by contract

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+    public CompressedVectors createCompressedVectors(Object[] compressedVectors) {
+        return new NVQVectors(this, (QuantizedVector[]) compressedVectors);
+    }
+
+    /**
+     * Encodes the given vectors in parallel using NVQ.
+     */
+    @Override
+    public NVQVectors encodeAll(RandomAccessVectorValues ravv, ForkJoinPool simdExecutor) {
+        return new NVQVectors(this,
```
jkni, line 170: It looks like we actually make this same mistake elsewhere, but I don't think this usage of `ravv` is safe by contract. RandomAccessVectorValues is permitted to reuse the same object across multiple calls to `getVector`, so having multiple threads call `getVector` in parallel is not safe. We should use the same ravv copy code structure as `ProductQuantization.extractTrainingVector`

jkni: Generally dislike pushing somewhat unrelated changes, but I think we should fix this for PQ/BQ encodeAll at the same time.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+                        .join());
+    }
+
+    /**
+     * Encodes the input vector using NVQ.
+     * @return one subvector per subspace
+     */
+    @Override
+    public QuantizedVector encode(VectorFloat<?> vector) {
+        var tempVector = VectorUtil.sub(vector, globalMean);
```
jkni: Maybe rewrite this to just create qv and call `encodeTo(vector, qv)`

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+    }
+
+    /**
+     * Writes the instance to a DataOutput.
+     * @param out DataOutput to write to
+     * @param version serialization version.
+     * @throws IOException fails if we cannot write to the DataOutput
+     */
+    public void write(DataOutput out, int version) throws IOException
+    {
```
jkni, line 246: not sure whether we need the legacy baggage here as we don't need to support any backwards compatibility. IMO, we can skip the magic and just always start by writing the version. We needed the MAGIC trick for other serializations because we didn't initially version them and had to detect version-less serializations.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+        return size;
+    }
+
+    /**
+     * Loads an instance from a RandomAccessReader
+     * @param in the RandomAccessReader
+     * @return the instance
+     * @throws IOException fails if we cannot read from the RandomAccessReader
+     */
+    public static NVQuantization load(RandomAccessReader in) throws IOException {
```
jkni, line 287: same as save, don't think this needs the MAGIC trick

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+        }
+
+
+        /**
+         * Write the instance to a DataOutput
+         * @param out the DataOutput
+         * @throws IOException fails if we cannot write to the DataOutput
+         */
+        public void write(DataOutput out) throws IOException {
+            // write the min and max
```
jkni: stale comment?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -0,0 +1,685 @@
...
+                    tempSolution[0] = gr;
+                    float lossValue = lossFunction.compute(tempSolution);
+                    if (lossValue > bestLossValue) {
+                        bestLossValue = lossValue;
+                        growthRateFineTuned = gr;
+                    }
+                }
+
+                growthRate = growthRateFineTuned;
+                midpoint = 0.f;
```
jkni: Redundant? It looks like midpoint is initialized at 0 and then assigned 0

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQVectors.java
```diff
@@ -0,0 +1,153 @@
...
+
+    @Override
+    public int count() {
+        return compressedVectors.length;
+    }
+
+    @Override
+    public void write(DataOutput out, int version) throws IOException
+    {
+        // serialization version
```
jkni: serialization version here seems too terse as a comment -- we're serializing nvq at the given version

jvector-base/src/main/java/io/github/jbellis/jvector/vector/DefaultVectorUtilSupport.java
```diff
@@ -328,29 +351,195 @@ public void quantizePartials(float delta, VectorFloat<?> partials, VectorFloat<?
...
+
+    int temp = Float.floatToIntBits(z);
+    int e = temp & 0x7f800000;
+    float p = (float) ((e >> 23) - 128);
+    float m = Float.intBitsToFloat((temp & 0x007fffff) + 0x3f800000);
+
+    return Math.fma(m + p, inverseAlpha, x0);
+  }
+
+  static float scaled_logistic_function(float value, float growthRate, float midpoint, float logisticScale, float logisticBias) {
```
jkni: remove snakecase

jvector-base/src/main/java/io/github/jbellis/jvector/vector/DefaultVectorUtilSupport.java
```diff
@@ -328,29 +351,195 @@ public void quantizePartials(float delta, VectorFloat<?> partials, VectorFloat<?
...
+    var y = logisticFunctionNQT(value, growthRate, midpoint);
+    return (y - logisticBias) * (1 / logisticScale);
+  }
+
+  static float scaledLogitFunctionNQT(float value, float inverseGrowthRate, float midpoint, float logisticScale, float logisticBias) {
+    var scaledValue = Math.fma(value, logisticScale, logisticBias);
+//    return Math.fma(MathUtil.fastLog(scaledValue / (1 - scaledValue)), inverseGrowthRate, midpoint);
```
jkni: remove commented out dead code

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtil.java
```diff
@@ -180,7 +192,7 @@ public static void quantizePartials(float delta, VectorFloat<?> partials, Vector
   /**
    * Calculates the maximum value in the vector.
    * @param v vector
-   * @return the maximum value, or -Float.MAX_VALUE if the vector is empty
+   * @return the maximum value, or -Float.maxValue if the vector is empty
```
jkni: bad find/replace or something? We want MAX_VALUE

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtil.java
```diff
@@ -189,7 +201,7 @@ public static float max(VectorFloat<?> v) {
   /**
    * Calculates the minimum value in the vector.
    * @param v vector
-   * @return the minimum value, or Float.MAX_VALUE if the vector is empty
+   * @return the minimum value, or Float.maxValue if the vector is empty
```
jkni: Same as above

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtilSupport.java
```diff
@@ -213,4 +222,86 @@ default float pqDecodedCosineSimilarity(ByteSequence<?> encoded, int clusterCoun
...
+   * Compute the squared error of quantizing the vector with NVQ.
+   * All values of the vector must be in [0, 1]. For example, the input vector has been
+   * normalized by subtracting the bias and dividing by the scale.
+   * @param growthRate The growth rate of the logistic function
+   * @param midpoint the midpoint of the logistic function
+   * @param minValue The minimum value of the subvector
+   * @param maxValue The maximum value of the subvector
+   * @param nBits the number of bits per dimension
+   */
+  float nvqLoss(VectorFloat<?> vector, float growthRate, float midpoint, float minValue, float maxValue, int nBits);
```
jkni, line 293: missing vector param in the javadoc

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtilSupport.java
```diff
@@ -213,4 +222,86 @@ default float pqDecodedCosineSimilarity(ByteSequence<?> encoded, int clusterCoun
...
+
+  /**
+   * Compute the squared error of quantizing the vector with a uniform quantizer.
+   * All values of the vector must be in [0, 1]. For example, the input vector has been
+   * normalized by subtracting the bias and dividing by the scale.
+   * @param minValue The minimum value of the subvector
+   * @param maxValue The maximum value of the subvector
+   * @param nBits the number of bits per dimension
+   */
+  float nvqUniformLoss(VectorFloat<?> vector, float minValue, float maxValue, int nBits);
```
jkni, line 302: missing vector param in the javadoc

jvector-examples/src/main/java/io/github/jbellis/jvector/example/distancesNVQ.java
```diff
@@ -0,0 +1,153 @@
...
+import io.github.jbellis.jvector.vector.VectorUtil;
+import io.github.jbellis.jvector.vector.types.VectorFloat;
+
+import java.io.IOException;
+import java.util.*;
+
+import static java.lang.Math.abs;
+
+// this class uses explicit typing instead of `var` for easier reading when excerpted for instructional use
+public class distancesNVQ {
```
jkni: Rename to DistancesNVQ

jvector-examples/src/main/java/io/github/jbellis/jvector/example/SiftSmall.java
```diff
@@ -319,11 +379,12 @@ public static void main(String[] args) throws IOException {
         System.out.format("%d base and %d query vectors loaded, dimensions %d%n",
                           baseVectors.size(), queryVectors.size(), baseVectors.get(0).length());
 
-        siftInMemory(baseVectors);
-        siftInMemoryWithSearcher(baseVectors);
-        siftInMemoryWithRecall(baseVectors, queryVectors, groundTruth);
-        siftPersisted(baseVectors, queryVectors, groundTruth);
-        siftDiskAnn(baseVectors, queryVectors, groundTruth);
+//        siftInMemory(baseVectors);
```
jkni: uncomment?

jvector-native/src/main/java/io/github/jbellis/jvector/vector/VectorSimdOps.java
```diff
@@ -588,4 +645,351 @@ public static void quantizePartials(float delta, MemorySegmentVectorFloat partia
             }
         }
     }
+
+    //---------------------------------------------
+    // NVQ quantization instructions start here
+    //---------------------------------------------
+    static FloatVector const1f = FloatVector.broadcast(FloatVector.SPECIES_PREFERRED, 1.f);
```
jkni: we should make these `static final` - the jdk.incubator.vector docs suggest vectors should only be in static final constants, locals, and parameters

jvector-native/src/main/java/io/github/jbellis/jvector/vector/VectorSimdOps.java
```diff
@@ -588,4 +645,351 @@ public static void quantizePartials(float delta, MemorySegmentVectorFloat partia
...
+                temp = arr.get(first + a);
+                arr.set(first + a, arr.get(cycle));
+                arr.set(cycle, temp);
+                visited[a] = true;
+            } while ((first + a) != cycle);
+        }
+    }
+
+    static void nvqShuffleQueryInPlace8bit(MemorySegmentVectorFloat vector) {
+        // To understand this shuffle, see nvqDequantize8bit
```
jkni, line 1005: Can we expound on the shuffle a bit more in nvqDequantize8bit? I'm not sure it's self-explanatory enough

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtilSupport.java
```diff
@@ -213,4 +222,86 @@ default float pqDecodedCosineSimilarity(ByteSequence<?> encoded, int clusterCoun
...
+   * @param midpoint the midpoint of the logistic function
+   * @param minValue The minimum value of the subvector
+   * @param maxValue The maximum value of the subvector
+   * @param centroid the global mean vector used to re-center the quantized subvectors.
+   * @return the cosine similarity
+   */
+  float[] nvqCosine8bit(VectorFloat<?> vector, ByteSequence<?> bytes, float growthRate, float midpoint, float minValue, float maxValue, VectorFloat<?> centroid);
+
+  /**
+   * When using 4-bit NVQ quantization and vector instructions, it is easier to unpack all even entries, and then all
```
jkni: Why the reference to 4-bit quantization in 8bit javadoc?

jvector-tests/src/test/java/io/github/jbellis/jvector/optimization/TestNaturalEvolutionStrategies.java
```diff
@@ -0,0 +1,116 @@
...
+            public float compute(float[] x) {
+                float sum = 0;
+                for (float num : x) {
+                    sum -= MathUtil.square(num);
+                }
+                return sum;
+            }
+        }
+
+        {
```
jkni: introducing the new variable scopes here is perfectly fine, but IMO, would also be fine to put this all in one scope and re-use the variables. There are definitely codebases where this would be more idiomatic, but I can't think of other places we do this in JVector. Just thinking out loud -- don't feel like this has to change.

jvector-tests/src/test/java/io/github/jbellis/jvector/quantization/TestCompressedVectors.java
```diff
@@ -87,7 +89,47 @@ public void testSaveLoadBQ() throws Exception {
         }
     }
 
-    private void testEncodings(int dimension, int codebooks) {
+    @Test
+    public void testSaveLoadNVQ() throws Exception {
+
+        int[][] testsConfigAndResults = {
+                //Tuples of: nDimensions, nSubvectors, number of bots per dimension, and the expected number of bytes
```
jkni: number of bots per dimension can be removed here (and also, typo, but it will get removed anyway)

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/SimdOps.java
```diff
@@ -742,4 +797,355 @@ public static float pqDecodedCosineSimilarity256(ByteSequence<byte[]> baseOffset
 
         return (float) (sumResult / Math.sqrt(aMagnitudeResult * bMagnitude));
     }
+
+    //---------------------------------------------
+    // NVQ quantization instructions start here
+    //---------------------------------------------
+
+    static FloatVector const1f = FloatVector.broadcast(FloatVector.SPECIES_PREFERRED, 1.f);
```
jkni: same static final discussion here as in VectorSimdOps

jvector-base/src/main/java/io/github/jbellis/jvector/vector/VectorUtilSupport.java
```diff
@@ -213,4 +222,86 @@ default float pqDecodedCosineSimilarity(ByteSequence<?> encoded, int clusterCoun
...
+   * uneven entries within register. This method shuffles the query entries so that it matches this order.
+   * See <a href="https://www.vldb.org/pvldb/vol16/p2132-afroozeh.pdf">this reference</a>
+   * @param vector the vector to be shuffled
+   */
+  void nvqShuffleQueryInPlace8bit(VectorFloat<?> vector);
+
+  /**
+   * Quantize a subvector as an 8-bit quantized subvector.
+   * All values of the vector must be in [0, 1]. For example, the input vector has been
```
jkni: I'm confused by `All values of the vector...` and then the references to bias/scale in this/nvqLoss/nvqUniformLoss. Are those remnants of a previous approach?

marianotepper:
The changes requested/suggested by @jkni have been put in place.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NVQ.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    ScoreFunction.ExactScoreFunction rerankerFor(VectorFloat<?> queryVector,
+                                                 VectorSimilarityFunction vsf,
+                                                 FeatureSource source) {
+        var function = scorer.scoreFunctionFor(queryVector, vsf);
+
+        return new ScoreFunction.ExactScoreFunction() {
+            private final ThreadLocal<QuantizedVector> scratch = ThreadLocal.withInitial(() -> NVQuantization.QuantizedVector.createEmpty(nvq.subvectorSizesAndOffsets, nvq.bitsPerDimension));
```
jkni: if we turn this into a ThreadLocal (as suggested in my initial review), we should turn it into a top-level ExplicitThreadLocal of the NVQ feature rather than the reranker. The goal here is to reduce repeated allocations of the scratch field by tying it to the lifecycle of a given index object rather than the ExactScoreFunction instance. Since a given thread will only need one instance of an ExactScoreFunction at once for a given NVQ Feature, this is a reasonable contract to establish, as we have with FusedADC feature. A reference here would be `reusableNeighbors` in FusedADC.

jbellis:
(Please use "squash and merge" when you commit.)

marianotepper:
README.md and UPGRADING.md are updated now.

xjtushilei:
May I ask a question? I'm curious about the comparison between NVQ and SQ8. SQ8 uses 1/4 of the original vector size, which is acceptable in terms of precision loss during the second stage, but I'm wondering how much better NVQ is compared to SQ8?

PR #379, opened 2024-12-24
Fix regression in assembleAndSum PQ decoder performance

jkni:
Following #368, performance of the assembling PQ fragments suffers on 128-bit SIMD platforms from simulating vector operations. I first attempted implementing an 128-bit version, which improved performance relative to post-#368, but it did not restore performance to pre-#368 levels. As a result, I switched back to the scalar implementation on 128-bit platforms.


marianotepper (approved):
Looks good!

PR #387, opened 2025-01-15
Improved use ScoreTracker to avoid wasteful searching for very large k

marianotepper:
This improves upon #384 by making the quantiles estimation more lightweight. It models the recent scores as a Normal distribution and uses incremental updates to track sufficient statistics of its mean and variance. Then, quantiles are computed from these statistics.


jbellis:
>     * This implementation does not consider the worstBestScore provided to shouldStop.

I think this comment must be left over from earlier changes?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ScoreTracker.java
```diff
@@ -93,8 +93,106 @@ public boolean shouldStop() {
...
+
+        // the sample variance multiplied by n-1
+        private double dSquared;
+
+        /**
+         * Constructor
+         * @param bestScoredTracked the number of tracked scores used to estimate if we are unlikely to improve
+         *                          the results anymore. An empirical rule of thumb is bestScoredTracked=rerankK.
+         */
+        RelaxedMonotonicityTracker(int bestScoredTracked) {
```
jbellis: typo in `bestScoresTracked`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ScoreTracker.java
```diff
@@ -93,8 +93,106 @@ public boolean shouldStop() {
...
+
+            // We're in phase 2 if the q-th percentile of the recent scores evaluated,
+            //     mean + SIGMA_FACTOR * sqrt(variance),
+            // is lower than the worst of the best scores seen.
+            // (paper suggests using the median of recent scores, but experimentally that is too prone to false positives)
+            double std = Math.sqrt(this.dSquared / (this.recentScores.length - 1));
+            double windowPercentile = this.mean + SIGMA_FACTOR * std;
+            double worstBestScore = sortableIntToFloat((int) bestScores.top());
+            return windowPercentile < worstBestScore;
+//            return false;
```
jbellis: cleanup

jbellis:
How does accuracy look for much smaller k/rrk?  Like 5/10 or 10/20?

marianotepper:
> How does accuracy look for much smaller k/rrk? Like 5/10 or 10/20?

Will look into this. We should consider whether in those cases it is worth applying this technique.

marianotepper:
The exploration savings are consistent for 5/10 or 10/20, but smaller. In the order of 5-10%

jbellis (approved):

PR #388, opened 2025-01-21
Change variable names to improve readability

marianotepper:
Switch usages of  "ef" to "overquery"

### DIFF
--- a/jvector-examples/README.md
+++ b/jvector-examples/README.md
@@ -73,7 +73,7 @@ RESULT [2,1,0]
   
   No spaces are allowed inside vector brackets.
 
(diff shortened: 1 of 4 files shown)


tlwillke:
There are other instances of efSearchGrid and efConstructionGrid in Bench, as well as efConstruction in IPCService and the README.  Should we change these as well?

jbellis:
efConstruction is search depth, not overquery

marianotepper:
"efConstruction is search depth, not overquery" all right. Should we use searchDepthConstruction instead of efConstruction then?

jbellis:
WFM

tlwillke (approved):
Looks good to me.

PR #389, opened 2025-01-22
Fix distance computations in Native provider

### LINKED ISSUE #385: AssertionError in testNVQEncodings
```
java.lang.AssertionError: 0.004292624942636493 > 0.002 for DOT_PRODUCT with 256 dimensions and 1 subvectors
```

reproduces close to 100% of the time for me

marianotepper:
Fixes #385, which was caused by not using MemorySegmentVectorFloat.offset to compute the correct offset when loading a vector with FloatVector.fromMemorySegment. Also fixed a couple of bugs in the tail computations.


marianotepper:
I opened #390 I will take care of it.

PR #394, opened 2025-02-13
Fix BQVectors#ramBytesUsed and BQVectors#getCompressedSize on empty BQVectors

jkni:
These can throw NPEs if they are called on a MutableBQVectors instance that hasn't yet had vectors added.


jbellis (approved):
LGTM

Issue #399, opened 2025-03-04
Test TestOnDiskGraphIndex.testReorderingWithHoles fails


marianotepper:
In line 182, onDiskGraph.getIdUpperBound() returns the size (3), when it should return the first ordinal greater than all node ids in the graph, i.e., 11.

marianotepper:
Solved by #400

PR #400, opened 2025-03-07
Fix TestOnDiskGraphIndex.testReorderingWithHoles

marianotepper:
Solves #399 and adds an additional assertion in the test to verify the corrected OnDiskGraphIndex.getNodes.

As part of the fix, I had to serialize the idUpperBound. The side effect is that OnDiskGraphIndex is in version 4 now.


jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -120,9 +122,33 @@ public int maxDegree() {
...
-    {
-        return NodesIterator.fromPrimitiveIterator(IntStream.range(0, size).iterator(), size);
+    public int getIdUpperBound() {
+        return idUpperBound;
+    }
+
+    @Override
+    public NodesIterator getNodes() {
+        try {
+            var reader = readerSupplier.get();
```
jkni: We should close this reader. I'd put it in try-with-resources like we do in `load`.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -120,9 +122,33 @@ public int maxDegree() {
...
+                                + inlineBlockSize // inline elements
+                                + (Integer.BYTES * (long) (maxDegree + 1)) // neighbor count + neighbors)
+                        ));
+                reader.seek(node_offset);
+                if (reader.readInt() != -1) {
+                    valid_nodes[pos++] = node;
+                }
+            }
+            return new NodesIterator.ArrayNodesIterator(valid_nodes, size);
+
```
jkni: clean up whitespace here

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
jkni, line 1: The IntStream import is unused now and can be removed.

PR #402, opened 2025-03-13
Add hierarchical structure to the graph index

marianotepper:
This PR introduces a hybrid HNSW-DiskANN graph. From HNSW, we take the idea of using multiple layers. This adds robustness to particularly hard datasets. Each layer  is a Vamana graph. The upper layers reside in memory while the base layer resides on disk (in DiskANN style). It also enable using a single layer. In that case, it is a plain Vamana graph.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,17 +10,19 @@ There are two broad categories of ANN index:
 
 Graph-based indexes tend to be simpler to implement and faster, but more importantly they can be constructed and updated incrementally.  This makes them a much better fit for a general-purpose index than partitioning approaches that only work on static datasets that are completely specified up front.  That is why all the major commercial vector indexes use graph approaches.
 
-JVector is a graph index in the DiskANN family tree.
+JVector is a graph index that takes a hybrid merging the the DiskANN and HNSW family trees.
+JVector borrows the hierarchical structure from HNSW, and uses Vamana (the algorithm behind DiskANN) within each layer.
 
 
 ## JVector Architecture
 
-JVector is a graph-based index that builds on the DiskANN design with composeable extensions.
+JVector is a graph-based index that builds on the HNSW anD DiskANN designs with composable extensions.
 
-JVector implements a single-layer graph with nonblocking concurrency control, allowing construction to scale linearly with the number of cores:
+JVector implements a multi-layer graph with nonblocking concurrency control, allowing construction to scale linearly with the number of cores:
 ![JVector scales linearly as thread count increases](https://github.com/jbellis/jvector/assets/42158/f0127bfc-6c45-48b9-96ea-95b2120da0d9)
 
-The graph is represented by an on-disk adjacency list per node, with additional data stored inline to support two-pass searches, with the first pass powered by lossily compressed representations of the vectors kept in memory, and the second by a more accurate representation read from disk.  The first pass can be performed with
+The upper layers of the hierarchy are represnted by an in-memory adjacency list per node. This allows for quick navigation with no IOs.
+The bottom layer of the graph is represented by an on-disk adjacency list per node. JVector uses additional data stored inline to support two-pass searches, with the first pass powered by lossily compressed representations of the vectors kept in memory, and the second by a more accurate representation read from disk.  The first pass can be performed with
 * Product quantization (PQ), optionally with [anisotropic weighting](https://arxiv.org/abs/1908.10396)
 * [Binary quantization](https://huggingface.co/blog/embedding-quantization) (BQ)
 * Fused ADC, where PQ codebooks are transposed and written inline with the graph adjacency list
@@ -51,7 +53,7 @@ First the code:
         int originalDimension = baseVectors.get(0).length();
         // wrap the raw vectors in a RandomAccessVectorValues
         RandomAccessVectorValues ravv = new ListRandomAccessVectorValues(baseVectors, originalDimension);
-
+        
         // score provider using the raw, in-memory vectors
         BuildScoreProvider bsp = BuildScoreProvider.randomAccessScoreProvider(ravv, VectorSimilarityFunction.EUCLIDEAN);
         try (GraphIndexBuilder builder = new GraphIndexBuilder(bsp,
@@ -59,7 +61,8 @@ First the code:
                                                                16, // graph degree
                                                                100, // construction search depth
                                                                1.2f, // allow degree overflow during construction by this factor
-                                                               1.2f)) // relax neighbor diversity requirement by this factor
+                                                               1.2f, // relax neighbor diversity requirement by this factor (alpha)
+                                                               true)) // use a hierarchical index
         {
             // build the index (in memory)
             OnHeapGraphIndex index = builder.build(ravv);
@@ -86,6 +89,7 @@ Commentary:
 * For the overflow Builder parameter, the sweet spot is about 1.2 for in-memory construction and 1.5 for on-disk.  (The more overflow is allowed, the fewer recomputations of best edges are required, but the more neighbors will be consulted in every search.)
 * The alpha parameter controls the tradeoff between edge distance and diversity; usually 1.2 is sufficient for high-dimensional vectors; 2.0 is recommended for 2D or 3D datasets.  See [the DiskANN paper](https://suhasjs.github.io/files/diskann_neurips19.pdf) for more details.
 * The Bits parameter to GraphSearcher is intended for controlling your resultset based on external predicates and won’t be used in this tutorial.
+* Setting the addHierarchy parameter to true, build a multi-layer index. This approach has proven more robust in highly challenging scenarios.
 
 
 #### Step 2: more control over GraphSearcher
@@ -129,7 +133,7 @@ This is expected given the approximate nature of the index being created and the
 The code:
 ```java
         Path indexPath = Files.createTempFile("siftsmall", ".inline");
-        try (GraphIndexBuilder builder = new GraphIndexBuilder(bsp, ravv.dimension(), 16, 100, 1.2f, 1.2f)) {
+        try (GraphIndexBuilder builder = new GraphIndexBuilder(bsp, ravv.dimension(), 16, 100, 1.2f, 1.2f, true)) {
             // build the index (in memory)
             OnHeapGraphIndex index = builder.build(ravv);
             // write the index to disk with default options
@@ -218,7 +222,7 @@ Then we need to set up an OnDiskGraphIndexWriter with full control over the cons
         Path indexPath = Files.createTempFile("siftsmall", ".inline");
         Path pqPath = Files.createTempFile("siftsmall", ".pq");
         // Builder creation looks mostly the same
-        try (GraphIndexBuilder builder = new GraphIndexBuilder(bsp, ravv.dimension(), 16, 100, 1.2f, 1.2f);
+        try (GraphIndexBuilder builder = new GraphIndexBuilder(bsp, ravv.dimension(), 16, 100, 1.2f, 1.2f, true);
              // explicit Writer for the first time, this is what's behind OnDiskGraphIndex.write
              OnDiskGraphIndexWriter writer = new OnDiskGraphIndexWriter.Builder(builder.getGraph(), indexPath)
                      .with(new InlineVectors(ravv.dimension()))
@@ -259,7 +263,7 @@ Commentary:
 
 ### Less-obvious points
 
-* Embeddings models product output from a consistent distribution of vectors. This means that you can save and re-use ProductQuantization codebooks, even for a different set of vectors, as long as you had a sufficiently large training set to build it the first time around. ProductQuantization.MAX_PQ_TRAINING_SET_SIZE (128,000 vectors) has proven to be sufficiently large.
+* Embeddings models produce output from a consistent distribution of vectors. This means that you can save and re-use ProductQuantization codebooks, even for a different set of vectors, as long as you had a sufficiently large training set to build it the first time around. ProductQuantization.MAX_PQ_TRAINING_SET_SIZE (128,000 vectors) has proven to be sufficiently large.
 * JDK ThreadLocal objects cannot be referenced except from the thread that created them.  This is a difficult design into which to fit caching of Closeable objects like GraphSearcher.  JVector provides the ExplicitThreadLocal class to solve this.
 * Fused ADC is only compatible with Product Quantization, not Binary Quantization.  This is no great loss since [very few models generate embeddings that are best suited for BQ](https://thenewstack.io/why-vector-size-matters/).  That said, BQ continues to be supported with non-Fused indexes.
 * JVector heavily utilizes the Panama Vector API(SIMD) for ANN indexing and search.  We have seen cases where the memory bandwidth is saturated during indexing and product quantization and can cause the process to slow down. To avoid this, the batch methods for index and PQ builds use a [PhysicalCoreExecutor](https://javadoc.io/doc/io.github.jbellis/jvector/latest/io/github/jbellis/jvector/util/PhysicalCoreExecutor.html) to limit the amount of operations to the physical core count. The default value is 1/2 the processor count seen by Java. This may not be correct in all setups (e.g. no hyperthreading or hybrid architectures) so if you wish to override the default use the `-Djvector.physical_core_count` property, or pass in your own ForkJoinPool instance.
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ The code:
         ReaderSupplier rs = ReaderSupplierFactor.open(indexPath);
         OnDiskGraphIndex index = OnDiskGraphIndex.load(rs);
         // measure our recall against the (exactly computed) ground truth
-        Function<VectorFloat<?>, SearchScoreProvider> sspFactory = q -> SearchScoreProvider.exact(q, VectorSimilarityFunction.EUCLIDEAN, ravv);
+        Function<VectorFloat<?>, SearchScoreProvider> sspFactory = q -> SearchScoreProvider.exact(q, VectorSimilarityFunction.EUCLIDEAN, index.getView());
         testRecall(index, queryVectors, groundTruth, sspFactory);
 ```
 
--- a/README.md
+++ b/README.md
@@ -10,13 +10,13 @@ There are two broad categories of ANN index:
 
 Graph-based indexes tend to be simpler to implement and faster, but more importantly they can be constructed and updated incrementally.  This makes them a much better fit for a general-purpose index than partitioning approaches that only work on static datasets that are completely specified up front.  That is why all the major commercial vector indexes use graph approaches.
 
-JVector is a graph index that takes a hybrid merging the the DiskANN and HNSW family trees.
+JVector is a graph index that merges the DiskANN and HNSW family trees.
 JVector borrows the hierarchical structure from HNSW, and uses Vamana (the algorithm behind DiskANN) within each layer.
 
 
 ## JVector Architecture
 
-JVector is a graph-based index that builds on the HNSW anD DiskANN designs with composable extensions.
+JVector is a graph-based index that builds on the HNSW and DiskANN designs with composable extensions.
 
 JVector implements a multi-layer graph with nonblocking concurrency control, allowing construction to scale linearly with the number of cores:
 ![JVector scales linearly as thread count increases](https://github.com/jbellis/jvector/assets/42158/f0127bfc-6c45-48b9-96ea-95b2120da0d9)
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,11 +5,20 @@
   in each vector with high accuracy by first applying a nonlinear transformation that is individually fit to each
   vector. These nonlinearities are designed to be lightweight and have a negligible impact on distance computation
   performance.
+- Support for hierarchical graph indices. These new type of indices blends HNSW and DiskANN in a novel way. An
+  HNSW-like hierarchy resides in memory for quickly seeding the search. This also reduces the need for caching the
+  DiskANN graph near the entrypoint. The base layer of the hierarchy is a DiskANN-like index and inherits its
+  properties. This hierarchical structure can be disabled, ending up with just the base DiskANN layer.  
 
 ## API changes
 - MemorySegmentReader.Supplier and SimpleMappedReader.Supplier must now be explicitly closed, instead of being
   closed by the first Reader created from them.
 - OnDiskGraphIndex no longer closes its ReaderSupplier
+- The constructor of GraphIndexBuilder takes an additional parameter which allows to enable or disable the use of the
+  hierarchy.
+- GraphSearcher can be configured to run pruned searches using GraphSearcher.usePruning. When this is set to true,
+  we do early termination of the search. In certain cases, this can accelerate the search at the potential cost of some
+  accuracy. It is set to false by default.
 
 ### API changes in 3.0.6
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -19,6 +19,8 @@
 - GraphSearcher can be configured to run pruned searches using GraphSearcher.usePruning. When this is set to true,
   we do early termination of the search. In certain cases, this can accelerate the search at the potential cost of some
   accuracy. It is set to false by default.
+- The constructors of GraphIndexBuilder allow to specify different maximum out-degrees for the graphs in each layer.
+  However, this feature does not work with FusedADC in this version.
 
 ### API changes in 3.0.6
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,7 +5,7 @@
   in each vector with high accuracy by first applying a nonlinear transformation that is individually fit to each
   vector. These nonlinearities are designed to be lightweight and have a negligible impact on distance computation
   performance.
-- Support for hierarchical graph indices. These new type of indices blends HNSW and DiskANN in a novel way. An
+- Support for hierarchical graph indices. This new type of index blends HNSW and DiskANN in a novel way. An
   HNSW-like hierarchy resides in memory for quickly seeding the search. This also reduces the need for caching the
   DiskANN graph near the entrypoint. The base layer of the hierarchy is a DiskANN-like index and inherits its
   properties. This hierarchical structure can be disabled, ending up with just the base DiskANN layer.  
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
@@ -96,7 +96,7 @@ public interface GraphIndex extends AutoCloseable, Accountable {
     void close() throws IOException;
 
     /**
-     * @return The maximum (coarser) level with that contains a vector in the graph.
+     * @return The maximum (coarser) level that contains a vector in the graph.
      */
     int getMaxLevel();
 
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
@@ -172,7 +172,7 @@ public interface GraphIndex extends AutoCloseable, Accountable {
         return sb.toString();
     }
 
-    // Comparable b/c it gets used in CSLM
+    // Comparable b/c it gets used in ConcurrentSkipListMap
     final class NodeAtLevel implements Comparable<NodeAtLevel> {
         public final int level;
         public final int node;
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
@@ -115,7 +115,7 @@ public interface GraphIndex extends AutoCloseable, Accountable {
         /**
          * @return the number of nodes in the graph
          */
-        // TODO can we just remove this in favor of graph's?
+        @Deprecated
         int size();
 
         /**
@@ -148,7 +148,7 @@ public interface GraphIndex extends AutoCloseable, Accountable {
 
     static String prettyPrint(GraphIndex graph) {
         StringBuilder sb = new StringBuilder();
-        sb.append(graph); // toString looks like OnHeapGraphIndex(size=9, entryPoint=NodeAtLevel(level=4, node=6))
+        sb.append(graph);
         sb.append("\n");
 
         try (var view = graph.getView()) {
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
@@ -95,10 +95,23 @@ public interface GraphIndex extends AutoCloseable, Accountable {
     @Override
     void close() throws IOException;
 
+    /**
+     * @return The maximum (coarser) level with that contains a vector in the graph.
+     */
     int getMaxLevel();
 
+    /**
+     * Return the maximum out-degree allowed of the given level.
+     * @param level The level of interest
+     * @return the maximum out-degree of the given level
+     */
     int getDegree(int level);
 
+    /**
+     * Return the number of vectors/nodes in the given level.
+     * @param level The level of interest
+     * @return the number of vectors in the given level
+     */
     int size(int level);
 
     /**
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
@@ -113,6 +113,8 @@ public interface GraphIndex extends AutoCloseable, Accountable {
         NodesIterator getNeighborsIterator(int level, int node);
 
         /**
+         * This method is deprecated as most View usages should not need size.
+         * Where they do, they could access the graph.
          * @return the number of nodes in the graph
          */
         @Deprecated
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
@@ -152,7 +152,7 @@ public interface GraphIndex extends AutoCloseable, Accountable {
         sb.append("\n");
 
         try (var view = graph.getView()) {
-            for (int level = 0; level < graph.getMaxLevel(); level++) {
+            for (int level = 0; level <= graph.getMaxLevel(); level++) {
                 sb.append(String.format("# Level %d\n", level));
                 NodesIterator it = graph.getNodes(level);
                 while (it.hasNext()) {
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
@@ -99,7 +99,7 @@ public interface GraphIndex extends AutoCloseable, Accountable {
 
     int getDegree(int level);
 
-    int size(int i);
+    int size(int level);
 
     /**
      * Encapsulates the state of a graph for searching.  Re-usable across search calls,
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -715,6 +715,24 @@ public class GraphIndexBuilder implements Closeable {
             throw new IllegalStateException("Cannot load into a non-empty graph");
         }
 
+        int maybeMagic = in.readInt();
+        int version; // This is not used in V4 but may be useful in the future, putting it as a placeholder.
+        if (maybeMagic != OnHeapGraphIndex.MAGIC) {
+            // JVector 3 format, no magic or version, starts straight off with the number of nodes
+            version = 3;
+            int size = maybeMagic;
+            loadV3(in, size);
+        } else {
+            version = in.readInt();
+            loadV4(in);
+        }
+    }
+
+    private void loadV4(RandomAccessReader in) throws IOException {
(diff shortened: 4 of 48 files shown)


marianotepper:
A point worth noting is that we recently updated the file format for graphs to V4 in main. Since this PR should be merged fairly quickly after that one and V4 has not been included in a release yet, I opted for not bumping up the file version to V5 but to just update V4 instead.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
```diff
@@ -100,17 +110,18 @@ interface View extends Closeable {
          * Iterator over the neighbors of a given node.  Only the most recently instantiated iterator
          * is guaranteed to be valid.
          */
-        NodesIterator getNeighborsIterator(int node);
+        NodesIterator getNeighborsIterator(int level, int node);
 
         /**
          * @return the number of nodes in the graph
          */
+        // TODO can we just remove this in favor of graph's?
```
jkni: I think we can. Most `View` usages shouldn't need size, and where they do, they could also pass the relevant graph.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
```diff
@@ -137,16 +148,20 @@ interface ScoringView extends View {
 
     static String prettyPrint(GraphIndex graph) {
         StringBuilder sb = new StringBuilder();
-        sb.append(graph);
+        sb.append(graph); // toString looks like OnHeapGraphIndex(size=9, entryPoint=NodeAtLevel(level=4, node=6))
```
jkni: I think we should remove this comment, since it describes the behavior of a specific GraphIndex implementation

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
```diff
@@ -156,4 +171,44 @@ static String prettyPrint(GraphIndex graph) {
 
         return sb.toString();
     }
+
+    // Comparable b/c it gets used in CSLM
```
jkni:
```suggestion
    // Comparable b/c it gets used in ConcurrentSkipListMap
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -159,11 +166,72 @@ public GraphIndexBuilder(BuildScoreProvider scoreProvider,
...
+     * @param scoreProvider    describes how to determine the similarities between vectors
+     * @param maxDegrees       the maximum number of connections a node can have in each layer; if fewer entries
+     *                         are specified than the number of layers, the last entry is used for all remaining layers.
+     * @param beamWidth        the size of the beam search to use when finding nearest neighbors.
+     * @param neighborOverflow the ratio of extra neighbors to allow temporarily when inserting a
+     *                         node. larger values will build more efficiently, but use more memory.
+     * @param alpha            how aggressive pruning diverse neighbors should be.  Set alpha &gt; 1.0 to
+     *                         allow longer edges.  If alpha = 1.0 then the equivalent of the lowest level of
+     *                         an HNSW graph will be created, which is usually not what you want.
+     * @param addHierarchy     whether we want to add an HNSW hierarchy on top of the Vamana index.
```
jkni: All other constructors in this class taking `addHierarchy` also need it in their Javadoc

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -159,11 +166,72 @@ public GraphIndexBuilder(BuildScoreProvider scoreProvider,
...
+                             float neighborOverflow,
+                             float alpha,
+                             boolean addHierarchy,
                              ForkJoinPool simdExecutor,
                              ForkJoinPool parallelExecutor)
     {
-        if (M <= 0) {
-            throw new IllegalArgumentException("maxConn must be positive");
+        if (maxDegrees.stream().mapToInt(i -> i).anyMatch(i -> i <= 0)) {
+            throw new IllegalArgumentException("layer degree must be positive");
```
jkni:
```suggestion
            throw new IllegalArgumentException("layer degrees must be positive");
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -159,11 +166,72 @@ public GraphIndexBuilder(BuildScoreProvider scoreProvider,
...
+                             int beamWidth,
+                             float neighborOverflow,
+                             float alpha,
+                             boolean addHierarchy,
                              ForkJoinPool simdExecutor,
                              ForkJoinPool parallelExecutor)
     {
-        if (M <= 0) {
-            throw new IllegalArgumentException("maxConn must be positive");
+        if (maxDegrees.stream().mapToInt(i -> i).anyMatch(i -> i <= 0)) {
```
jkni:
```suggestion
        if (maxDegrees.stream().anyMatch(i -> i <= 0)) {
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -179,48 +247,67 @@ public GraphIndexBuilder(BuildScoreProvider scoreProvider,
...
+                other.alpha,
+                other.addHierarchy,
+                other.simdExecutor,
+                other.parallelExecutor);
 
         // Copy each node and its neighbors from the old graph to the new one
-        for (int i = 0; i < other.graph.getIdUpperBound(); i++) {
-            if (!other.graph.containsNode(i)) {
-                continue;
+        IntStream.range(0, other.graph.getIdUpperBound()).parallel().forEach(i -> {
```
jkni: We should probably do this on `other`'s parallel executor, so that we don't have an impact on the `commonPool`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -423,6 +412,20 @@ public long addGraphNode(int node, RandomAccessVectorValues ravv) {
         return addGraphNode(node, ravv.getVector(node));
     }
 
+    private int getRandomGraphLevel() {
```
jkni, line 428: Can you add a quick Javadoc of how we're weighting to the various levels? I wouldn't have predicted this purely from the method name.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -423,6 +412,20 @@ public long addGraphNode(int node, RandomAccessVectorValues ravv) {
...
+        double ml;
+        if (addHierarchy) {
+            ml = graph.getDegree(0) == 1 ? 1 : 1 / log(1.0 * graph.getDegree(0));
+        } else {
+            ml = 0;
+        }
+        double randDouble;
+        do {
+            randDouble = this.rng.nextDouble();  // avoid 0 value, as log(0) is undefined
```
jkni: Might as well put this RNG bit in the conditional branch above, as it'll get multiplied by 0 if we don't have a hierarchy.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -431,113 +434,79 @@ public long addGraphNode(int node, RandomAccessVectorValues ravv) {
      * other in-progress updates as neighbor candidates.
      *
      * @param node the node ID to add
-     * @return an estimate of the number of extra bytes used by the graph after adding the given node
```
jkni, line 434: Why did we remove the `return` from the Javadoc? It looks like that's still the meaning of the return value

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -431,113 +434,79 @@ public long addGraphNode(int node, RandomAccessVectorValues ravv) {
...
      */
     public long addGraphNode(int node, VectorFloat<?> vector) {
         // do this before adding to in-progress, so a concurrent writer checking
         // the in-progress set doesn't have to worry about uninitialized neighbor sets
-        graph.addNode(node);
+        var nodeLevel = new NodeAtLevel(getRandomGraphLevel(), node);
```
jkni: Let's move the comment so that it's directly above `graph.addNode` again, as its discussing addNode behavior

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -558,129 +527,114 @@ public synchronized long removeDeletedNodes() {
...
         // Remove the deleted nodes from the graph
         assert toDelete.cardinality() == nRemoved : "cardinality changed";
+        int nodeLayers = 0;
         for (int i = toDelete.nextSetBit(0); i != NO_MORE_DOCS; i = toDelete.nextSetBit(i + 1)) {
-            graph.removeNode(i);
+            nodeLayers += graph.removeNode(i);
         }
-
-        return nRemoved * graph.ramBytesUsedOneNode();
+        // TODO this is not correct since different layers can use more or less ram due to different degrees
```
jkni: I think this TODO is worth fixing before merging

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -797,6 +696,36 @@ public boolean get(int index) {
     }
 
     public void load(RandomAccessReader in) throws IOException {
+        if (graph.size(0) != 0) {
```
jkni, line 716: I'm concerned about breaking deserialization for previously serialized graphs -- can we add a magic number here and use the old v3 deserialization if we don't hit it? If not, we should make sure to emphasize this point in UPGRADING.md so that consumers can handle this by plugging in loadV3 themselves.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
```diff
@@ -256,4 +259,15 @@ public void foreach(NodeConsumer consumer) {
     public interface NodeConsumer {
         void accept(int node, float score);
     }
+
+    public void copyFrom(NodeQueue other) {
```
jkni, line 268: Can we add a quick javadoc here and use an if/else instead of the early return, given that the conditional branches are pretty small

pom.xml
```diff
@@ -41,7 +41,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.install.skip>true</maven.install.skip>
-        <revision>4.0.0-beta.2-SNAPSHOT</revision>
```
jkni: We can undo this change. I don't think we want `hnsw` in the revision

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -327,52 +348,70 @@ private SearchResult resume(int initialVisited, int topK, int rerankK, float thr
...
         }
     }
 
+    SearchResult resume(int numVisited, int topK, int rerankK, float threshold, float rerankFloor) {
+        // rR is persistent to save on allocations
+        rerankedResults.clear();
+        rerankedResults.setMaxSize(topK);
+
+        // add evicted results from the last call back to the candidates
+        var previouslyEvicted = evictedResults.size() > 0 ? new SparseBits() : Bits.NONE;
```
jkni: Do we use previouslyEvicted at all? I don't see anywhere we use it other than updating it in the following forEach

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -24,55 +24,87 @@
...
-    final int maxDegree;
-
-    OnHeapGraphIndex(int M, int maxOverflowDegree, BuildScoreProvider scoreProvider, float alpha) {
-        this.maxDegree = M;
-        this.nodes = new ConcurrentNeighborMap(scoreProvider, maxDegree, maxOverflowDegree, alpha);
+    private final AtomicInteger maxNodeId = new AtomicInteger(-1);
+
+    // Maximum number of neighbors (edges) per node per layer
+    final IntArrayList maxDegrees;
+    // The ratio by which we can overflow the neighborhood of a node during construction. Should be higher than 1
```
jkni: Can you make this comment just slightly richer on why it should be higher than 1?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -191,53 +267,150 @@ public ThreadSafeGrowableBitSet getDeletedNodes() {
...
-        return nodes.contains(nodeId);
+        return layers.get(0).contains(nodeId);
     }
 
-    public double getAverageDegree() {
-        return IntStream.range(0, getIdUpperBound())
-                .filter(this::containsNode)
-                .mapToDouble(i -> getNeighbors(i).size())
+    /**
+     * Returns the average degree computed over nodes in the base layer.
```
jkni: Javadoc seems wrong? The method takes a `level` argument

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/CommonHeader.java
```diff
@@ -85,8 +135,47 @@ int size() {
...
+        public final int degree;
+
+        public LayerInfo(int size, int degree) {
+            this.size = size;
+            this.degree = degree;
+        }
+
+        public static List<LayerInfo> fromGraph(GraphIndex graph, OrdinalMapper mapper) {
+            int l0Size = mapper.maxOrdinal() + 1;
+            return IntStream.range(0, graph.getMaxLevel() + 1)
```
jkni: `IntStream.rangeClosed` would be more idiomatic here

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -126,16 +185,29 @@ public int getIdUpperBound() {
...
+        long layerOffset = neighborsOffset;
+        layerOffset += level > 0? layer0NodeSize * size(0) : 0;
+        for (int lvl = 1; lvl < level; lvl++) {
+            layerOffset += layerUpperNodeSize * size(lvl);
+        }
+
         try (var reader = readerSupplier.get()) {
-            int[] valid_nodes = new int[size];
+            int[] valid_nodes = new int[size(level)];
+            int upperBound = level == 0? getIdUpperBound() : size(level);
```
jkni:
```suggestion
            int upperBound = level == 0 ? getIdUpperBound() : size(level);
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -191,19 +217,19 @@ public synchronized void write(Map<FeatureId, IntFunction<Feature.State>> featur
...
                     feature.writeInline(out, supplier.apply(originalOrdinal));
                 }
             }
 
-            var neighbors = view.getNeighborsIterator(originalOrdinal);
+            var neighbors = view.getNeighborsIterator(0, originalOrdinal);
             if (neighbors.size() > graph.maxDegree()) {
                 var msg = String.format("Node %d has more neighbors %d than the graph's max degree %d -- run Builder.cleanup()!",
-                                        originalOrdinal, neighbors.size(), graph.maxDegree());
+                        originalOrdinal, neighbors.size(), graph.maxDegree());
```
jkni: I think this should be l0's degree and not `graph.maxDegree` in both the conditional check and error message

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -220,10 +246,67 @@ public synchronized void write(Map<FeatureId, IntFunction<Feature.State>> featur
...
+                    out.writeInt(-1);
+                }
+                nodesWritten++;
+            }
+            if (nodesWritten != layerSize) {
+                throw new IllegalStateException("Mismatch between layer size and nodes written");
+            }
+        }
+
+
```
jkni:
```suggestion
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -220,10 +246,67 @@ public synchronized void write(Map<FeatureId, IntFunction<Feature.State>> featur
...
+        // Write separated features
+        for (var featureEntry : featureMap.entrySet()) {
+            if (isSeparated(featureEntry.getValue())) {
+                var fid = featureEntry.getKey();
+                var supplier = featureStateSuppliers.get(fid);
+                if (supplier == null) continue;
```
jkni: Is it ever the case where we'd have a null supplier for a separated feature and it's not a logic error?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -336,12 +419,17 @@ public OnDiskGraphIndexWriter build() throws IOException {
...
+            } else if (features.containsKey(FeatureId.SEPARATED_NVQ)) {
+                dimension = ((SeparatedNVQ) features.get(FeatureId.SEPARATED_NVQ)).dimension();
             } else {
-                throw new IllegalArgumentException("Inline or NVQ vectors must be provided.");
+                throw new IllegalArgumentException("Inline or separated vector feature must be provided");
             }
 
             if (ordinalMapper == null) {
                 ordinalMapper = new OrdinalMapper.MapMapper(sequentialRenumbering(graphIndex));
+                ordinalMapper = new OrdinalMapper.MapMapper(sequentialRenumbering(graphIndex));
```
jkni: Not sure why this is assigned twice

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FusedADC.java
```diff
@@ -68,19 +70,20 @@ public int headerSize() {
     }
 
     @Override
-    public int inlineSize() {
+    public int featureSize() {
         return pq.compressedVectorSize() * maxDegree;
     }
 
     static FusedADC load(CommonHeader header, RandomAccessReader reader) {
+        // TODO doesn't work with different degrees
```
jkni, line 78: totally fine leaving this as a TODO but we should probably elevate it to somewhere more visible in the documentation

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FusedADC.java
```diff
@@ -100,7 +103,7 @@ public void writeInline(DataOutput out, Feature.State state_) throws IOException
         var state = (FusedADC.State) state_;
         var pqv = state.pqVectors;
 
-        var neighbors = state.view.getNeighborsIterator(state.nodeId);
+        var neighbors = state.view.getNeighborsIterator(0, state.nodeId); // TODO
```
jkni: Not sure what the TODO here is for

jvector-base/src/main/java/io/github/jbellis/jvector/util/AbstractLongHeap.java
```diff
@@ -28,7 +28,7 @@
 
 /**
  * A min heap that stores longs; a primitive priority queue that like all priority queues maintains
- * a partial ordering of its elements such that the least element can always be found in constant
+ * a partial ordering of its elements such that the leastbo element can always be found in constant
```
jkni: typo?

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestDeletions.java
```diff
@@ -114,20 +115,26 @@ public void testCleanup() throws IOException {
         try (var out = TestUtil.openDataOutputStream(outputPath)) {
             graph.save(out);
         }
-        var b2 = new GraphIndexBuilder(ravv, VectorSimilarityFunction.COSINE, 2, 10, 1.0f, 1.0f);
-        try (var readerSupplier = new SimpleMappedReader.Supplier(outputPath)) {
-            b2.load(readerSupplier.get());
-        }
-        var reloadedGraph = b2.getGraph();
-        assertGraphEquals(graph, reloadedGraph);
+        // TODO when we fix load()
```
jkni: Is this still a TODO?

jvector-tests/src/test/java/io/github/jbellis/jvector/quantization/TestADCGraphIndex.java
```diff
@@ -63,12 +62,11 @@ public void testFusedGraph() throws Exception {
...
 
         try (var readerSupplier = new SimpleMappedReader.Supplier(outputPath);
-             var onDiskGraph = OnDiskGraphIndex.load(readerSupplier, 0);
-             var cachedOnDiskGraph = new CachingGraphIndex(onDiskGraph, 5))
+             var onDiskGraph = OnDiskGraphIndex.load(readerSupplier, 0))
         {
             TestUtil.assertGraphEquals(graph, onDiskGraph);
-            TestUtil.assertGraphEquals(graph, cachedOnDiskGraph);
-            try (var cachedOnDiskView = cachedOnDiskGraph.getView())
+            TestUtil.assertGraphEquals(graph, onDiskGraph);
```
jkni:
```suggestion
```

This duplicates the above assertion now.

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/GraphIndexBuilderTest.java
```diff
@@ -95,38 +100,39 @@ public void testRescore() {
...
-        try(var readerSupplier = new SimpleMappedReader.Supplier(indexDataPath)) {
-            builder.load(readerSupplier.get());
-        }
-
-        assertEquals(ravv.size(), builder.graph.size());
-        for (int i = 0; i < ravv.size(); i++) {
-            assertTrue(builder.graph.containsNode(i));
-        }
-    }
+    // TODO
```
jkni: Should fix this TODO

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -48,47 +62,92 @@
...
+    final NodeAtLevel entryNode;
     final int idUpperBound;
     final int inlineBlockSize; // total size of all inline elements contributed by features
     final EnumMap<FeatureId, ? extends Feature> features;
     final EnumMap<FeatureId, Integer> inlineOffsets;
+    private final List<CommonHeader.LayerInfo> layerInfo;
+    // offset of L0 adjacency data
     private final long neighborsOffset;
+    /** For layers > 0, store adjacency fully in memory. */
+    private volatile AtomicReference<List<Int2ObjectHashMap<int[]>>> inMemoryNeighbors;
```
jkni:
```suggestion
    private final AtomicReference<List<Int2ObjectHashMap<int[]>>> inMemoryNeighbors;
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -126,16 +185,29 @@ public int getIdUpperBound() {
...
+
+        long layerOffset = neighborsOffset;
+        layerOffset += level > 0? layer0NodeSize * size(0) : 0;
+        for (int lvl = 1; lvl < level; lvl++) {
+            layerOffset += layerUpperNodeSize * size(lvl);
+        }
+
         try (var reader = readerSupplier.get()) {
-            int[] valid_nodes = new int[size];
+            int[] valid_nodes = new int[size(level)];
```
jkni:
```suggestion
            int[] validNodes = new int[size(level)];
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
```diff
@@ -45,15 +46,18 @@
  */
 public interface GraphIndex extends AutoCloseable, Accountable {
     /** Returns the number of nodes in the graph */
-    int size();
+    @Deprecated
```
jkni, line 49: We should expand this javadoc with the reason for deprecation.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
```diff
@@ -45,15 +46,18 @@
...
+    }
 
     /**
      * Get all node ordinals included in the graph. The nodes are NOT guaranteed to be
      * presented in any particular order.
      *
      * @return an iterator over nodes where {@code nextInt} returns the next node.
      */
-    NodesIterator getNodes();
+    NodesIterator getNodes(int level);
```
jkni, line 60: We should update the javadoc with the new param/behavior.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -92,6 +94,10 @@ public GraphIndex.View getView() {
         return view;
     }
 
+    public void usePruning(boolean usage) {
```
jkni, line 102: Can we add a javadoc for which behavioral changes usePruning enables?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -105,6 +111,10 @@ public static SearchResult search(VectorFloat<?> queryVector, int topK, RandomAc
         }
     }
 
+    public void setView(GraphIndex.View view) {
```
jkni, line 128: Let's add a Javadoc here with an example of when this would be/is used.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/CommonHeader.java
```diff
@@ -85,8 +135,47 @@ int size() {
...
+        public final int size;
+        public final int degree;
+
+        public LayerInfo(int size, int degree) {
+            this.size = size;
+            this.degree = degree;
+        }
+
+        public static List<LayerInfo> fromGraph(GraphIndex graph, OrdinalMapper mapper) {
+            int l0Size = mapper.maxOrdinal() + 1;
```
jkni: Can we remove this specialized definition of `size` for l0's LayerInfo? It looks like this special-case is perhaps only used in `loadInMemoryLayers` where idUpperBound is also in scope, so we could switch those special L0 callers (used to seek ahead and in an assert) to idUpperBound and keep L0's size to actual node count.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
marianotepper: I'll deprecate the method. Or is it better to just remove it?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
jkni: Maybe something like the following:
```
    /**
     * Sets the view of the graph to be used by the searcher.
     * <p>
     * This method should be used when the searcher operates over a view whose contents might not reflect all changes
     * to the underlying graph, such as {@link OnHeapGraphIndex.ConcurrentGraphIndexView}. This is an optimization over
     * creating a new graph searcher with every update to the view.
     *
     * @param view the new view
     */
    public void setView(GraphIndex.View view) {
        this.view = view;
    }
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FusedADC.java
marianotepper: How about putting a note in UPGRADING.md?

jkni: That works for me.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
marianotepper: Can you please expand your comment?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
marianotepper: We are using it [here](https://github.com/datastax/jvector/blob/7b495e63a23bf79109b53db6df2d89f101e1f556/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java#L402). Do you think it is not needed?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
marianotepper: I think I understand now. Please let me know if the change captured your intent

README.md
```diff
@@ -10,17 +10,19 @@ There are two broad categories of ANN index:
 
 Graph-based indexes tend to be simpler to implement and faster, but more importantly they can be constructed and updated incrementally.  This makes them a much better fit for a general-purpose index than partitioning approaches that only work on static datasets that are completely specified up front.  That is why all the major commercial vector indexes use graph approaches.
 
-JVector is a graph index in the DiskANN family tree.
+JVector is a graph index that takes a hybrid merging the the DiskANN and HNSW family trees.
```
tlwillke: "... that merges the best of the DiskANN and HNSW algorithms."

README.md
```diff
@@ -10,17 +10,19 @@ There are two broad categories of ANN index:
...
+JVector borrows the hierarchical structure from HNSW, and uses Vamana (the algorithm behind DiskANN) within each layer.
 
 
 ## JVector Architecture
 
-JVector is a graph-based index that builds on the DiskANN design with composeable extensions.
+JVector is a graph-based index that builds on the HNSW anD DiskANN designs with composable extensions.
```
tlwillke: Typo

UPGRADING.md
```diff
@@ -5,11 +5,22 @@
   in each vector with high accuracy by first applying a nonlinear transformation that is individually fit to each
   vector. These nonlinearities are designed to be lightweight and have a negligible impact on distance computation
   performance.
+- Support for hierarchical graph indices. These new type of indices blends HNSW and DiskANN in a novel way. An
```
tlwillke: blends --> blend

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
```diff
@@ -95,10 +95,23 @@ default boolean containsNode(int nodeId) {
     @Override
     void close() throws IOException;
 
+    /**
+     * @return The maximum (coarser) level with that contains a vector in the graph.
```
jkni:
```suggestion
     * @return The maximum (coarser) level that contains a vector in the graph.
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -94,6 +94,11 @@ public GraphIndex.View getView() {
         return view;
     }
 
+    /**
+     * Whe using pruning, we are using a heuristic to terminate the search earlier.
```
jkni:
```suggestion
     * When using pruning, we are using a heuristic to terminate the search earlier.
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
```diff
@@ -260,14 +260,18 @@ public interface NodeConsumer {
         void accept(int node, float score);
     }
 
+    /**
+     * Copies the other NodeQueue to this one. If its order (MIN_HEAP or MAX_HEAP) is the same as this,
+     * it is copied verbatim. If it differs, every lement is re-inserted into this.
```
jkni:
```suggestion
     * it is copied verbatim. If it differs, every element is re-inserted into this.
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndex.java
jkni: Deprecate works for me.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
jkni: Yes, confirming the change captured my intent.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -162,21 +188,21 @@ public synchronized void write(Map<FeatureId, IntFunction<Feature.State>> featur
...
             // if no node exists with the given ordinal, write a placeholder
             if (originalOrdinal == OrdinalMapper.OMITTED) {
                 out.writeInt(-1);
-                for (var feature : featureMap.values()) {
-                    out.seek(out.position() + feature.inlineSize());
+                for (var feature : inlineFeatures) {
+                    out.seek(out.position() + feature.featureSize());
                 }
                 out.writeInt(0);
                 for (int n = 0; n < graph.maxDegree(); n++) {
```
jkni: I missed this on the first pass, but the loop bound here should be `graph.getDegree(0)`, not `graph.maxDegree()`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
jkni: We use evictedResults, but not the previouslyEvicted bitset created/populated here (https://github.com/datastax/jvector/blob/main/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java#L302). We used to use this bitset in `resume` (https://github.com/datastax/jvector/blob/main/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java#L302), but we don't in the new hierarchical structure. I think we can keep evictedResults but don't need to spend the CPU to create/update previouslyEvicted (it's set but never read).

jvector-tests/src/test/java/io/github/jbellis/jvector/TestUtil.java
```diff
@@ -162,17 +163,17 @@ public static void writeFusedGraph(GraphIndex graph, RandomAccessVectorValues ra
...
         int n = it.nextInt();
         neighbors.add(n);
       }
       return neighbors;
     }
 
-    static List<Integer> sortedNodes(GraphIndex h) {
-          var graphNodes = h.getNodes();
+    static List<Integer> sortedNodes(GraphIndex h, int level) {
+          var graphNodes = h.getNodes(level); // TODO
```
tlwillke, line 176: This is marked TODO without explanation.  Is there something remaining?

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/GraphIndexBuilderTest.java
```diff
@@ -66,13 +72,13 @@ public void testRescore() {
...
         var bsp = BuildScoreProvider.randomAccessScoreProvider(ravv, VectorSimilarityFunction.EUCLIDEAN);
-        var builder = new GraphIndexBuilder(bsp, 2, 2, 10, 1.0f, 1.0f);
+        var builder = new GraphIndexBuilder(bsp, 2, 2, 10, 1.0f, 1.0f, addHierarchy);
 
         // Add 3 nodes
         builder.addGraphNode(0, ravv.getVector(0));
         builder.addGraphNode(1, ravv.getVector(1));
         builder.addGraphNode(2, ravv.getVector(2));
-        var neighbors = builder.graph.getNeighbors(0);
+        var neighbors = builder.graph.getNeighbors(0, 0); // TODO
```
tlwillke, line 81: TODO?  Something missing?

jvector-tests/src/test/java/io/github/jbellis/jvector/quantization/TestADCGraphIndex.java
```diff
@@ -78,14 +75,14 @@ public void testFusedGraph() throws Exception {
                         var fusedScoreFunction = cachedOnDiskView.approximateScoreFunctionFor(queryVector, similarityFunction);
                         var ordinal = getRandom().nextInt(graph.size());
                         // first pass compares fused ADC's direct similarity to reranker's similarity, used for comparisons to a specific node
-                        var neighbors = cachedOnDiskView.getNeighborsIterator(ordinal);
+                        var neighbors = cachedOnDiskView.getNeighborsIterator(0, ordinal); // TODO
```
tlwillke, line 78: TODO?  Something left to do?

tlwillke (approved):
Some TODOs remaining but seem minor.  A major contribution.  Will affect public interfaces but it is the path forward, so worth it!

PR #403, opened 2025-03-20
Remove query-time usage of ByteSequence::slice in PQVectors to reduce object allocations

michaeljmarshall:
With https://github.com/datastax/jvector/pull/370, I introduced the `ByteSequence::slice` method and started using it on the query path. The PR replaces long lived pq vector ByteSequences with short lived slices of larger ByteSequences created at query time. However, profiling reveals that these objects get created very often for certain workloads, especially brute force workloads in Cassandra. After analyzing the code, it looks trivial to migrate to a solution where we do not create a slice object per access to a pq vector. The main downside is the addition of an `offset` and a `length` argument to many methods and the fact that these changes feel brittle, if we want to maintain the `slice()` construct, which I think we do for the simplicity of non-query path code. However, given the performance implications and the fact that the interface changes are all internal to jvector, I think this change is worth considering.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQDecoder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQDecoder.java
@@ -53,8 +53,8 @@ abstract class PQDecoder implements ScoreFunction.ApproximateScoreFunction {
             }
         }
 
-        protected float decodedSimilarity(ByteSequence<?> encoded) {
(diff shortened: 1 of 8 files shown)


michaeljmarshall:
I keep seeing failures for: 

> Test2DThreshold.testThreshold20k:44->testThreshold:74 visited 75.38348% of the vectors, which is more than 75.0%

This doesn't seem related and feels potentially flaky since the tests pass on retry.

marianotepper:
I agree the comments above and the PR looks good to me. I suggest we open an issue about the interface so that we do not forget.

marianotepper (approved):

PR #404, opened 2025-03-21
Fix flaky tests and eliminate console output

marianotepper:
Two minor updates:
- In Test2DThreshold.testThreshold20k, use a more lenient threshold. The previous value was randomly failing (sometimes).
- In TestCompressedVectors.testNVQEncodings, remove the print statements that were polluting the console.


tlwillke (approved):
+1

PR #405, opened 2025-03-28
Search pruning & fix the reported number of visited nodes

marianotepper:
There were two bugs in the measurements in Bench:
1. The start time for measuring runtime was wrongly placed, so we were getting the wrong measurement for the second value in usePruningGrid. This bug caused to identify the use of search pruning (early termination) as a pessimization with the new hierarchical index, but the contrary is true. **We should enable search pruning by default**
2. The time was also the aggregate of both runs.
3. The reported node count was the aggregate of both runs through the queries. We are now correcting that.

This PR also changes returning the aggregate count of all queries to the average, which is more informative.


marianotepper:
You are right. The test failure was legit. The problem goes away if we keep pruning=true and add a bit of reranking (2x). I think this is a reasonable assumption. And I fixed the test that way.

PR #407, opened 2025-04-01
Mwolters/dev

MarkWolters:
Reworking of JMH tests to include additional tests with tabular outputs for both random and static set vector tests. Includes configurable parameterization of output representation including text table, json and sqlite backed.


tlwillke:
Let's close the PR for now but keep the branch for further development.

PR #408, opened 2025-04-02
Rework the computation of accuracy

marianotepper:
- Added the AccuracyMetrics class to have a single point where these computations happen. Easier to debug and reuse.
- Added mean average precision as an additional order-aware accuracy metric.
- Switched the ground truth from a List<Set<Integer>> to List<List<Integer>> to enable the computation of order-aware metrics like "n-recall@k" (see section 3.2 in [this paper](https://arxiv.org/pdf/2401.08281)) and mean average precision.


tlwillke (commented):
We discussed a couple of exceptions that could be added.  Otherwise good to go.

tlwillke (approved):
The new checks look good.

Issue #409, opened 2025-04-02
Optimize NodeQueue/AbstractLongHeap with option to bulk add elements


michaeljmarshall:
The current `NodeQueue` implementation forces users to add elements one at a time, which requires O(n*log(n)) time. A bulk addition of elements followed by a single heapify operation is instead O(n) time. In some brute force scenarios, I found that we could add 400k elements to a queue at a time, which makes for a significant difference in the time to build the queue. This is particularly relevant for a brute force scenario https://github.com/datastax/cassandra/pull/1643.

I propose that we add an option for the `NodeQueue` (and the `AbstractLongHeap`) to consume an iterator that produces the node id and score, adds those scores to the heap without running upheap, and then applies the bulk downheap operation to re-heapify the heap. The iterator solution would help keep the space complexity down.

michaeljmarshall:
It's not clear to me if there are applications for this logic within jvector, which might determine the utility of this feature to the project. Note that there are several places where we call `push()` iteratively, which suggests they might benefit from this change. However, for small cardinalities, the performance difference is likely negligible.

michaeljmarshall:
Back of the envelope math (a.k.a chatgpt) suggests that for the 400k example I provided, we're talking about 800k comparisons in the bulk add scenario and 8M in the iterative add scenario.

PR #415, opened 2025-04-03
Implement NodeQueue#pushAll and AbstractLongHeap#addAll

michaeljmarshall:
Fixes https://github.com/datastax/jvector/issues/409

Adds bulk add operations to the AbstractLongHeap
and to the NodeQueue to reduce comparisons
required when adding many elements at once.

The current NodeQueue implementation forces users to add elements one at a time, which requires O(n*log(n)) time. A bulk addition of elements followed by a single heapify operation is instead O(n) time. In some brute force scenarios, I found that we could add 400k elements to a queue at a time, which makes for a significant difference in the time to build the queue. This is particularly relevant for a brute force scenario https://github.com/datastax/cassandra/pull/1643.

I propose that we add an option for the NodeQueue (and the AbstractLongHeap) to consume an iterator that produces the node id and score, adds those scores to the heap without running upheap, and then applies the bulk downheap operation to re-heapify the heap. The iterator solution would help keep the space complexity down.

It's not clear to me if there are applications for this logic within jvector, which might determine the utility of this feature to the project. Note that there are several places where we call push() iteratively, which suggests they might benefit from this change. However, for small cardinalities, the performance difference is likely negligible.

Further, if this library takes on brute force calculations, this change will become meaningful, so I propose we add it.

Back of the envelope math (a.k.a chatgpt) suggests that for the 400k example I provided, we're talking about 800k comparisons in the bulk add scenario and 8M in the iterative add scenario.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeQueue.java
@@ -28,9 +28,8 @@ import io.github.jbellis.jvector.graph.similarity.ScoreFunction;
 import io.github.jbellis.jvector.util.AbstractLongHeap;
 import io.github.jbellis.jvector.util.BoundedLongHeap;
 import io.github.jbellis.jvector.util.NumericUtils;
+import java.util.PrimitiveIterator;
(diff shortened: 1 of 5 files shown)


marianotepper (approved):
LGTM

Issue #418, opened 2025-04-07
Significant allocation of vector objects in ConcurrentNeighborMap::retainDiverseInternal


michaeljmarshall:
When we call `retainDiverseInternal`, we create a vector in nested for loops through the `diversityProviderFor` method. This creates enough objects to show up on a memory profile for graph construction. Is there a way to decrease allocations by either using and reusing a thread local scratch vector or a vector per call to `retainDiverseInternal` so that the vector gets overwritten on each iteration of the nested for loops?

https://github.com/datastax/jvector/blob/00a13a8a793009ffd13bb61374725653e0c50b87/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ConcurrentNeighborMap.java#L306-L333

ultimately calls:

https://github.com/datastax/jvector/blob/e8d5c3c6ca915fe58dbda175c460e40d388ca155/jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java#L158

Screen shot of heap allocation profile:

![Image](https://github.com/user-attachments/assets/546c3a60-6241-439e-973e-d682ec291ff2)

marianotepper:
Addressed by #419

PR #419, opened 2025-04-07
Reduce the number of vector allocations in BuildScoreProvider.pqBuilderScoreProvider

marianotepper:
Addresses #418. It should help reduce the number of vector allocations during graph construction when using PQ compression

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
@@ -138,6 +138,7 @@ public interface BuildScoreProvider {
      * with reranking performed using RandomAccessVectorValues (which is intended to be
      * InlineVectorValues for building incrementally, but should technically
      * work with any RAVV implementation).
+     * This class is not thread safe, we should never its results to another thread.
      */
     static BuildScoreProvider pqBuildScoreProvider(VectorSimilarityFunction vsf, PQVectors pqv) {
         int dimension = pqv.getOriginalSize() / Float.BYTES;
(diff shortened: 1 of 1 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
```diff
@@ -153,15 +154,15 @@ public SearchScoreProvider diversityProviderFor(int node1) {
                 // like searchProviderFor, this skips reranking; unlike sPF, it uses pqv.scoreFunctionFor
                 // instead of precomputedScoreFunctionFor; since we only perform a few dozen comparisons
                 // during diversity computation, this is cheaper than precomputing a lookup table
-                VectorFloat<?> v1 = vts.createFloatVector(dimension);
+                VectorFloat<?> v1 = reusableVector.get();
```
michaeljmarshall, line 158: We should probably document here or in the parent interface that the resulting `SearchScoreProvider` is not thread safe. I reviewed the current usages of this result and of the result from `searchProviderFor`, and we never publish the result of the `SearchScoreProvider` to another thread, so it is correct as is.

marianotepper: I agree that it should be documented. I will add a comment. Thanks!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
```diff
@@ -138,6 +138,7 @@ public SearchScoreProvider diversityProviderFor(int node1) {
      * with reranking performed using RandomAccessVectorValues (which is intended to be
      * InlineVectorValues for building incrementally, but should technically
      * work with any RAVV implementation).
+     * This class is not thread safe, we should never its results to another thread.
```
michaeljmarshall: Nit: we should never publish its results to another thread.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
michaeljmarshall: Thanks!

michaeljmarshall (approved):
LGTM

tlwillke (commented):
+1

PR #420, opened 2025-04-07
Fix native implementations of PQ assembleAndSum and pqDecodedCosineSimilarity

jkni:
This addresses two regressions from #403.

The first is the two assertions in the `assembleAndSum` path and corresponding implementation, asserting we'll have 0 offset into the PQ buffer and that the length will be equal to the PQ buffer's length. These asserts consistently fire. I believe we conflated these offsets with the `offset` of a sliced MemorySegment-backed ByteSequence, which is always 0.

The second is that the implementation of `pqDecodedCosineSimilarity` taking offset/length args is falling back to the scalar default implementation, so we provide an implementation.

We don't actually need to change the native code; it already takes a pointer and a length. As long as we slice into the PQ chunk by the offset and pass the right length across the native boundary, we're fine.

### DIFF
--- a/jvector-native/src/main/java/io/github/jbellis/jvector/vector/NativeVectorUtilSupport.java
+++ b/jvector-native/src/main/java/io/github/jbellis/jvector/vector/NativeVectorUtilSupport.java
@@ -125,7 +125,7 @@ final class NativeVectorUtilSupport implements VectorUtilSupport
     {
         assert baseOffsets.offset() == 0 : "Base offsets are expected to have an offset of 0. Found: " + baseOffsets.offset();
         // baseOffsets is a pointer into a PQ chunk - we need to index into it by baseOffsetsOffset and provide baseOffsetsLength to the native code
-        return NativeSimdOps.assemble_and_sum_f32_512(((MemorySegmentVectorFloat) data).get(), dataBase, ((MemorySegmentByteSequence) baseOffsets).get().asSlice(baseOffsetsOffset, baseOffsetsLength), baseOffsetsLength);
+        return NativeSimdOps.assemble_and_sum_f32_512(((MemorySegmentVectorFloat) data).get(), dataBase, ((MemorySegmentByteSequence) baseOffsets).get(), baseOffsetsOffset, baseOffsetsLength);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


jvector-native/src/main/java/io/github/jbellis/jvector/vector/NativeVectorUtilSupport.java
```diff
@@ -117,16 +117,15 @@ public void minInPlace(VectorFloat<?> v1, VectorFloat<?> v2) {
...
 
     @Override
     public float assembleAndSum(VectorFloat<?> data, int dataBase, ByteSequence<?> baseOffsets, int baseOffsetsOffset, int baseOffsetsLength)
     {
-        assert baseOffsetsOffset == 0;
-        assert baseOffsetsLength == baseOffsets.length();
-        return assembleAndSum(data, dataBase, baseOffsets);
+        assert baseOffsets.offset() == 0 : "Base offsets are expected to have an offset of 0. Found: " + baseOffsets.offset();
+        // baseOffsets is a pointer into a PQ chunk - we need to index into it by baseOffsetsOffset and provide baseOffsetsLength to the native code
+        return NativeSimdOps.assemble_and_sum_f32_512(((MemorySegmentVectorFloat) data).get(), dataBase, ((MemorySegmentByteSequence) baseOffsets).get().asSlice(baseOffsetsOffset, baseOffsetsLength), baseOffsetsLength);
```
michaeljmarshall: How hard would it be to update the native code to use the `baseOffsetsOffset`? The initial motivation for https://github.com/datastax/jvector/pull/403 was to reduce object allocation, and this essentially reintroduces those temporary objects in the native path.

michaeljmarshall (commented):
Thanks for the fix @jkni, I see my mistake now.

jvector-native/src/main/java/io/github/jbellis/jvector/vector/NativeVectorUtilSupport.java
jkni: small effort, I'll push shortly

marianotepper (approved):
LGTM

michaeljmarshall (approved):

PR #422, opened 2025-04-08
Update test resume

marianotepper:
Test resume was assuming a strict equivalence between a full search and a resumed search. Of course, they are not exactly equivalent, so we are relaxing the test to avoid flakiness in its output.


marianotepper:
The flakiness, although rare, is still present. Re-opening the issue.

tlwillke (approved):
+1

PR #425, opened 2025-04-09
Fix NPE in GraphIndexBuilder.load

marianotepper:
When using a scoreProvider that does not have a reranker, GraphIndexBuilder.load was throwing a null pointer exception. This checks whether a reranker is available and use it. If not, it will use the primary score function.


marianotepper:
Thanks for the review!

PR #426, opened 2025-04-09
Fix calls to deprecated GraphIndex.size()

marianotepper:
GraphIndex.size() is deprecated. It was returning size(0), but we should just call the method we want.


tlwillke (approved):
+1

PR #433, opened 2025-04-11
Fix/refactor NodeScoreIterator, BoundedLongHeap, and GrowableLongHeap bulk addition implementations

michaeljmarshall:
- **Fix boundary check on BoundedLongHeap::pushAll**
Test added that fails without fix
- **Breaking change: refactor NodeScoreIterator**
The methods in the NodeScoreIterator
had opposite semantics from the NodeQueue.
One key problem was that nextScore
advanced the iterator. I propose instead
that we align with the NodeQueue names
and add topScore along with pop as the
key methods and pop will advance the
iterator. This is a breaking change
both in terms of the API and the semantics,
but because this interface was only
just released last week and we are
also breaking the compile-compatibility,
I view this as pretty low risk.
- **Fix missed requirement that heap is 1-based, not 0-based**
- **Rename pushAll to pushN to match actual semantics**

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/util/AbstractLongHeap.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/util/AbstractLongHeap.java
@@ -66,12 +66,12 @@ public abstract class AbstractLongHeap {
     public abstract boolean push(long element);
 
     /**
-     * Adds all elements from the given iterator to this heap, in bulk.
+     * Adds `elementsSize` elements from the given iterator to this heap, in bulk.
      *
      * @param elements the elements to add
      * @param elementsSize the maximum number of elements to pull from the elements iterator
      */
-    public abstract void pushAll(PrimitiveIterator.OfLong elements, int elementsSize);
+    public abstract void pushN(PrimitiveIterator.OfLong elements, int elementsSize);
 
     protected long add(long element) {
         size++;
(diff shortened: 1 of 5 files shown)


marianotepper (dismissed):
Looks good to me. I agree with the new interface as it is more aligned with the rest of the library.

jvector-base/src/main/java/io/github/jbellis/jvector/util/AbstractLongHeap.java
```diff
@@ -84,32 +84,35 @@ protected long add(long element) {
     }
 
     /**
-     * Bulk-adds all elements from the given iterator to this heap, then re-heapifies
+     * Bulk-adds `elementsSize` elements from the given iterator to this heap, then re-heapifies
      * in O(n) time (Floyd's build-heap). For a proof explaining the linear time
      * complexity, see <a href="https://stackoverflow.com/a/18742428">this stackoverflow answer</a>.
```
marianotepper, line 90: I think we should clarify that it bulk-adds the minimum between elementsSize and the number of elements in the iterator

marianotepper (changes requested):
Looks good in general. Just one comment about clarifying the intent of the function pushN in the documentation.

Only as a suggestion, what about using "pushMany" instead of "pushN"?

jvector-base/src/main/java/io/github/jbellis/jvector/util/AbstractLongHeap.java
michaeljmarshall: I like `pushMany`, and I'll modify this javadoc too.

marianotepper (approved):

PR #434, opened 2025-04-11
Fix minor bug in OnDiskGraphIndex.getNodes

marianotepper:
The method getNodes was using the maxDegree from the wrong layer to compute layer0NodeSize


tlwillke (approved):
+1

PR #438, opened 2025-04-15
Fixes for GHA tag-release workflow

tlwillke:
This PR updates the GHA tag-release workflow to:

1. Fix the top-level POM version update to match tag-release
2. Reorder the tag and changelog steps so the changelog captures the latest changes
3. Create a second PR for the SNAPSHOT version bump


marianotepper (approved):
LGTM

PR #458, opened 2025-04-18
Simplify OnDiskGraphIndex.View to avoid code duplication

marianotepper:
This PR adds two small improvements:
1.  The methods OnDiskGraphIndex.View.getVector and OnDiskGraphIndex.View.getVectorInto shared a significant amount of code. This PR makes getVector call getVectorInto, simplifying the code to avoid potential errors.
2. The computation of the offsets for reading the file were (almost) duplicated in OnDiskGraphIndex.View.offsetFor and in OnDiskGraphIndex.View.neighborsOffsetFor. This PR refactors the computation so that the shared code goes into a common method.


tlwillke (approved):
LGTM

PR #459, opened 2025-04-18
Improve perf metrics measurement and summarization

tlwillke:
We are lacking comprehensive benchmarking with compact (tabular) summaries.  In addition to the current recall, execution time, and nodes visited measurements, this PR adds latency and recall.  Also:
- Each benchmark is now its own class
- There is a composite summary class that puts them together
- A table printer presents the summary
- The global configuration for each table is also provided

In addition, all benchmarks share a QueryExecutor for standardization.  The queries are consumed to ensure they are executed by the JVM.  And in some cases, a warm-up phase is provided.


marianotepper:
The PR looks look. I think it is a quality of life improvement and will facilitate comparative studies. I have a few suggestions about the design of a few classes.

The class QueryTester could receive a list if QueryBenchmark objects to be executed in its constructor, and run those in a non-static version of performQueries (as a minor point, I would rename this method to "run," "execute," or similar). The user can select which benchmarks to include in a run by creating a list QueryBenchmark objects and only those will be executed. I think that this change alone makes the design much more configurable for a user.

This takes me to the CompositeSummary class. Do we need it? It looks like it is essentially working as a collection of summaries. What would be downside of replacing the result of performQueries with List<BenchmarkSummary>? Since QueryTester could execute a configurable number of benchmarks, this change would provide the freedom to return a variable number of BenchmarkSummary objects as the result of performQueries.

Now, regarding BenchmarkSummary, it looks like a common functionality to all benchmarks is to return statistics about something. I suggest that BenchmarkSummary takes on a structure similar to this one:

```
public interface BenchmarkSummary {
    class QueryStat {
        private final String name;
        private final double value;
        
        public QueryStat(String name, double value) {
            this.name = name;
            this.value = value;
        }
        
        public String getName() {
            return name;
        }
        
        public double getValue() {
            return value;
        }
    }
    
    int size();
    
    QueryStat getStat(int i);
}
```

For example, ExecutionTimeBenchmark will return a summary with a size of 1, while CountBenchmark will return a summary with a size of 3.

This structure is something that we can bring to a table readily. Now, BenchmarkTablePrinter.printRow can have the signature:
```
public void printRow(double overquery, List<BenchmarkSummary> summaries)
```
The size of summaries and the number of QueryStat's in each summary will determine the number of columns in the table. The QueryStat.name provides the column header and QueryStat.value the value for that column.

I think that these relatively minor changes may make the contribution more adaptable, configurable, and extensible.

marianotepper:
For future reference, my previous comment:
> The class QueryTester could receive a list if QueryBenchmark objects
can enable in the future to configure this list from the command line, by using the QueryBenchmark name as a proxy.

tlwillke:
> The PR looks look. I think it is a quality of life improvement and will facilitate comparative studies. I have a few suggestions about the design of a few classes.

Thanks!

> The class QueryTester could receive a list if QueryBenchmark objects to be executed in its constructor, and run those in a non-static version of performQueries (as a minor point, I would rename this method to "run," "execute," or similar). The user can select which benchmarks to include in a run by creating a list QueryBenchmark objects and only those will be executed. I think that this change alone makes the design much more configurable for a user.

Great suggestion.  QueryTester now accepts a list of QueryBenchmarks.  Renamed performQueries as run.

> This takes me to the CompositeSummary class. Do we need it? It looks like it is essentially working as a collection of summaries. What would be downside of replacing the result of performQueries with `List<BenchmarkSummary>`? Since QueryTester could execute a configurable number of benchmarks, this change would provide the freedom to return a variable number of BenchmarkSummary objects as the result of performQueries.

Agreed and removed.  QueryTester now returns a LinkedHashMap, which is used to return BenchmarkSummary instances and their classes.

> Now, regarding BenchmarkSummary, it looks like a common functionality to all benchmarks is to return statistics about something. I suggest that BenchmarkSummary takes on a structure similar to this one:
> 
> ```
> public interface BenchmarkSummary {
>     class QueryStat {
>         private final String name;
>         private final double value;
>         
>         public QueryStat(String name, double value) {
>             this.name = name;
>             this.value = value;
>         }
>         
>         public String getName() {
>             return name;
>         }
>         
>         public double getValue() {
>             return value;
>         }
>     }
>     
>     int size();
>     
>     QueryStat getStat(int i);
> }
> ```
> 
> For example, ExecutionTimeBenchmark will return a summary with a size of 1, while CountBenchmark will return a summary with a size of 3.
> 
> This structure is something that we can bring to a table readily. Now, BenchmarkTablePrinter.printRow can have the signature:
> 
> ```
> public void printRow(double overquery, List<BenchmarkSummary> summaries)
> ```
> 
> The size of summaries and the number of QueryStat's in each summary will determine the number of columns in the table. The QueryStat.name provides the column header and QueryStat.value the value for that column.

This is the only suggestion I decided to skip for now.  I envision users wanting to frequently select a subset of stats from each benchmark.  So, using the available number (e.g. size of 3 for CountBenchmark) is not as useful.  Instead, the user uses the Metric list to explicitly select the stats of interest for the table.  It's very flexible to add and drop columns now.  And the String header, data format, and stat are all specified in one place.

> I think that these relatively minor changes may make the contribution more adaptable, configurable, and extensible.

Agreed!  Thank you for the review.

marianotepper (approved):
LGTM

PR #460, opened 2025-04-21
Perf metrics improvement v1.1

marianotepper:
This PR extends #459 in a few simple ways:
- It streamlines the way we specify which metrics to run and print
- Enables specifying the number of decimal points for each benchmark
- It simplifies the class structure
- RecallBenchmark was replaced by AccuracyBenchmark that additionally allows to print the mean average precision.
- LatencyBenchmark now returns the standard deviation instead of the latency to have all units in ms.


jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/AbstractQueryBenchmark.java
```diff
@@ -16,8 +16,20 @@
...
-public interface BenchmarkSummary { }
+public abstract class AbstractQueryBenchmark implements QueryBenchmark {
+    private String fmt;
+
+    AbstractQueryBenchmark(String fmt) {
+        this.fmt = fmt;
+    }
+
+    @Override
+    public void setPrintPrecision(String fmt) {
```
tlwillke: Can we return a QueryBenchmark here for chaining?  See other comments.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ExecutionTimeBenchmark.java
```diff
@@ -83,9 +63,6 @@ public Summary runBenchmark(
         }
 
         double avgRuntimeSec = totalRuntime / queryRuns / 1e9;
-        return new Summary(avgRuntimeSec);
+        return List.of(Metric.of("QPS", getPrintPrecision(), avgRuntimeSec));
```
tlwillke: The string should be "Avg Runtime (s)"

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/QueryBenchmark.java
```diff
@@ -18,20 +18,25 @@
...
-    T runBenchmark(
+    List<Metric> runBenchmark(
             ConfiguredSystem cs,
             int topK,
             int rerankK,
             boolean usePruning,
             int queryRuns
     );
+
+    void setPrintPrecision(String fmt);
```
tlwillke: If you like the idea of chaining setPrintPrecision() in Grid, we just need a return type of QueryBenchmark here.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -376,50 +376,18 @@ private static void testConfiguration(ConfiguredSystem cs,
         int queryRuns = 2;
         System.out.format("Using %s:%n", cs.index);
         // 1) Select benchmarks to run
-        var benchmarks = List.of(
+        List<QueryBenchmark> benchmarks = List.of(
```
tlwillke, line 379: Would like to be able to chain the .setPrintPrecision on any of these benchmarks here when instantiated.  See two other comments on this.

tlwillke, line 379: Suggest a slight default (in Grid) reordering of the benchmarks into kind of a canonical order when printed in the table: indep variable(s), tput, latency, count(s), recall.  Up to you.  Just looks a little random now.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/AbstractQueryBenchmark.java
tlwillke: Add return type QueryBenchmark.
return this;

tlwillke: I like the way you are suggesting we set and get the precision.  The only drawback I see is that it was previously set at the value level within a benchmark.  That way, it could be set individually per value, not globally per benchmark.  But not a big deal if each benchmark limits its produced values to ones sharing the same units and approx ranges of values.

PR #461, opened 2025-04-23
Better controls for setting string formats for the benchmark metrics

marianotepper:
Improved setup for setting up which benchmark metrics to display when running Bench.

Now, we have two options:
1- Setting everything with the default format:
```
List<QueryBenchmark> benchmarks = List.of(
        ThroughputBenchmark.createDefault(2, 0.1),
        LatencyBenchmark.createDefault(),
        CountBenchmark.createDefault(),
        AccuracyBenchmark.createDefault(),
        ExecutionTimeBenchmark.createDefault()
);
```
3- Setting everything from scratch:
```
List<QueryBenchmark> benchmarks = List.of(
        ThroughputBenchmark.createDefault(2, 0.1)
                .setFormat(".4f"),
        LatencyBenchmark.createEmpty()
                .displayAvgLatency(".4f")
                .displayLatencySTD(".4f")
                .displayP999Latency(".4f"),
        CountBenchmark.createEmpty()
                .displayAvgNodesVisited(".4f")
                .displayAvgNodesExpanded(".4f")
                .displayAvgNodesExpandedBaseLayer(".4f"),
        AccuracyBenchmark.createEmpty()
                .displayRecall(".4f")
                .displayMAP(".4f"),
        ExecutionTimeBenchmark.createDefault()
                 .setFormat(".4f")
);
```
These two options can be mixed-and-matched too. For example,
`LatencyBenchmark.createDefault().displayP999Latency(".4f")` works as expected.

Additionally, we can do `LatencyBenchmark.createEmpty().displayAvgLatency().displayLatencySTD(".4f")`  that will display the average latency with the default format and the standard deviation with the specified one. This is just an example but it applies to all classes and metrics.

For the classes that do have a `createEmpty` method, `runBenchmark` will throw an exception if no metrics are selected.

The two classes that are slightly different are `ThroughputBenchmark` and `ExecutionTimeBenchmark`. Since they report a single metric, there's no point in having `createEmpty` methods.

Thanks to @tlwillke  for his suggestions in #460 that were the bases for this design.

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
@@ -377,10 +377,10 @@ public class Grid {
         System.out.format("Using %s:%n", cs.index);
         // 1) Select benchmarks to run
         List<QueryBenchmark> benchmarks = List.of(
-                new ThroughputBenchmark(2, 0.1),
-                new LatencyBenchmark(),
-                new CountBenchmark(),
-                new AccuracyBenchmark()
+                ThroughputBenchmark.createDefault(2, 0.1),
+                LatencyBenchmark.createDefault(),
+                CountBenchmark.createDefault(),
+                AccuracyBenchmark.createDefault()
         );
         QueryTester tester = new QueryTester(benchmarks);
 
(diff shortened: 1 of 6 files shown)


tlwillke (approved):
The new ergonomics look great.

PR #462, opened 2025-04-30
Bench improvements with config files

marianotepper:
This PR introduces a few ease-of-use tools in jvector-examples. These tools are mainly related to codeless benchmarking through the use of config files for setting hyperparameters.

1. Bench now loads the list of available datasets from a [YAML](https://yaml.org/) file. The list is provided in `jvector-examples/yaml-configs/datasets.yml`. The datasets in `datasets.yml` can be commented to run a subset, for example. Additionally, Bench now allows to specify a list of overquery values for each topK value.
2. It creates `BenchYAML` that allows to read config files with JVector hyperparameters in [YAML](https://yaml.org/) format. For details on the format of these config files, see below.
3. It creates `HelloVectorWorld` with a single, clean, and simple example using ada002-100k. This example does not do a grid search over the hyperparameters, but it does try a few overquery values for topK=10 and topK=100. The corresponding config file is in `jvector-examples/yaml-configs/ada002-100k.yml`.

Here's an example YAML file showing what and how we can specify hyperparameters:
```
configVersion: 4 # do not change this number unless you know what you are doing

dataset: ada002-100k # needs to match one of the names in jvector-examples/yaml-configs/datasets.yml

construction:
  outDegree: [32, 48, 64, 96, 128]
  efConstruction: [60, 80, 100, 120, 160, 200, 400, 600, 800]
  neighborOverflow: [1.2f, 2.0f]
  addHierarchy: [No, Yes]
  compression:
    - type: None
    - type: PQ
      parameters:
        m: 192 # we can either specify the integer m or the integer mFactor. In this case, m will be set to the data dimensionality divided by mFactor
        # mFactor: 8
        k: 256 # optional parameter. By default, k=256
        centerData: No # optional parameter. By default, centerData=Yes if the dataset similarity function is Euclidean
        anisotropicThreshold: -1.0 # optional parameter. By default, anisotropicThreshold=-1 (i.e., no anisotropy)
    - type: PQ
      parameters: # here, the default values of k, centerData, and anisotropicThreshold are used
        mFactor: 2
  reranking:
    - FP
    - NVQ
  useSavedIndexIfExists: Yes

search:
  topKOverquery:
    # the value of topK followed by a list with the overquery rates we want to cover for that topK
    10: [1.0, 2.0, 5.0, 10.0]
    100: [1.0, 2.0]
  useSearchPruning: [No, Yes]
  compression:
    - type: None
    - type: PQ
      parameters:
        m: 192 # we can either specify the integer m or the integer mFactor.
        k: 256 # optional parameter. By default, k=256
        centerData: No # optional parameter. By default, centerData=Yes if the dataset similarity function is Euclidean
        anisotropicThreshold: -1.0 # optional parameter. By default, anisotropicThreshold=-1 (i.e., no anisotropy)
```


tlwillke (approved):
Looks good.

PR #466, opened 2025-05-06
Factorize the computation of diverse edges

marianotepper:
This PR does not change anything algorithmically but it separates the computation of diversity from the graph (i.e., ConcurrentNeighborMap). It is only meant to improve code clarity and separate responsibilities.


tlwillke (approved):
+1.  Thanks for cleaning things up.

PR #467, opened 2025-05-12
Reduce object creation in BSP#pqBuildScoreProvider

michaeljmarshall:
In a profile of the latest released version of jvector, I can see that during graph construction, we create many byte sequence slices due to the logic in the `pqBuildScoreProvider` method. We can leverage the additions from #403 to reduce allocations. In my testing, this code was responsible for ~17% of the total allocations during a 30 second profile.


michaeljmarshall:
I found an even better solution that will result in a reduction of 45% of allocations and renders this change unnecessary. I'll post a PR tomorrow.

PR #469, opened 2025-05-13
avx512 test runners, jobs, and assertions

jshook:
This completes bring-up of a test path for both panama and native test modes.
A summary of how it works:
* A separate arch-specific test matrix is run focusing only on ubuntu runners with avx512+ support
* The job verifies that the ISA has all the feature flags that are tested for in our native code stubs.
* GCC is installed, and the profiles which build the native library are used to build it fresh
* For native access mode (in Java 24)
  * A single base test is run to ensure that the native provider is active with the command options.
  * All tests in jvector-tests are run with the same options and assertion.
* For Panama mode (in java 20)
  * A single base test is run to ensure that the panama provider is active with the given options.
  * All tests in jvector-tests are run with the same options and assertion.

Additionally, the unit test jobs have been given a summary output that makes it easy to jump directly to the cause of any errors without going into the logs. Otherwise, a green "all good" indicator is given for each job summary.

An example that illustrates all this is here: https://github.com/datastax/jvector/actions/runs/15006463091

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -51,7 +51,7 @@ jobs:
       - name: Verify Panama Vector Support (JDK ${{ matrix.jdk }})
         if: matrix.jdk == '20'
         run: >-
-          mvn -X -B -Pjdk20 -pl jvector-tests -am test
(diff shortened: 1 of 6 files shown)


tlwillke (approved):
Looks good.  Thanks for the contribution!

PR #477, opened 2025-06-16
Improve the efficiency and memory usage of GraphIndexBuilder.cleanup

marianotepper:
This PR does a couple of simple changes that improve the efficient of GraphIndexBuilder.cleanup:
- GraphIndexBuilder.improveConnections was doing a second pass over every node trying to improve its connectivity. However, it does not seem to have an effect on recall in higher dimensions (tested on all NB datasets). We preserve its use in 2D as it was the originally intent of the method, but it should be revised at some point.
- We avoid creating a list of live nodes in GraphIndexBuilder.removeDeletedNodes by directly sampling without replacement. This removes a big allocation if the index is very large.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/disk/MappedChunkReader.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/disk/MappedChunkReader.java
@@ -1,0 +1,308 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.github.jbellis.jvector.disk;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.channels.FileChannel;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+
+/**
+ * {@code MappedChunkReader} provides random access reading of large files using memory-mapped I/O,
+ * supporting files larger than 2GB by mapping them in manageable chunks.
+ * <p>
+ * This class implements {@link RandomAccessReader} and allows reading primitive types and arrays
+ * from a file channel, handling chunk remapping transparently.
+ * <p>
+ * This class is intended as a replacement for {@link SimpleMappedReader} to provide the capability
+ * to handle files larger than 2GB in size regardless of the OS or JDK in use.
+ * </p>
+ */
+public class MappedChunkReader implements RandomAccessReader {
+    private static final long CHUNK_SIZE = Integer.MAX_VALUE; // ~2GB
+    private final FileChannel channel;
+    private final long fileSize;
+    private final ByteOrder byteOrder;
+    private long position;
+
+    private ByteBuffer currentBuffer;
+    private long currentChunkStart;
+
+    /**
+     * Constructs a new {@code MappedChunkReader} for the given file channel and byte order.
+     *
+     * @param channel   the file channel to read from
+     * @param byteOrder the byte order to use for reading
+     * @throws IOException if an I/O error occurs
+     */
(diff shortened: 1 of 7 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -356,8 +356,11 @@ public void cleanup() {
             return;
         }
 
-        // improve connections on everything in L1
-        if (graph.getMaxLevel() > 0) {
+        if (dimension == 2 && graph.getMaxLevel() > 0) {
```
sam-herman: Should we consider using KD trees for 2D cases? Just wondering if it's worth making a special case for that.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -618,7 +614,10 @@ public synchronized long removeDeletedNodes() {
                         var R = ThreadLocalRandom.current();
                         // doing actual sampling-without-replacement is expensive so we'll loop a fixed number of times instead
                         for (int i = 0; i < 2 * graph.getDegree(level); i++) {
-                            int randomNode = liveNodes.get(R.nextInt(liveNodes.size()));
+                            int randomNode = R.nextInt(graph.getIdUpperBound());
```
sam-herman, line 692: nit: it's a good practice to use `final` similar notation to c++ `const`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -559,13 +562,6 @@ public synchronized long removeDeletedNodes() {
         if (nRemoved == 0) {
             return 0;
         }
-        // make a list of remaining live nodes
```
sam-herman, line 562: As I was looking in the PQ issue, I found that our tests might not cover a situation of missing connectivity in the resulting graph.
Might want to add a test for that to verify that aside of recall being the same for predefined parameters our graph remains fully traversable by beam search.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
marianotepper: Eventually, we should stop supporting 2d vectors (remove the soecial case) and use a more appropriate data structure (a quadtree, kdtree, etc.). Until we have that, I would keep this.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
marianotepper: Maybe I'm missing something but since we are rewriting the int, it cannot be final.
```
int randomNode = R.nextInt(graph.getIdUpperBound());
while(toDelete.get(randomNode)) {
    randomNode = R.nextInt(graph.getIdUpperBound());
}
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
tlwillke: I agree about removing support for 2d.  Let's leave this for now.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
tlwillke: It can't be final.  So, I'll mark this as resolved.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
marianotepper: Ensuring traversable graphs is not trivial and testing for it would only cover a few special cases. I think that we should skip this one.

marianotepper: Ensuring navigability in practice is quite tricky. I understand the rationale behind this request, but practical indices are usually not fully navigable anyway.

marianotepper:
@tlwillke found a few cases in which this yields slightly degraded recall when rerankK=topK (analogously, overquery=1.0). I confirmed those results. When using more backtracking, this difference goes away.

I propose shelving this PR until we have separate procedures for building initial indices and for merging existing indices. This will play a role there.

sam-herman (approved):

tlwillke (approved):
Good with it.  Thanks for making it optional.

PR #483, opened 2025-06-26
Fix issue when calling cleanup while concurrently executing searches

marianotepper:
A NullPointerException that was thrown when we were:
1. deleting nodes
2. running cleanup (that commits the deletions) while concurrently executing searches

This is because there is no good isolation between a FrozenView and the commit of the deletions. If a deleted node was encountered during a concurrent search, its neighborhood was empty and caused the NPE.

This PR fixes this problem by simply returning an empty neighborhood for those nodes.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodesIterator.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodesIterator.java
@@ -96,4 +96,21 @@ public interface NodesIterator extends PrimitiveIterator.OfInt {
             return cur < size;
         }
     }
+
+    class EmptyNodeIterator implements NodesIterator {
+        @Override
+        public int size() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public int nextInt() {
+            return Integer.MIN_VALUE;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return false;
+        }
+    }
 }
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
@@ -366,40 +366,60 @@ public class OnHeapGraphIndex implements GraphIndex {
 
         @Override
         public NodesIterator getNeighborsIterator(int level, int node) {
-            var it = getNeighbors(level, node).iterator();
-            return new NodesIterator() {
-                int nextNode = advance();
+            NodesIterator it;
+            try {
+                it = getNeighbors(level, node).iterator();
+                return new NodesIterator() {
+                    int nextNode = advance();
 
-                private int advance() {
-                    while (it.hasNext()) {
-                        int n = it.nextInt();
-                        if (completions.completedAt(n) < timestamp) {
-                            return n;
+                    private int advance() {
+                        while (it.hasNext()) {
+                            int n = it.nextInt();
+                            if (completions.completedAt(n) < timestamp) {
+                                return n;
+                            }
                         }
+                        return Integer.MIN_VALUE;
                     }
-                    return Integer.MIN_VALUE;
-                }
 
-                @Override
-                public int size() {
-                    throw new UnsupportedOperationException();
-                }
-
-                @Override
-                public int nextInt() {
-                    int current = nextNode;
-                    if (current == Integer.MIN_VALUE) {
-                        throw new IndexOutOfBoundsException();
+                    @Override
+                    public int size() {
+                        throw new UnsupportedOperationException();
                     }
-                    nextNode = advance();
-                    return current;
-                }
 
-                @Override
-                public boolean hasNext() {
-                    return nextNode != Integer.MIN_VALUE;
-                }
-            };
+                    @Override
+                    public int nextInt() {
+                        int current = nextNode;
+                        if (current == Integer.MIN_VALUE) {
+                            throw new IndexOutOfBoundsException();
+                        }
+                        nextNode = advance();
+                        return current;
+                    }
+
+                    @Override
+                    public boolean hasNext() {
+                        return nextNode != Integer.MIN_VALUE;
+                    }
+                };
+            } catch (NullPointerException e) {
+                return new NodesIterator() {
+                    @Override
+                    public int size() {
+                        throw new UnsupportedOperationException();
+                    }
+
+                    @Override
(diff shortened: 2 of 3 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -366,47 +366,56 @@ public class ConcurrentGraphIndexView extends FrozenView {
...
-                int nextNode = advance();
-
-                private int advance() {
-                    while (it.hasNext()) {
-                        int n = it.nextInt();
-                        if (completions.completedAt(n) < timestamp) {
-                            return n;
+            NodesIterator it;
+            try {
+                it = getNeighbors(level, node).iterator();
```
michaeljmarshall: Instead of using exceptions as control flow for an expected event, let's branch based on whether the result of getNeighbors is null.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodesIterator.java
```diff
@@ -96,4 +96,21 @@ public boolean hasNext() {
             return cur < size;
         }
     }
+
+    class EmptyNodeIterator implements NodesIterator {
+        @Override
+        public int size() {
+            throw new UnsupportedOperationException();
```
michaeljmarshall: Wouldn't the size just be 0?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -366,47 +366,55 @@ public class ConcurrentGraphIndexView extends FrozenView {
...
+                    @Override
+                    public boolean hasNext() {
+                        return nextNode != Integer.MIN_VALUE;
+                    }
+                };
+            } else {
+                // This case may be encountered when concurrently searching while GraphIndexBuilder.cleanup is running.
+                // In such a case, the neighborhoods may get temporarily broken and we just assume that the node has
+                // no neighbors.
+                return new NodesIterator.EmptyNodeIterator();
```
michaeljmarshall: It's minor, but I think we can make this a final static object in the `NodesIterator` to skip object allocation. Given the frequency we hit this code block, it probably doesn't matter much, but in general, we've avoided unnecessary allocation on the query path, so I want to mention it here.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -366,47 +366,55 @@ public class ConcurrentGraphIndexView extends FrozenView {
...
-            var it = getNeighbors(level, node).iterator();
-            return new NodesIterator() {
-                int nextNode = advance();
-
-                private int advance() {
-                    while (it.hasNext()) {
-                        int n = it.nextInt();
-                        if (completions.completedAt(n) < timestamp) {
-                            return n;
+            NodesIterator it = getNeighbors(level, node).iterator();
```
michaeljmarshall: This will throw the NPE still, right? `getNeighbors(level, node)` returns `null`, and we need to check that output before calling the `iterator()` method on it.

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestConcurrentReadWriteDeletes.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Test
+    public void testConcurrentReadsWritesDeletes() throws ExecutionException, InterruptedException {
+        var vv = ravv.threadLocalSupplier();
+
+        testConcurrentOps(i -> {
+            var R = ThreadLocalRandom.current();
+            if (R.nextDouble() < 0.2 || keysInserted.isEmpty())
+            {
+                writeLock.lock();
```
michaeljmarshall, line 87: Very minor nit: in order to limit the amount of blocking in this code and to increase the probability of concurrent queries during cleanup, we could modify this code to only enter into this branch if we get true from `writeLock.tryLock()`.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -366,47 +366,55 @@ public class ConcurrentGraphIndexView extends FrozenView {
 
         @Override
         public NodesIterator getNeighborsIterator(int level, int node) {
-            var it = getNeighbors(level, node).iterator();
```
michaeljmarshall, line 369: Did you review all other usages of `getNeighbors()`? I see some usages that use the result without checking if it is `null`. Would it be valid to return an empty `Neighbors` object from the `getNeighbors` method and therefore change the contract to simplify the code that calls this method?

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestConcurrentReadWriteDeletes.java
marianotepper: I added a comment for a possible future improvement of the test.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
marianotepper: This is a really good point. I did this and I think that the code is actually more clear now.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodesIterator.java
```diff
@@ -96,4 +96,23 @@ public boolean hasNext() {
...
+
+    class EmptyNodeIterator implements NodesIterator {
+        @Override
+        public int size() {
+            return 0;
+        }
+
+        @Override
+        public int nextInt() {
+            return Integer.MIN_VALUE;
```
jkni: contractually, I think it's more appropriate to throw NoSuchElementException here (https://docs.oracle.com/javase/8/docs/api/java/util/PrimitiveIterator.OfInt.html#nextInt--)

marianotepper: Good catch. I fixed it there and in more more place, where an IndexOutOfBoundsException was thrown instead.

michaeljmarshall (approved):
LGTM

PR #491, opened 2025-07-02
Fix bad assert hit on CC cluster

tjake:
The assert in this PR was hit on a test Cassandra cluster.  Looking at it it doesn't make sense to me as the size of the layer will not correspond to the max nodeId in the layer.


marianotepper:
Thanks for catching the bug.

However, I believe that the right assertion should be:
`assert nodeId >= 0 && nodeId < idUpperBound`
This way, we check both ends of the range.

marianotepper:
I have added a new commit with the change I proposed above

tlwillke:
@tjake can you please try this fix with your test case?  Thanks!

PR #492, opened 2025-07-03
AUX counters and correlated benchmarks

sam-herman:
### Overview

Today JMH benchmarks only measure `ops/timeunit` however we often want to know associated characteristics that can be correlated to the run.
Examples would be:
1. **Recall** - when we are looking into query `ops/ms` we would have to associate with recall to infer about the quality of the result.
2. **expandedCountBaseLayer** - we would like to gauge on the amount of potential IO the run will induce.

In this change we are introducing AUX counters which hold that secondary information as described above and also include a simple script to reformat the secondary metrics back into the main table for presentation.

### Changes

1. Add `RecallWithRandomVectorsBenchmark` and it's AUX metrics
2. Add formatter script
3. Updated readme

### Testing

1. Running the benchmark and generating a result export:

```shell
VERSION="4.0.0-beta.6"
BENCHMARK_NAME="RecallWithRandomVectorsBenchmark"
mvn clean install -DskipTests=true
java --enable-native-access=ALL-UNNAMED \
  --add-modules=jdk.incubator.vector \
  -XX:+HeapDumpOnOutOfMemoryError \
  -Xmx20G -Djvector.experimental.enable_native_vectorization=true \
  -jar benchmarks-jmh/target/benchmarks-jmh-${VERSION}-SNAPSHOT.jar $BENCHMARK_NAME -rf csv -rff results.csv
```

2. Reformatting

Create local python virtual environment
```shell
python3 -m venv .venv
source .venv/bin/activate

# Install pandas dependency
pip install pandas
```

Run the formatter script:
```shell
# Run the formatter script (assumes results.csv is in the current directory)
python benchmarks-jmh/scripts/jmh_results_formatter.py
```

Results:
```shell
 k  PQ_Subspaces  Time_ms  Recall  ReRanked_Count  Visited_Count  Expanded_Count_BaseLayer
50             0    4.616     1.0             0.0          908.1                      56.4
50            16    4.789     0.4           250.0         2838.5                     253.2
50            32    5.355     0.5           250.0         2832.8                     251.7
50            64    5.395     0.7           250.0         2826.5                     250.6
50            96    7.495     0.8           250.0         2882.2                     260.1
50           192   10.861     1.0           250.0         3626.4                     269.3
```


benchmarks-jmh/README.md
```diff
@@ -69,6 +69,71 @@ You can just do the following to set M and beamWidth:
 VERSION="4.0.0-beta.6"
 java -jar benchmarks-jmh/target/benchmarks-jmh-${VERSION}-SNAPSHOT.jar IndexConstructionWithStaticSetBenchmark -p M=32 -p beamWidth=100 
 ```
+### Running benchmarks with auxiliary counters
+
+For benchmarks that include auxiliary counters (like `RecallWithRandomVectorsBenchmark`), run with CSV output to capture all metrics:
+
+```shell
+VERSION="4.0.0-beta.6"
```
tlwillke: Do you want to leave this as an example that the user will update before running?  Or provide something that pulls the revision from the pom.xml?  We're at 4.0.0-rc.2-SNAPSHOT now.

sam-herman: sounds good, will add a simple script we can run to populate this and will put a value in example as commented

PR #498, opened 2025-07-09
SimdOps and NativeSimd ops refactored to remove duplicate code, Vecto…

tjake:
…rUtilSupport simplified for Native vs Panama


jvector-native/src/main/java/io/github/jbellis/jvector/vector/cnative/NativeSimdOps.java
```diff
@@ -100,9 +100,9 @@ private static class check_compatibility {
         public static final FunctionDescriptor DESC = FunctionDescriptor.of(
             NativeSimdOps.C_BOOL    );
 
-        public static final MethodHandle HANDLE = Linker.nativeLinker().downcallHandle(
```
tjake, line 103: I think I must have a newer version of jextract?

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -18,122 +18,129 @@
...
 import io.github.jbellis.jvector.vector.types.ByteSequence;
 import io.github.jbellis.jvector.vector.types.VectorFloat;
+import jdk.incubator.vector.FloatVector;
+import jdk.incubator.vector.VectorSpecies;
 
 import java.util.List;
 
-final class PanamaVectorUtilSupport implements VectorUtilSupport {
+class PanamaVectorUtilSupport implements VectorUtilSupport {
+    protected final SimdOps simdOps;
```
sam-herman: Is it needed to keep here an instance of `SimdOps`? It looks like previously it was static which is more efficient when no state is required to keep for method calls.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/SimdOps.java
```diff
@@ -26,24 +26,50 @@
...
+class SimdOps {
     static final int PREFERRED_BIT_SIZE = FloatVector.SPECIES_PREFERRED.vectorBitSize();
     static final IntVector BYTE_TO_INT_MASK_512 = IntVector.broadcast(IntVector.SPECIES_512, 0xff);
     static final IntVector BYTE_TO_INT_MASK_256 = IntVector.broadcast(IntVector.SPECIES_256, 0xff);
 
     static final ThreadLocal<int[]> scratchInt512 = ThreadLocal.withInitial(() -> new int[IntVector.SPECIES_512.length()]);
     static final ThreadLocal<int[]> scratchInt256 = ThreadLocal.withInitial(() -> new int[IntVector.SPECIES_256.length()]);
 
-    static float sum(ArrayVectorFloat vector) {
+    protected FloatVector fromVectorFloat(VectorSpecies<Float> SPEC, VectorFloat<?> vector, int offset) {
```
sam-herman: No local member variables in this class, why not make methods static?

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
marianotepper: Since the PR is using inheritance and function overloading for `fromVectorFloat`, I think we need runtime method resolution.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/SimdOps.java
marianotepper: The method `fromVectorFloat` is solved at runtime using inheritance and overloading. Unless I'm missing a key Java feature, I think the methods cannot be static.

tjake: Correct, we need the inheritance

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
tjake: We don't need separate SimdOps classes etc, we can move things to VectorSupportUtils implementations?

marianotepper: That's not a bad idea. It might work.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/SimdOps.java
```diff
@@ -26,24 +26,50 @@
 import jdk.incubator.vector.ShortVector;
 import jdk.incubator.vector.VectorMask;
 import jdk.incubator.vector.VectorOperators;
+import jdk.incubator.vector.VectorSpecies;
 
+import java.nio.ByteOrder;
 import java.util.List;
 
-final class SimdOps {
+class SimdOps {
```
marianotepper: Minor suggestion: Maybe having SimdOps as a singleton class would not be a bad idea

jvector-native/src/main/java/io/github/jbellis/jvector/vector/cnative/NativeSimdOps.java
jkni: These are generated with a version from February 2024, so very possible. Maybe we should make a note of the version used in the README so that we can avoid unnecessary churn if someone uses an old version again?

tjake:
Consensus seems to be we should remove the SimdOps stuff and move it into VectorSupport classes.  I'll do that

marianotepper:
@MarkWolters is running some benchmarks on this branch to see if it is as performant as the previous version.

tjake:
Ok I reverted the jmh change and moved SimdOps and NativeSimdOps logic to VectorUtilSupport

sam-herman (approved):

PR #499, opened 2025-07-14
GitHub actions regression test

MarkWolters:
This PR introduces a new github workflow and associated code to perform automated regression testing across branches. The workflow can be triggered manually with the branches to compare included as inputs. It is also triggered to run automatically when a PR to main is opened, in which case it will run a regression test comparing the requested branch with main. 

Metrics compared include QPR, average latency, and recall at k10. See attached for an example of the output.
[benchmark_report.zip](https://github.com/user-attachments/files/21237766/benchmark_report.zip)


visualize_benchmarks.py
```diff
@@ -0,0 +1,315 @@
...
+from typing import Dict, List, Any, Tuple, Optional
+import matplotlib.pyplot as plt
+import numpy as np
+from matplotlib.ticker import MaxNLocator
+
+# Define metrics where higher values are better and lower values are better
+HIGHER_IS_BETTER = ["QPS", "Recall@10"]
+LOWER_IS_BETTER = ["Mean Latency"]
+
+class BenchmarkData:
```
marianotepper, line 36: I suggest leaving two blank lines between top-level functions and two blank lines between the import statements and other code: https://stackoverflow.com/questions/2953250/python-pep8-blank-lines-convention

jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
```diff
@@ -0,0 +1,183 @@
+/*
```
marianotepper, line 1: Should we tidy up the main example directory? IMO, it is getting polluted. Maybe we should add a "tools" directory or similar. We can do a separate PR for this, no need to do it here as there are other files that should probably go there too and that have nothing to do with this PR.

visualize_benchmarks.py
MarkWolters: Done and commited

jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
MarkWolters: Noted and agreed

marianotepper (approved):
Looks ready to merge. We can continue improving this feature as we use it, but it is a solid starting point.

.github/workflows/run-bench.yml
```diff
@@ -0,0 +1,245 @@
...
+          fi
+
+          # Check if the jar with dependencies was built
+          JAR_COUNT=$(ls jvector-examples/target/jvector-examples-*-jar-with-dependencies.jar 2>/dev/null | wc -l)
+          if [ "$JAR_COUNT" -eq 0 ]; then
+            echo "Warning: No jar with dependencies found in branch ${{ matrix.branch }}. Skipping benchmark."
+            exit 0
+          fi
+
+          # Determine available memory and set heap size to half of it
```
tlwillke, line 145: Double check that this setting will take priority over the Xmx set in the module's pom.xml (statically set to 14GB).

tlwillke (approved):
LGTM

marianotepper (approved):

PR #500, opened 2025-07-15
Fix dataset naming when using default.yml

marianotepper:
The config file default.yml had the wrong name. Fixing the piping code so that the correct name is applied.


tlwillke (approved):
Looks good.

PR #501, opened 2025-07-17
Reducing the number of allocations in GraphSearcher

marianotepper:
In GraphSearcher::search, we were instantiating one ScoreTracker each time we called search. This PR creates a ScoreTrackerFactory class that allows instantiating once and reusing the ScoreTracker object for each search.

Additionally, it introduces some internal refactoring (no functional changes) for easier understanding of the code.


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -64,6 +61,9 @@ public class GraphSearcher implements Closeable {
     private SearchScoreProvider scoreProvider;
     private CachingReranker cachingReranker;
 
+    private boolean pruneSearch;
```
sam-herman, line 64: should this be set once from constructor and become final as well?
In all instances of the code, this seems to be set only once right after the construction of the searcher.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -208,6 +225,35 @@ public SearchResult search(SearchScoreProvider scoreProvider,
             return new SearchResult(new SearchResult.NodeScore[0], 0, 0, 0, 0, Float.POSITIVE_INFINITY);
         }
 
+        internalSearch(scoreProvider, entry, topK, rerankK, threshold, acceptOrds);
+        return reranking(topK, rerankK, rerankFloor);
+    }
+
+    /**
+     * Performs a search, leaving the results in the internal member variable approximateResults.
+     * It does not perform reranking.
```
sam-herman, line 234: It's a bit confusing that it doesn't perform re-ranking and yet takes re-rank K as a parameter.
Any reason for that?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ScoreTracker.java
```diff
@@ -132,11 +185,32 @@ class RelaxedMonotonicityTracker implements ScoreTracker {
...
+            this(100);
+        }
+
+        private static int getRecentScoresSize(int bestScoresTracked) {
             // A quick empirical study yields that the number of recent scores
             // that we need to consider grows by a factor of ~sqrt(bestScoresTracked / 2)
             int factor = (int) Math.round(Math.sqrt(bestScoresTracked / 2.0));
-            this.recentScores = new double[200 * factor];
-            this.bestScores = new BoundedLongHeap(bestScoresTracked);
+            return 200 * factor;
```
sam-herman: It was already there before, but curious where did the number `200` came from?
Also just as a nit should we move it to a static parameter to clarify it's intention?
For example:
```java
private static int DEFAULT_SOMETHING_SOMETHING = 200
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ScoreTracker.java
```diff
@@ -16,13 +16,54 @@
...
+
+        ScoreTrackerFactory() {
+            twoPhaseTracker = null;
+            relaxedMonotonicityTracker = null;
+            noOpTracker = null;
+        }
+
+        public ScoreTracker getScoreTracker(boolean pruneSearch, int rerankK, float threshold) {
+            // track scores to predict when we are done with threshold queries
+            ScoreTracker scoreTracker;
```
sam-herman: nit: use final to help compiler make sure it's assigned at the end of the if conditions.

sam-herman (approved):
Mostly just need comments, overall looks good!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
marianotepper: We need to know the length of the search queue, which is given by rerankK. The length of the search queue determines the amount of backtracking done during search. Even without reranking, it improves the results in general (that is, the search is less greedy).

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ScoreTracker.java
marianotepper: It was found empirically. TBH, there is not much behind it. I will make this change, I agree that it makes the code more maintainable.

tlwillke (commented):
LGTM.  I asked @MarkWolters to run the perf regression test on it.

tlwillke (approved):
Looks good with the final changes.

PR #502, opened 2025-07-18
Improvements to throughput benchmark

tlwillke:
- Warmup now uses random vectors instead of a split of the test vectors
- In addition to Avg QPS, can report Median QPS and Max QPS
- Number of warmup and test passes is configurable
- Overloaded QueryExecutor to accept random vector queries

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
@@ -16,20 +16,19 @@
 
 package io.github.jbellis.jvector.example.benchmarks;
 
+import io.github.jbellis.jvector.example.Grid.ConfiguredSystem;
+import io.github.jbellis.jvector.graph.SearchResult;
+import io.github.jbellis.jvector.vector.types.VectorFloat;
+import io.github.jbellis.jvector.vector.types.VectorTypeSupport;
+import io.github.jbellis.jvector.vector.VectorUtil;
+import io.github.jbellis.jvector.vector.VectorizationProvider;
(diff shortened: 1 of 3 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
```diff
@@ -63,40 +118,75 @@ public List<Metric> runBenchmark(
...
-                        });
-            }
+        int dim = cs.getDataSet().getDimension();
+
+        // Warmup Phase: Use randomly-generated vectors
+        for (int warmupRun = 0; warmupRun < numWarmupRuns; warmupRun++) {
+            IntStream.range(0, totalQueries)
+                    .parallel()
+                    .forEach(k -> {
+                        // Generate a random vector
```
MarkWolters, line 132: If the associated ConfiguredSystem has a Dataset already why are we generating a random vector rather than using the queryVector? I understand this works with the change to QueryExecutor, I'm just not clear on why we are making this change?

MarkWolters (approved):
Looks good, I would just like to understand why we are modifying this process to use a randomly generated vector rather than the set of query vectors from the dataset.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
tlwillke: There are often very few queries in a dataset.  Best practice is to not use the same vectors for warmup that you use for testing (due to potential caching, fitting, etc.).  Using the query set reduces the split available for testing.  Now, we are using random, since it really doesn't matter what you run through for warmup.

PR #508, opened 2025-08-21
pq ranging bugfix and refactoring

jshook:
This fixes a bug which affected how PQ chunks were arranged. It was originally an off-by-one error, but I decided to improve the general structure of it by introducing a cleaner PQLayout type. This centralizes calculations for layout into a value type with some computed fields.

The off-by-one error was where the total number of chunks was being calculated duplicitously, correctly in one place and incorrectly in another. The new code effectively makes the calculation DRY and much easier to understand. And it's not in any critical performance area. By short-changing the number of chunks, the last chunk would end up with more than a chunk should have, and this would cause an erroneous buffer calculation bigger than (2^32)-1 (edited)

[UPDATE]
A few key improvements since the last push:
* The anecdotal PQVectors.MAX_CHUNK_SIZE has been removed. It was only used as a sizing hint. Instead, the PQ chunk layout is calculated with a safe margin for buffer indices.
* the ctor structure is more obvious. Apart from the two ctor params and a couple intermediate values, everything else is a final field. So I dropped the duplicitous _this._ everywhere.
* Unit test fixes, expansions
* A separate boundary condition test was added to illustrate layout parameters under different corner cases.

[LATEST UPDATE]
Made PQLayout package private [And an inner class]


jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
```diff
@@ -0,0 +1,68 @@
...
+        if (fullChunkVectors == 0) {
+            throw new IllegalArgumentException("Compressed dimension " + compressedDimension + " too large for chunking");
+        }
+        this.lastChunkVectors = vectorCount % this.fullChunkVectors;
+
+        this.fullChunkBytes = Math.multiplyExact(compressedDimension, this.fullChunkVectors);
+        this.lastChunkBytes = Math.multiplyExact(compressedDimension, lastChunkVectors);
+
+        this.fullSizeChunks = vectorCount / fullChunkVectors;
+        this.totalChunks = fullSizeChunks + ((vectorCount % fullChunkVectors == 0) ? 0 : 1);
```
marianotepper: Should this be this.fullSizeChunks to make it easier to follow?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package io.github.jbellis.jvector.quantization;
+
+/**
+ * Chunk Dimensions and Layout
+ * This is emulative of modern Java records, but keeps to J11 standards.
+ * This class consolidates the layout calculations for PQ data into one place
+ */
+public class PQLayout {
```
marianotepper: Instead of a public class in a separate file, should this not be a private subclass of PQVectors? They are coupled anyway... We can expose it for testing with the VisibleForTesting tag if needed.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQVectors.java
```diff
@@ -120,18 +88,17 @@ public static PQVectors load(RandomAccessReader in, long offset) throws IOExcept
...
-        int chunkSize = vectorsPerChunk * compressedDimension;
-        for (int i = 0; i < numChunks - 1; i++)
-            chunks[i] = vectorTypeSupport.createByteSequence(chunkSize);
+//        long totalSize = (long) vectorCount * compressedDimension;
+//        int vectorsPerChunk = totalSize <= PQVectors.MAX_CHUNK_SIZE ? vectorCount : PQVectors.MAX_CHUNK_SIZE / compressedDimension;
 
-        // Last chunk might be smaller
-        int remainingVectors = vectorCount - (vectorsPerChunk * (numChunks - 1));
-        chunks[numChunks - 1] = vectorTypeSupport.createByteSequence(remainingVectors * compressedDimension);
+        PQLayout dims = new PQLayout(vectorCount,compressedDimension);
```
marianotepper: Should we use `layout` instead of `dims` for clarity of intent?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /** total number of vectors **/
+    public final int vectorCount;
```
marianotepper: vectorCount and compressedDimension are stored twice, in PQVectors and here. Should we drop one of them for clarity? Maybe instantiate the layout as
`PQLayout layout = new PQLayout(in.readInt(), in.readInt());`
and we drop the variables in PQVectors

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
```diff
@@ -0,0 +1,68 @@
...
+
+        long totalSize = (long) vectorCount * compressedDimension;
+
+        this.fullChunkVectors = totalSize <= PQVectors.MAX_CHUNK_SIZE ? vectorCount : PQVectors.MAX_CHUNK_SIZE / compressedDimension;
+        if (fullChunkVectors == 0) {
+            throw new IllegalArgumentException("Compressed dimension " + compressedDimension + " too large for chunking");
+        }
+        this.lastChunkVectors = vectorCount % this.fullChunkVectors;
+
+        this.fullChunkBytes = Math.multiplyExact(compressedDimension, this.fullChunkVectors);
```
marianotepper: In the next two lines, should we catch and handle the ArithmeticException from Math.multiplyExact?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
```diff
@@ -0,0 +1,68 @@
...
+
+        if (compressedDimension < 0) {
+            throw new IllegalArgumentException("Invalid compressed dimension " + compressedDimension);
+        }
+        this.compressedDimension = compressedDimension;
+
+        long totalSize = (long) vectorCount * compressedDimension;
+
+        this.fullChunkVectors = totalSize <= PQVectors.MAX_CHUNK_SIZE ? vectorCount : PQVectors.MAX_CHUNK_SIZE / compressedDimension;
+        if (fullChunkVectors == 0) {
```
marianotepper: Replace fullChunkVectors  by this.fullChunkVectors?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQVectors.java
```diff
@@ -120,18 +88,17 @@ public static PQVectors load(RandomAccessReader in, long offset) throws IOExcept
...
         int compressedDimension = pq.compressedVectorSize();
-        long totalSize = (long) vectorCount * compressedDimension;
-        int vectorsPerChunk = totalSize <= PQVectors.MAX_CHUNK_SIZE ? vectorCount : PQVectors.MAX_CHUNK_SIZE / compressedDimension;
-
-        int numChunks = vectorCount / vectorsPerChunk;
-        final ByteSequence<?>[] chunks = new ByteSequence<?>[numChunks];
-        int chunkSize = vectorsPerChunk * compressedDimension;
-        for (int i = 0; i < numChunks - 1; i++)
-            chunks[i] = vectorTypeSupport.createByteSequence(chunkSize);
+//        long totalSize = (long) vectorCount * compressedDimension;
```
marianotepper: I think that we can drop these two commented lines

marianotepper (commented):
The PR looks good in general. A few comments here and there.

MarkWolters (dismissed):
LGTM

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
jshook: I'm making the code more defensive now. Arranging in this order: 1) argument range checks 2) instance field assignment and then 3) all other logic will work exclusively on this.parameters.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
jshook: I'm a bit ambivalent about this. I'm not sure that the encapsulation of this data is beneficial in that it is effectively a memoized computation which acts as a value type. There is very little that can go wrong by sharing it, and it makes understanding that much better.

I differ on this when it comes to an API view. This should clearly be behind an API curtain more properly, but that is usually handled on a different level with different tools (and essentially by composition and designing an explicit contract layer around the moat)

For now, I'm ok leaving it visible since it has no mutable state and explains things a bit for developers.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQVectors.java
jshook: yes, updating this

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
jshook: This would likely be a developer-facing error, but I don't mind making it more clear.

jshook: Actually, I'm going to make this safe, so that if the bound is heeded with a safe ceiling no matter what.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
marianotepper: PQLayout looks to me like an implementation detail about how PQVectors stores its data. By itself, it bears nor meaning. But I understand that this is a stylistic opinion.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
jshook: I think this section of code is due for some other refactorings. I'd rather avoid putting buffer-mutating calls into an argument position in this particular case.

As another change, we should probably consider making this initialization logic a little less ad-hoc with respect to raw data and buffer pointers.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
jshook: The code structure is completely simplified since this comment. I'll resolve these and as for a new refresh on the review.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQVectors.java
jshook: done

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
jshook: I made all variable usage consistent in the next push.

tlwillke:
> The PR looks good to me. I still think that PQLayout should be a class inside PQVectors as it does not have proper meaning by itself. It computes some internal parameters that PQVectors uses to arrange its internal data structures.

I agree.  Let's incorporate PQLayout into PQVectors.

tlwillke (changes requested):
Thanks for the fix and refactor.  Looks better and the tests are very welcome!

Please address the comment about PQLayout so we can approve.

marianotepper:
In similar cases, JVector is mostly built using nested classes so I think that setting up PQLayout as a static nested class inside PQVectors would better reflect the library's style. Additionally, there is no other class inside the package that will ever want to use PQLayout, as its functionality is tightly coupled with the internal data structures in PQVectors.

marianotepper (approved):
Looks good to me. Thanks for being flexible on the class structure.

PR #510, opened 2025-08-26
Remove unused construction batch member from OnHeapGraphIndex

marianotepper:
Removes the unused class member:
`public final ConcurrentMap<NodeAtLevel, VectorFloat<?>> constructionBatch;`
from OnHeapGraphIndex.

Edit:
The constructor was also modified to not receive a batchSize argument.


michaeljmarshall (dismissed):
Confirmed we are not using this field in CC

tlwillke (approved):
LGTM

PR #512, opened 2025-08-27
proposed starting contributor guide

jshook:
Proposed project rules to clarify and streamline pull requests and merging


jshook:
We will come back to this after some planning.

PR #513, opened 2025-08-28
Vectordata streaming support for examples/bench

jshook:
[UPDATE 2025-12-04]
This is being broken apart into smaller pieces, but I'm leaving the PR here for visibility for now.

---

This extends the examples/bench capabilities in jvector:
* The DataSet type is virtualized behind an interface
* The way DataSets are loaded is more modular
   * A DataSet loader has been added to support the vectordata API

The net effect is that bench can no access vectordata hosted datasets.
The benefits of this are several:
* remote vector test data hosting
* uniform API for finding and using vector datasets
* merkle-based automatic download of chunks for dynamic access
* efficient download and automatic caching of data locally on test nodes
* mapping and ranging subsets of data, such as "first 1M", "first 10M" and so on under profile names
* packing various vector data views in a consistent API: base, query, ...
* management of datasets via catalogs, orthogonal to access control

Most of the core wiring of these capabilities is provided by another library which is part of the nosqlbench project. The changes to jvector are to adapt it to use these APIs.

There are a couple issues yet to resolve with the Java version configs and GHA.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,4 +44,6 @@ hdf5/
 dependency-reduced-pom.xml
 results.csv
 
+# Local testing files
+local/**
 
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/TestDataViewWrapper.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/TestDataViewWrapper.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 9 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/quantization/PQLayout.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+package io.github.jbellis.jvector.quantization;
+
+/**
+ * Chunk Dimensions and Layout
+ * This is emulative of modern Java records, but keeps to J11 standards.
+ * This class consolidates the layout calculations for PQ data into one place
+ */
+public class PQLayout {
```
jshook: When the other PR is resolved, I'll pull the affecting changes in here.

jshook:
Due to conflicts with the executable jar approach, I'll need to refactor this a bit, so I moved it to a draft.
My plan for this is to combine the commands so that autobench is included and update the GHA wiring.

PR #514, opened 2025-08-29
Optimizes reducelanes in diversityCalculation of PQVectors, for Euclidean function

shyla226:
This pull request introduces improvement to Euclidean similarity function in `PQVectors.diversityFunctionFor`. From flamegraph, it is observed that considerable amount of time is spent in `jdk/incubator/vector/FloatVector.reducelanesTemplate`. This is mainly because `FloatVector.reducelanes()` is expensive and it is being called inside a `for` loop (via `VectorUtil.squareL2Distance`). Modification in this pull request moves call to `reduceLanes()` outside the `for` loop.
 
Change proposed here was tested with the benchmark, `PQDistanceCalculationBenchmark.diversityCalculation`
With this benchmark, ~18% reduction in time was observed  when M=64 and ~22%  when M=192.
 
**Code modifications:**
1.	Added a new function `pqDiversityEuclidean` in `VectorUtilSupport` and its corresponding implementations
2.	Removed `for` loop in `PQVectors.diversityFunctionFor` and moved it into `pqDiversityEuclidean`
3.	Moved `FloatVector.reducelanes()` outside the `for` loop
 
**Test setup:**
Jvector version : main branch (as of 2025-08-28)
JDK version : openjdk version "24.0.2" 2025-07-15
Platform : INTEL(R) XEON(R) PLATINUM 8592+
Benchmark : PQDistanceCalculationBenchmark.diversityCalculation

**New changes**
I have modified the code to include dot product & cosine functions and implemented similar changes for scoreFunctionFor.
With the changes applied to scoreFunctionFor, when M=64: dot product shows ~30% reduction in latency & cosine shows ~43% reduction.
I can add data points for other subspace counts, if required.

**Code modifications:**

1. Added new functions `pqScoreEuclidean, pqScoreDotProduct, pqScoreCosine` in `VectorUtilSupport` and its corresponding implementations for `diversityFunctionFor`
2. Added overloaded version of above functions for `scoreFunctionFor`
3. Removed `for` loop in `PQVectors.diversityFunctionFor` and `PQVectors.scoreFunctionFor` and moved them into respective functions in `PanamaVectorUtilSupport`
4. Moved `FloatVector.reducelanes()` outside the `for` loop
5. Added a new benchmark which uses `MutablePQVectors` to test this

**Test setup:**
Jvector version : main branch (as of 2025-10-22)
JDK version : openjdk version "25.0.1" 2025-10-21
Platform : Intel(R) Xeon(R) 6979P
Benchmark :  `PQDistanceCalculationMutableVectorBenchmark` (Added this by replicating `PQDistanceCalculationBenchmark` to measure performance for MutablePQVectors)

### DIFF
--- a/jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
+++ b/jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
@@ -1543,4 +1543,1025 @@ class PanamaVectorUtilSupport implements VectorUtilSupport {
     public float pqDecodedCosineSimilarity(ByteSequence<?> encoded, int clusterCount, VectorFloat<?> partialSums, VectorFloat<?> aMagnitude, float bMagnitude) {
         return pqDecodedCosineSimilarity(encoded, 0, encoded.length(),  clusterCount, partialSums, aMagnitude, bMagnitude);
     }
+
+    /*-------------------- Score functions--------------------*/
+    //adding SPECIES_64 & SPECIES_128 for completeness, will it get there?
+    float pqScoreEuclidean_512(VectorFloat<?>[] codebooks, int[][] subvectorSizesAndOffsets, ByteSequence<?> node1Chunk, int node1Offset, ByteSequence<?> node2Chunk, int node2Offset, int subspaceCount) {
+        float res = 0;
+        FloatVector sum = FloatVector.zero(FloatVector.SPECIES_PREFERRED);
+
+        for (int m = 0; m < subspaceCount; m++) {
+            int centroidIndex1 = Byte.toUnsignedInt(node1Chunk.get(m + node1Offset));
+            int centroidIndex2 = Byte.toUnsignedInt(node2Chunk.get(m + node2Offset));
+            int centroidLength = subvectorSizesAndOffsets[m][0];
+            final int vectorizedLength = FloatVector.SPECIES_PREFERRED.loopBound(centroidLength);
+            int length1 = centroidIndex1 * centroidLength;
+            int length2 = centroidIndex2 * centroidLength;
+
+            if (centroidLength == FloatVector.SPECIES_PREFERRED.length()) {
+                FloatVector a = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length1);
+                FloatVector b = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length2);
+                var diff = a.sub(b);
+                sum = diff.fma(diff, sum);
+            }
+            else {
+                int i = 0;
+                for (; i < vectorizedLength; i += FloatVector.SPECIES_PREFERRED.length()) {
+                    FloatVector a = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length1 + i);
+                    FloatVector b = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length2 + i);
+                    var diff = a.sub(b);
+                    sum = diff.fma(diff, sum);
+                }
+                // Process the tail
+                for (; i < centroidLength ; ++i) {
+                    var diff = codebooks[m].get(length1 + i) - codebooks[m].get(length2 + i);
+                    res += MathUtil.square(diff);
+                }
+            }
+        }
+        res += sum.reduceLanes(VectorOperators.ADD);
+        return res;
+    }
+
+    float pqScoreEuclidean_256(VectorFloat<?>[] codebooks, int[][] subvectorSizesAndOffsets, ByteSequence<?> node1Chunk, int node1Offset, ByteSequence<?> node2Chunk, int node2Offset, int subspaceCount) {
+        float res = 0;
+        FloatVector sum = FloatVector.zero(FloatVector.SPECIES_256);
+
+        for (int m = 0; m < subspaceCount; m++) {
+            int centroidIndex1 = Byte.toUnsignedInt(node1Chunk.get(m + node1Offset));
+            int centroidIndex2 = Byte.toUnsignedInt(node2Chunk.get(m + node2Offset));
+            int centroidLength = subvectorSizesAndOffsets[m][0];
+            final int vectorizedLength = FloatVector.SPECIES_256.loopBound(centroidLength);
+            int length1 = centroidIndex1 * centroidLength;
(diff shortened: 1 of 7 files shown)


MarkWolters:
This PR does successfully remove the calls to reduceLanes from the inside of the loop iterating over the subspaces the Euclidean case. I do have a concern about applying this pattern to one specific case but leaving the handling of other cases as is, leaving the code with 2 different implementations for the same problem. Some questions:

1. Why is the optimization only implemented for the Euclidean case? In PanamaVectorUtilSupport, where the call to reduceLanes is made from the squareDistance methods which are called in a loop for the subspace count, likewise reduceLanes is called from the dotProduct methods. Why only fix one?

2. Why is the optimization only applied to diversityFunctionFor(…) and not scoreFunctionFor(…) which uses the same pattern of calling squareL2Distance(…) inside of a loop for each subspace?

shyla226:
@MarkWolters,  thank you very much for the feedback. I can make the changes to dotProduct methods and to function calls in scoreFunctionFor()

MarkWolters:
Thanks @shyla226, I will review the updated code asap

shyla226:
Thank you very much @MarkWolters . With the changes, when M=64, dot product shows ~30% reduction in latency & COSINE shows ~43% reduction

shyla226:
@MarkWolters Thank you very much. I will rebase the changes.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -1543,4 +1543,1025 @@ public void calculatePartialSums(VectorFloat<?> codebook, int codebookIndex, int
...
+        else if(subvectorSizesAndOffsets[0][0] >= FloatVector.SPECIES_256.length()) {
+            return pqScoreDotProduct_256(codebooks, subvectorSizesAndOffsets, node1Chunk, node1Offset, node2Chunk, node2Offset, subspaceCount);
+        }
+        else if (subvectorSizesAndOffsets[0][0]  >=  FloatVector.SPECIES_128.length()) {
+            return pqScoreDotProduct_128(codebooks, subvectorSizesAndOffsets, node1Chunk, node1Offset, node2Chunk, node2Offset, subspaceCount);
+        }
+        return pqScoreDotProduct_64(codebooks, subvectorSizesAndOffsets, node1Chunk, node1Offset, node2Chunk, node2Offset, subspaceCount);
+    }
+
+    float pqScoreCosine_512(VectorFloat<?>[] codebooks, int[][] subvectorSizesAndOffsets, ByteSequence<?> node1Chunk, int node1Offset, ByteSequence<?> node2Chunk, int node2Offset, int subspaceCount) {
```
r-devulap: This name is a mis-leading, probably better to rename this to `pqScoreCosine_preferred`?

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -1543,4 +1543,1025 @@ public void calculatePartialSums(VectorFloat<?> codebook, int codebookIndex, int
...
+
+        for (int m = 0; m < subspaceCount; m++) {
+            int centroidIndex1 = Byte.toUnsignedInt(node1Chunk.get(m + node1Offset));
+            int centroidIndex2 = Byte.toUnsignedInt(node2Chunk.get(m + node2Offset));
+            int centroidLength = subvectorSizesAndOffsets[m][0];
+            final int vectorizedLength = FloatVector.SPECIES_PREFERRED.loopBound(centroidLength);
+            int length1 = centroidIndex1 * centroidLength;
+            int length2 = centroidIndex2 * centroidLength;
+
+            if (centroidLength == FloatVector.SPECIES_PREFERRED.length()) {
```
r-devulap, line 1913: Why do we need this special case? It introduces a lot of duplicated code across all the functions. Wouldn’t the loop below already handle this scenario correctly?

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -1543,4 +1543,1025 @@ public void calculatePartialSums(VectorFloat<?> codebook, int codebookIndex, int
...
+        if(subvectorSizesAndOffsets[0][0] >= FloatVector.SPECIES_PREFERRED.length()) {
+            return pqScoreEuclidean_512(codebooks, subvectorSizesAndOffsets, node1Chunk, node1Offset, node2Chunk, node2Offset, subspaceCount);
+        }
+        else if(subvectorSizesAndOffsets[0][0] >= FloatVector.SPECIES_256.length()) {
+            return pqScoreEuclidean_256( codebooks,  subvectorSizesAndOffsets, node1Chunk, node1Offset,  node2Chunk, node2Offset, subspaceCount);
+        }
+        else if (subvectorSizesAndOffsets[0][0]  >=  FloatVector.SPECIES_128.length()) {
+            return pqScoreEuclidean_128( codebooks,  subvectorSizesAndOffsets, node1Chunk, node1Offset,  node2Chunk, node2Offset, subspaceCount);
+        }
+        return pqScoreEuclidean_64( codebooks,  subvectorSizesAndOffsets, node1Chunk, node1Offset,  node2Chunk, node2Offset, subspaceCount);
```
r-devulap: We currently have 4 separate implementations (`pqScoreEuclidean_64/_128/_256/_512`) that are structurally identical aside from the `FloatVector.SPECIES_*` used. This is a lot of duplicated hot-path code and increases the risk of fixes/optimizations diverging across variants.

Could we fold these into a single generic-ish implementation that takes the species as a parameter, e.g.:

- `pqScoreEuclideanImpl(VectorSpecies<Float> species, ...)`
- `pqScoreEuclidean(...)` selects `species` (`SPECIES_PREFERRED`, `SPECIES_256`, `SPECIES_128`, `SPECIES_64`) once and dispatches to the shared impl.

This keeps the vectorized logic in one place and avoids the confusing `_512` naming given it actually uses `SPECIES_PREFERRED` on some paths. It should also make it much easier to keep Euclidean/Dot/Cosine implementations consistent.

r-devulap (commented):
First‑pass review: there’s quite a bit of duplicated code, and I think we can consolidate it significantly.

r-devulap:
Duplicate of #623. Closing.

PR #518, opened 2025-09-12
Switch from syncronized to concurrent map for pq codebook

tjake:
This was causing lots of blocking on large cpu box

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/ImmutablePQVectors.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/ImmutablePQVectors.java
@@ -54,7 +54,7 @@ public class ImmutablePQVectors extends PQVectors {
         return vectorCount;
     }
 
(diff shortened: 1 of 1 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/quantization/ImmutablePQVectors.java
```diff
@@ -41,7 +41,7 @@ public ImmutablePQVectors(ProductQuantization pq, ByteSequence<?>[] compressedDa
         this.compressedDataChunks = compressedDataChunks;
         this.vectorCount = vectorCount;
         this.vectorsPerChunk = vectorsPerChunk;
-        this.codebookPartialSumsMap = new HashMap<>();
+        this.codebookPartialSumsMap = new ConcurrentHashMap<>();
```
FRosner, line 44: Don't you now have a synchronized ConcurrentHashMap? Shouldn't you remove the external synchronization?

tjake: oops, missed when breaking this into its own commit, thx for catching!

PR #521, opened 2025-09-17
GraphIndexBuilder::addGraphNode must iterate all graph levels to estimate used bytes

michaeljmarshall:
The current implementation for `GraphIndexBuilder::addGraphNode` has a bug in the byte estimation logic. It calls `IntStream.range(0, nodeLevel.level)`, which in turn calls:

```java
    public static IntStream range(int startInclusive, int endExclusive) {
        if (startInclusive >= endExclusive) {
            return empty();
        } else {
            return StreamSupport.intStream(
                    new Streams.RangeIntSpliterator(startInclusive, endExclusive, false), false);
        }
    }
```

This is problematic because the `nodeLevel.level` is inclusive while the right bound of the range is exclusive.

The end result is an estimate that systematically misses the numerically highest level of the graph, thereby under counting bytes used. In reviewing the `IntStream` api, it seems most appropriate to use the `rangeClosed` method to achieve the desired semantics.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -597,7 +597,7 @@ public class GraphIndexBuilder implements Closeable {
             insertionsInProgress.remove(nodeLevel);
         }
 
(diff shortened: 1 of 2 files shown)


marianotepper (approved):
Good catch! Looks good.

tlwillke (approved):
Really nice catch indeed.  LGTM

PR #522, opened 2025-09-17
Ensures that no node duplicates exist in the adjacency list of any node.

marianotepper:
Most nodes in a graph have MANY duplicated entries in their adjacency list. This is undesirable as it reduces the effective degree  of the graph.

This was occurring because scores where the main vehicle to check whether a node was inserted twice in an adjacency list. However, when we are building a graph with PQ we use a mix of `sim(x1, quant(x2))` and `sim(quant(x1), quant(x2))` similarities. Because quantization is lossy, `sim(x1, x2) != sim(x1, quant(x2)) != sim(quant(x1), quant(x2))`. Thus, we can have two different scores associated with a given node, depending on which quantized similarity we use.

This PR changes the way we are computing duplicates in NodeArray, to prevent the emergence of these duplicates.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeArray.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeArray.java
@@ -64,9 +64,8 @@ public class NodeArray {
         NodeArray merged = new NodeArray(a1.size() + a2.size());
         int i = 0, j = 0;
 
-        // since nodes are only guaranteed to be sorted by score -- ties can appear in any node order --
-        // we need to remember all the nodes with the current score to avoid adding duplicates
-        var nodesWithLastScore = new IntHashSet();
+        // To avoid duplicates, we need to remember all the nodes added so far
+        var mergedNodes = new IntHashSet();
         float lastAddedScore = Float.NaN;
 
         // loop through both source arrays, adding the highest score element to the merged array,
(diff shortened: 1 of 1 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeArray.java
```diff
@@ -112,30 +108,22 @@ static NodeArray merge(NodeArray a1, NodeArray a2) {
...
             }
-            // the remaining nodes have a different score, so we can bulk-add them
-            System.arraycopy(a1.nodes, i, merged.nodes, merged.size, a1.size - i);
-            System.arraycopy(a1.scores, i, merged.scores, merged.size, a1.size - i);
             merged.size += a1.size - i;
         }
 
         // If elements remain in a2, add them
         if (j < a2.size()) {
             // avoid duplicates while adding nodes with the same score
```
jkni: stale comment describing old strategy -- can clean up on commit to make this clearer

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeArray.java
```diff
@@ -112,30 +108,22 @@ static NodeArray merge(NodeArray a1, NodeArray a2) {
         // If elements remain in a1, add them
         if (i < a1.size()) {
             // avoid duplicates while adding nodes with the same score
```
jkni: stale comment describing old strategy -- can clean up on commit

jvector-base/src/main/java/io/github/jbellis/jvector/graph/NodeArray.java
```diff
@@ -209,21 +197,20 @@ private int insertInternal(int insertionPoint, int newNode, float newScore) {
         return insertionPoint;
     }
 
-    private boolean duplicateExistsNear(int insertionPoint, int newNode, float newScore) {
-        // Check to the left
-        for (int i = insertionPoint - 1; i >= 0 && scores[i] == newScore; i--) {
-            if (nodes[i] == newNode) {
-                return true;
+    private boolean duplicateExists(int insertionPoint, int newNode) {
```
jkni: I think this could be marginally cleaner -- at i = 0, we check nodes[insertion] point at each conditional, and the loop bounds are unnecessarily pessimistic (i.e., if the insertion point is right in the middle, we'll waste a bunch of loop iterations in both directions, when we could just go to the max radius around the insertion point to hit one end of the array). No idea if this will matter in practice but might be worth measuring.

jkni: e.g., something like
```
    int n = this.size;
    int[] a = this.nodes;

    int left  = insertionPoint - 1;
    int right = insertionPoint;

    // Exact hit fast path
    if (right < n && a[right] == newNode) return true;

    // Expand outward
    while (left >= 0 || right < n) {
        if (left >= 0 && a[left] == newNode) return true;
        if (right < n && a[right] == newNode) return true;
        left--;
        right++;
    }
    return false;
```

marianotepper:
Initial benchmarking shows that index construction got a bit slower (~10%), which is not surprising. Even if conceptually, this PR moves things in the right direction, in practice it does not offer a net benefit. I think that we need a better solution and not just a patch. Will transform the PR into a draft until that superior solution is in place.

PR #526, opened 2025-09-18
Regression enhancements

MarkWolters:
This PR addresses the following issues with the run-bench GitHub workflow for regression testing:

- Links benchmark_config input parameter to regression test arguments: This parameter was intended for use as a filter on the datasets to test with (e.g. provide a regex to run the regression tests on a subset of the default) but was not previously linked to the test execution so that all default datasets were tested regardless of the value of this argument
- Fixes issue where output files were not created when branch name contains an illegal character. Illegal characters will now be replaced with underscores in the filename.
- Fix for issue where report generation was not occurring in the final step because search for csv files was using an incorrect naming pattern.
- Fix for diagnostic logging interfering with expected output of tests


jshook (approved):
LGTM

PR #530, opened 2025-09-19
fix duplicate changelog header

jshook:
This merely removes a duplicitous header which was added when I "updated" changelog. The changelog logic doesn't know how to reconcile deltas to the same labeled section yet.


jshook:
I'll add a new PR which will avoid the breaking path

PR #533, opened 2025-09-30
jvector task tracking internal API

jshook:
**NOTE** This is being shared in its current state for feedback. It is not ready to merge yet.
A few improvements are needed first:
* The integration burden on tracked types needs to be lowered. (this is in progress)
* (better) docs need to be added

The focus of this PR should be on the status API implementation itself. The way the tracked tasks integrate with it will be minimized and simplified in the next phase.

---
This is a status tracking subsystem for jvector. It provides a way to expose the task structure and state of batch or true-up processes that run within jvector during normal operations. It was based on these requirements and design principles:
* Enable tactile user feedback on the status of ongoing tasks through multiple output channels (console, logging, metrics)
* Provide efficient instrumentation for internal JVector use with minimal overhead and comprehensive task lifecycle tracking
* Support complex task structures including batch processing, nested task groups, and DAG-based execution patterns.
* Enable consumer-agnostic monitoring through flexible sink implementations and configurable scope management.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 target/
+local/
 .mvn/wrapper/maven-wrapper.jar
 .java-version
 
--- a/jvector-base/pom.xml
+++ b/jvector-base/pom.xml
@@ -12,6 +12,20 @@
     <artifactId>jvector-base</artifactId>
     <name>Base</name>
 
+    <dependencies>
+        <dependency>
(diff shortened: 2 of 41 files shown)


jshook:
The next form of this will be substantially different enough that this is no longer a good reference point. I'll close this one and land a new PR with a better structure.

PR #534, opened 2025-09-30
New MutableGraphIndex and ImmutableGraphIndex interfaces

marianotepper:
The main intent of this PR is to add an interface in front of the OnHeapGraphIndex. This is because it has become clear recently that we will need new implementations for mutable graphs (that are more conservative in the use of heap memory). This PR paves the way towards that path.

Importantly, MutableGraphIndex is protected inside its package, because modifications should only be done through the GraphIndexBuilder class and never directly.

These changes triggered a lot of minor changes everywhere.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -18,7 +18,7 @@ package io.github.jbellis.jvector.graph;
 
 import io.github.jbellis.jvector.annotations.VisibleForTesting;
 import io.github.jbellis.jvector.disk.RandomAccessReader;
-import io.github.jbellis.jvector.graph.GraphIndex.NodeAtLevel;
+import io.github.jbellis.jvector.graph.ImmutableGraphIndex.NodeAtLevel;
 import io.github.jbellis.jvector.graph.SearchResult.NodeScore;
 import io.github.jbellis.jvector.graph.diversity.VamanaDiversityProvider;
 import io.github.jbellis.jvector.graph.similarity.BuildScoreProvider;
@@ -30,7 +30,6 @@ import io.github.jbellis.jvector.util.ExplicitThreadLocal;
 import io.github.jbellis.jvector.util.PhysicalCoreExecutor;
 import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
 import io.github.jbellis.jvector.vector.types.VectorFloat;
-import org.agrona.collections.IntArrayList;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -49,7 +48,7 @@ import static io.github.jbellis.jvector.util.DocIdSetIterator.NO_MORE_DOCS;
 import static java.lang.Math.*;
 
 /**
- * Builder for Concurrent GraphIndex. See {@link GraphIndex} for a high level overview, and the
+ * Builder for Concurrent GraphIndex. See {@link ImmutableGraphIndex} for a high level overview, and the
  * comments to `addGraphNode` for details on the concurrent building approach.
  * <p>
  * GIB allocates scratch space and copies of the RandomAccessVectorValues for each thread
@@ -71,7 +70,7 @@ public class GraphIndexBuilder implements Closeable {
     private final boolean refineFinalGraph;
 
     @VisibleForTesting
-    final OnHeapGraphIndex graph;
+    final MutableGraphIndex graph;
 
     private final ConcurrentSkipListSet<NodeAtLevel> insertionsInProgress = new ConcurrentSkipListSet<>();
 
@@ -343,7 +342,7 @@ public class GraphIndexBuilder implements Closeable {
     public static GraphIndexBuilder rescore(GraphIndexBuilder other, BuildScoreProvider newProvider) {
         var newBuilder = new GraphIndexBuilder(newProvider,
                 other.dimension,
-                other.graph.maxDegrees,
+                other.graph.maxDegrees(),
                 other.beamWidth,
                 other.neighborOverflow,
                 other.alpha,
@@ -352,17 +351,13 @@ public class GraphIndexBuilder implements Closeable {
                 other.simdExecutor,
                 other.parallelExecutor);
 
+        var otherView = other.graph.getView();
+
         // Copy each node and its neighbors from the old graph to the new one
         other.parallelExecutor.submit(() -> {
             IntStream.range(0, other.graph.getIdUpperBound()).parallel().forEach(i -> {
                 // Find the highest layer this node exists in
-                int maxLayer = -1;
-                for (int lvl = 0; lvl < other.graph.layers.size(); lvl++) {
-                    if (other.graph.getNeighbors(lvl, i) == null) {
-                        break;
-                    }
-                    maxLayer = lvl;
-                }
+                int maxLayer = other.graph.getMaxLevelForNode(i);
                 if (maxLayer < 0) {
                     return;
                 }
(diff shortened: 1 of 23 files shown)


marianotepper:
@sam-herman @michaeljmarshall Please make sure that the new design does not preclude any usages of JVector

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -442,8 +450,8 @@ public void cleanup() {
         // clean up overflowed neighbor lists
         parallelExecutor.submit(() -> {
             IntStream.range(0, graph.getIdUpperBound()).parallel().forEach(id -> {
-                for (int layer = 0; layer < graph.layers.size(); layer++) {
-                    graph.layers.get(layer).enforceDegree(id);
+                for (int layer = 0; layer <= graph.getMaxLevel(); layer++) {
```
tlwillke, line 453: I notice we're using the terms layer and level.  I think we mean the same thing and should probably settle on a single term.  Not a new issue BTW.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
```diff
@@ -0,0 +1,170 @@
...
+    int removeNode(int node);
+
+    /**
+     * this does call get() internally to filter level 0, so if you're going to use it in a pipeline
+     * that also calls get(), consider using your own raw IntStream.range instead
+     */
+    IntStream nodeStream(int level);
+
+    /**
+     * @return the maximum (coarser) level that contains a vector in the graph.
```
michaeljmarshall: I see that this is `-1` when the node isn't in the graph and that the internal caller uses that design detail. Should we add it to the javadoc?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
```diff
@@ -0,0 +1,170 @@
...
+     * @param node the node id to remove
+     * @return the number of layers from which it was removed
+     */
+    int removeNode(int node);
+
+    /**
+     * this does call get() internally to filter level 0, so if you're going to use it in a pipeline
+     * that also calls get(), consider using your own raw IntStream.range instead
+     */
+    IntStream nodeStream(int level);
```
michaeljmarshall, line 141: Is this javadoc correct?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
```diff
@@ -76,6 +78,8 @@ default int size() {
      */
     int maxDegree();
 
+    List<Integer> maxDegrees();
```
tlwillke, line 81: Maybe we should do away with layer-wise maxDegrees in the future since we don't use the flexibility.

marianotepper: Yes, maybe. Since this PR is not changing anything functionally but just refactoring existing stuff, I'd rather leave that work for the future.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -260,41 +287,21 @@ public void close() {
...
-    public ConcurrentGraphIndexView getView() {
-        return new ConcurrentGraphIndexView();
-    }
-
-    /**
-     * A View that assumes no concurrent modifications are made
-     */
-    public GraphIndex.View getFrozenView() {
+    public View getView() {
         return new FrozenView();
```
michaeljmarshall: Is there any reason to prefer a `ConcurrentGraphIndexView` as opposed to a `FrozenView` here? We have code that calls the `getView()` method, and this change switches which object we get.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Remove edges to deleted nodes and add the new connections, maintaining diversity
+     */
+    void replaceDeletedNeighbors(int level, int node, BitSet toDelete, NodeArray candidates);
+
+    /**
+     * A View compatible with concurrent modifications
+     */
+    View getConcurrentView();
```
michaeljmarshall: Nit: this seems like it could be clarified that the concurrent modifications are in reference to the graph, not the view, since the `View` is not necessarily thread safe. I might be overthinking it, but just mentioning it since this is an interface review.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Enforce the degree of the given node in all layers.
+     */
+    void enforceDegree(int node);
+
+    /**
+     * Returns an iterator over the neighbors for the given node at the specified level, which can be empty if the node does not belong to that layer.
+     */
+    NodesIterator getNeighborsIterator(NodeAtLevel nodeLevel);
```
michaeljmarshall, line 123: This method isn't used, perhaps that is okay though since it could be a convenience method.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
marianotepper: I partially corrected this. This will need a good pass over everything to be fully corrected.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
marianotepper: Fixed

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
marianotepper: I added it for symmetry with the other methods that have both versions. Doesn't hurt.

michaeljmarshall: I agree, it doesn't hurt.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
marianotepper: This is a great question. I modified the logic so that `OnHeapGraphIndex.getView` returns a `ConcurrentGraphIndexView` while it is being mutated by `GraphIndexBuilder` and a `FrozenView` once mutations are done (after `GraphIndexBuilder.cleanup`). I think this captures the best of both worlds.

michaeljmarshall: Sounds great

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -72,11 +69,13 @@ public class OnHeapGraphIndex implements GraphIndex {
     private final AtomicInteger maxNodeId = new AtomicInteger(-1);
 
     // Maximum number of neighbors (edges) per node per layer
-    final IntArrayList maxDegrees;
+    final List<Integer> maxDegrees;
     // The ratio by which we can overflow the neighborhood of a node during construction. Since it is a multiplicative
     // ratio, i.e., the maximum allowable degree if maxDegree * overflowRatio, it should be higher than 1.
     private final double overflowRatio;
 
+    private boolean allMutationsCompleted = false;
```
michaeljmarshall: Did you consider making this `volatile`? In its current usage, it doesn't need it, strictly speaking, because the race is benign and simply results in getting a safer view of the graph.

marianotepper: Done

tlwillke (approved):
LGTM

michaeljmarshall (approved):

PR #475, opened 2025-06-11
Sequential disk writer

sam-herman:
### Description
Main motivation for this change is to introduce a disk writer that keeps immutability and is pure sequential.
This is essential for integration with frameworks such as Lucene and OpenSearch.

### Changes
Additional changes in this PR
- Introduce `OnDiskSequentialGraphIndexWriter` and abstract the `GraphWriter` interface
- Fix serialization issue where previously written nodeIds were include when calling `getNodes`
- Reuse cache for in memory loaded layers to remove unnecessary disk seeks and bugs due to branching of logic related to offset calculation.

### Tests
Introduce tests for `TestOnDiskSequentialGraphIndexWriter` and add proper configuration for log4j debug logs in tests.

PR #535, opened 2025-10-02
Prefer footer constants in AbstractGraphIndexWriter

michaeljmarshall:
While working on the upgrade to jvector file format 5, I ran into some file validation errors regarding footer magic. I discovered that with https://github.com/datastax/jvector/pull/475, we have the same `FOOTER_MAGIC` defined in two places with the same hard coded value `0x4a564244`. Notably, we wrote to disk using one variable and then validated using the other, which seems error prone.

I propose that we remove the unnecessary duplication. This is technically a breaking change because we are removing public variables, but a quick search of OSS projects indicates no one is using these variables yet https://grep.app/search?q=OnDiskSequentialGraphIndexWriter.

The alternative is to define the `OnDiskSequentialGraphIndexWriter` constants with references to the `AbstractGraphIndexWriter` constants.


MarkWolters (approved):
LGTM

PR #536, opened 2025-10-07
Reconstruct heap graph from disk graph

sam-herman:
#### Description
Multiple database projects such as C* and OpenSearch/Solr/Lucene use LSM mechanism with frequent merges.
This in turn creates a high overhead that forces us to reconstruct the entire graph from scratch upon every merge.
A more economic approach would be to pick a leading graph that was previously persisted to disk and incrementally add small graph nodes to it.

This PR makes some of the required changes to support that behavior in upstream systems such as C* and OpenSearch.

#### Changes

1. **Serializable Neighbor Distance Cache** - Add a serializable cache that can store the node distances within the `OnDiskGraphIndex` for a faster re-creation of the `OnHeapGraphIndex` when read back from disk. The cache is separate and is optional, therefore we can choose whether to apply it or not, without any breaking changes to the current `OnDiskGraphIndex`. This can later be augmented to a format if we choose to.
2. **Help Methods And Constructors** - Add helper method and constructors to facilitate easier usage by other projects with graph merge use cases.

#### Testing

Added tests for graph overlap and recall for reconstructed graphs.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -327,7 +327,7 @@ public class GraphIndexBuilder implements Closeable {
         this.simdExecutor = simdExecutor;
         this.parallelExecutor = parallelExecutor;
 
-        this.graph = new OnHeapGraphIndex(maxDegrees, neighborOverflow, new VamanaDiversityProvider(scoreProvider, alpha));
+        this.graph = new OnHeapGraphIndex(dimension, maxDegrees, neighborOverflow, new VamanaDiversityProvider(scoreProvider, alpha));
         this.searchers = ExplicitThreadLocal.withInitial(() -> {
             var gs = new GraphSearcher(graph);
             gs.usePruning(false);
@@ -346,7 +346,7 @@ public class GraphIndexBuilder implements Closeable {
      * copy it into {@link OnHeapGraphIndex} and then start mutating it with minimal overhead of recreating the mutable {@link OnHeapGraphIndex} used in the new GraphIndexBuilder object
      *
      * @param buildScoreProvider the provider responsible for calculating build scores.
-     * @param onDiskGraphIndex the on-disk representation of the graph index to be processed and converted.
+     * @param immutableGraphIndex the on-disk representation of the graph index to be processed and converted.
      * @param perLevelNeighborsScoreCache the cache containing pre-computed neighbor scores,
      *                                    organized by levels and nodes.
      * @param beamWidth the width of the beam used during the graph building process.
@@ -359,10 +359,10 @@ public class GraphIndexBuilder implements Closeable {
      *
      * @throws IOException if an I/O error occurs during the graph loading or conversion process.
      */
-    public GraphIndexBuilder(BuildScoreProvider buildScoreProvider, OnDiskGraphIndex onDiskGraphIndex, NeighborsScoreCache perLevelNeighborsScoreCache, int beamWidth, float neighborOverflow, float alpha, boolean addHierarchy, boolean refineFinalGraph, ForkJoinPool simdExecutor, ForkJoinPool parallelExecutor) throws IOException {
+    public GraphIndexBuilder(BuildScoreProvider buildScoreProvider, ImmutableGraphIndex immutableGraphIndex, NeighborsScoreCache perLevelNeighborsScoreCache, int beamWidth, float neighborOverflow, float alpha, boolean addHierarchy, boolean refineFinalGraph, ForkJoinPool simdExecutor, ForkJoinPool parallelExecutor) throws IOException {
         this.scoreProvider = buildScoreProvider;
         this.neighborOverflow = neighborOverflow;
-        this.dimension = onDiskGraphIndex.getDimension();
+        this.dimension = immutableGraphIndex.getDimension();
         this.alpha = alpha;
         this.addHierarchy = addHierarchy;
         this.refineFinalGraph = refineFinalGraph;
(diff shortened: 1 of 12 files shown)


marianotepper:
@sam-herman thanks for the contribution.

I created an alternative [branch](https://github.com/datastax/jvector/tree/reconstruct-heap-graph-from-disk-graph2) where:
- Interface boundaries are respected (no unprotected downcasts).
- The NeighborsScoreCache is removed in favor of recomputing the scores live when creating the OnHeapGraphIndex.

The last change is mainly due to a major refactoring that needs to happen in OnHeapGraphIndex to not store the scores anymore. This is creating issues, see #522 for example. So in essence, once the scores are removed, we will have no more need for a cache or for recomputing the scores themselves. I'll be working on this change next week.

sam-herman:
> @sam-herman thanks for the contribution.
> 
> I created an alternative [branch](https://github.com/datastax/jvector/tree/reconstruct-heap-graph-from-disk-graph2) where:
> 
> * Interface boundaries are respected (no unprotected downcasts).
> * The NeighborsScoreCache is removed in favor of recomputing the scores live when creating the OnHeapGraphIndex. The NeighborsScoreCache was essentially forcing to write the graph twice on disk and read it twice (i.e., instantiate two copies in memory, the cache and the OnHeapGraph).
> 
> The last change is mainly due to a major refactoring that needs to happen in OnHeapGraphIndex to not store the scores anymore. This is creating issues, see #522 for example. So in essence, once the scores are removed, we will have no more need for a cache or for recomputing the scores themselves. I'll be working on this change next week.

@marianotepper the issue with recomputing scores of the entire graph is that it forces a brute force pass. 
For frequent merges between small graphs and large graphs, this is a situation we would prefer to avoid to keep the performance as we've previously seen in https://github.com/opensearch-project/opensearch-jvector/pull/167

One option that we mentioned was to not include scores at all in the OnHeapGraphIndex and then only re calculating the relative neighbor scores upon traversal. Would it make sense to keep the score cache for now and later remove it when we are ready to implement this approach?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -490,46 +497,101 @@ public String toString() {
...
-                        out.writeInt(iterator.size());
-                        for (int n = 0; n < iterator.size(); n++) {
-                            out.writeInt(iterator.nextInt());
-                        }
-                        assert !iterator.hasNext();
-                    } catch (IOException e) {
-                        throw new UncheckedIOException(e);
-                    }
-                });
+    @Experimental
```
jshook, line 500: We should probably also mark these as @Deprecated if they are expected to be removed and should not be used in general. (Comments apply to all @Experimental annotations)

marianotepper: Done

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -860,6 +920,7 @@ private void loadV4(RandomAccessReader in) throws IOException {
                 var ca = new NodeArray(nNeighbors);
                 for (int j = 0; j < nNeighbors; j++) {
                     int neighbor = in.readInt();
+                    float score = in.readFloat();
```
MarkWolters, line 923: "score" seems to be read in but then never used for anything. Is this intentional?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
```diff
@@ -166,5 +166,11 @@ interface MutableGraphIndex extends ImmutableGraphIndex {
      * Signals that all mutations have been completed and the graph will not be mutated any further.
      * Should be called by the builder after all mutations are completed (during cleanup).
      */
-    void allMutationsCompleted();
+    void setAllMutationsCompleted();
+
+    /**
```
MarkWolters, line 171: The documentation for these 2 methods is identical. Which one should be called by the builder?

MarkWolters: actually in looking at the rest of the code I believe the doc above the method returning a boolean should be updated.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
marianotepper: I could have used `seek` there. This function is deprecated but still used in some tests, so I just made it work. GraphIndexBuilder.load should go away altogether anyway.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/MutableGraphIndex.java
marianotepper: Done

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
MarkWolters: understood. I'll mark resolved.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -227,10 +225,6 @@ public Set<FeatureId> getFeatureSet() {
         return features.keySet();
     }
 
-    public int getDimension() {
```
michaeljmarshall, line 230: We are using this in C*, why did this get removed?

michaeljmarshall: I submitted a PR proposing we re-add it. https://github.com/datastax/jvector/pull/550

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
```diff
@@ -83,8 +85,18 @@ public interface BuildScoreProvider {
 
     /**
      * Returns a BSP that performs exact score comparisons using the given RandomAccessVectorValues and VectorSimilarityFunction.
+     *
+     * Helper method for the special case that mapping between graph node IDs and ravv ordinals is the identity function.
      */
     static BuildScoreProvider randomAccessScoreProvider(RandomAccessVectorValues ravv, VectorSimilarityFunction similarityFunction) {
+        return randomAccessScoreProvider(ravv, IntStream.range(0, ravv.size()).toArray(), similarityFunction);
```
michaeljmarshall, line 92: This method operates on a mutable ravv, so it's not valid to have a fixed array as a mapper. Further, this has linear memory complexity to represent an identify function. Instead, we should use the `OrdinalMapper` class to handle this use case, assuming I understand it correctly.

PR #537, opened 2025-10-08
Introduce jvector-apis module and status tracker

jshook:
[Updates 10/15]:

With some first-user input from Mark and some other cleanups and improved docs, the core code is tighter, safer, and cleaner. There were several files added in test, specifically to make it easy to understand the usage patterns for implementing tracked tasks.

I believe that we should have living examples where we can, and that means they should be tested. If necessary, we can put a group tag on the "example" tests to bypass them conveniently except for release-level tests. For now, they are enabled for review here.

[Previously ...]

The status tracking module may have use in both prod and test code, but we wanted to isolate and manage it better as a separate module. For modules like this, I propose we use a jvector-apis module. This is where you would put new modular functionality, like the status tracking API, which may be used by multiple modules. There is a good primer on this in the README in the module root.

For the status tracking API, this is a new facility to allow us to collect and share the status of internal jobs that are being run by jvector. The first use of it will be to ease testing and baseline work around performance and accuracy for different vector spaces and indexing configurations. However, it does have some hooks which may be lightweight enough to instrument prod code with, and this is a separate concern not addressed specifically in this PR. For now, this merely introduces the status tracking API, which will be pulled into the dataset streaming work once when ready.

There is a substantial amount of testing included in this PR. The demo scaffolding is there as a test layer for improving the API and making sure it is ergonomic and non-invasive enough to be added to extant code. If necessary, we can gate the unit tests of this module with an optional test group, but I'd like to see how it works as is first.

There are a couple of unrelated cleanups in this PR as well, from previous commits by other committers, around mvn exec configs, logger and test dependency inclusions which were not intended. I've personally contacted these committers and verified their intent before removing the extra deps and configurations.

### DIFF
--- a/internal-apis/src/main/java/io/github/jbellis/jvector/status/README.md
+++ b/internal-apis/src/main/java/io/github/jbellis/jvector/status/README.md
@@ -15,7 +15,7 @@
 -->
 # Status API
 
-**📚 For complete user guide with examples:** See [User Guide](../../../test/java/io/github/jbellis/jvector/status/userguide/README.md)
+**📚 For complete user guide with examples:** See [User Guide](../../../test/java/io/github/jbellis/jvector/status/userguide/README.md) section in the tests package. It includes incremental examples and working code.
 
 This internal API was added to solve a few problems around a common theme:
 
--- a/internal-apis/src/test/java/io/github/jbellis/jvector/status/userguide/UserGuideExamplesTest.java
+++ b/internal-apis/src/test/java/io/github/jbellis/jvector/status/userguide/UserGuideExamplesTest.java
@@ -1,0 +1,240 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.jbellis.jvector.status.userguide;
+
+import io.github.jbellis.jvector.status.StatusContext;
+import io.github.jbellis.jvector.status.StatusScope;
+import io.github.jbellis.jvector.status.StatusTracker;
+import io.github.jbellis.jvector.status.eventing.RunState;
+import io.github.jbellis.jvector.status.sinks.ConsoleLoggerSink;
+import io.github.jbellis.jvector.status.sinks.MetricsStatusSink;
+import io.github.jbellis.jvector.status.userguide.fauxtasks.DataLoader;
+import io.github.jbellis.jvector.status.userguide.fauxtasks.DataProcessor;
+import io.github.jbellis.jvector.status.userguide.fauxtasks.DataValidator;
+import io.github.jbellis.jvector.status.userguide.fauxtasks.ParallelDataLoader;
+import org.junit.jupiter.api.Test;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Verification tests for all user guide examples.
+ * Ensures that each level example works as described and demonstrates the concepts correctly.
+ */
+public class UserGuideExamplesTest {
+
+    @Test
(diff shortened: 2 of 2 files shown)


jvector-apis/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>2.0.9</version>
```
MarkWolters: the top level pom brings in version 2.0.16

jvector-apis/src/main/java/io/github/jbellis/jvector/status/StatusContext.java
```diff
@@ -0,0 +1,576 @@
...
+    private void scopeStarted(TrackerScope scope) {
+        notifySinks(sink -> {
+            // Use reflection to avoid circular dependency
+            try {
+                var method = sink.getClass().getMethod("scopeStarted", TrackerScope.class);
+                method.invoke(sink, scope);
+            } catch (NoSuchMethodException e) {
+                // Sink doesn't support scopes - ignore
+            } catch (Exception e) {
+                // Log but don't fail
```
MarkWolters: This says "Log but don't fail" but the behavior seems to be the opposite, i.e. if an exception is thrown it fails but does so silently without logging anything, unless I am misinterpreting the behavior of this method?

jvector-apis/src/main/java/io/github/jbellis/jvector/status/StatusContext.java
```diff
@@ -0,0 +1,576 @@
...
+     *
+     * @param tracked the object to track
+     * @param statusFunction function to extract status from the tracked object
+     * @param pollInterval the interval between status observations
+     * @param <T> the type of object being tracked
+     * @return a new tracker registered with this context's monitor
+     * @deprecated StatusTrackers must belong to a TrackerScope. Use {@link TrackerScope#trackTask} instead.
+     *             Create a scope first with {@link #createScope(String)}, then track tasks within it.
+     */
+    @Deprecated
```
MarkWolters: this module is new and does not need to maintain any backwards compatibility, why include deprecated methods?

internal-apis/src/main/java/io/github/jbellis/jvector/status/StatusTracker.java
```diff
@@ -0,0 +1,326 @@
...
+     *
+     * @param status the observed status containing runstate information
+     */
+    private void updateTiming(StatusUpdate<T> status) {
+        if (status == null || status.runstate == null) {
+            return;
+        }
+
+        synchronized (timingLock) {
+            switch (status.runstate) {
```
MarkWolters, line 378: slight risk of NPE here as the initial check is outside of the synchronized block

jvector-apis/src/main/java/io/github/jbellis/jvector/status/StatusContext.java
jshook: This is being improved in the next wave of changes.

jvector-apis/src/main/java/io/github/jbellis/jvector/status/StatusContext.java
jshook: removed in the next wave of changes

jvector-apis/pom.xml
jshook: aligned to root pom dep management in the next wave of changes

internal-apis/src/main/java/io/github/jbellis/jvector/status/StatusTracker.java
jshook: On second thought, let me dig into this one a bit.

jshook: Static analysis agrees when zoomed in, but zooming out slightly shows that this is safe in the current form. javadocs are added to explain it a bit better.

jvector-apis/src/main/java/io/github/jbellis/jvector/status/StatusContext.java
jshook: error handling reviewed and improved in current version

pom.xml
```diff
@@ -68,15 +69,6 @@
                     <skip>true</skip>
                 </configuration>
             </plugin>
-            <plugin>
-                <groupId>org.codehaus.mojo</groupId>
```
MarkWolters, line 72: why remove this plugin?

MarkWolters (approved):
I think this looks good. The documentation and the example code really helps.

pom.xml
jshook: This was added before by a dev that I follow up with before removing it. This wasn't removing the plugin entirely, but it was removing a default that was affecting all invocations of the exec plugin which didn't have their own local overrides. Specifically it was skipping execution, and that seemed to have an undesired side-effect on sub-modules as well.

PR #542, opened 2025-10-15
Parallelize graph writes

MarkWolters:
This PR adds the option to specify that graph indexes should be written in parallel rather than sequentially. By default the existing sequential write behavior is maintained, **_parallel writes will only be used when the withParallelWrites(true) option is set through the OnDiskGraphIndexWriter.Builder class_**. Testing results below show the speedup achieved in the write phase across a number of cores. These gains appear to scale linearly with respect to dataset size (ie writing a dataset of 10 million records will take about 10x as long as a dataset of 1 million records but the speedup in parallel v sequential is roughly equal).

ETA: Testing also showed that the performance gains using a "prod-like" i3.4xlarge w/16 vCPUs and 8 disks striped RAID0 were roughly equivalent to the performance gains using a 64 vCPU m5.16xlarge with standard SSD

sequential

parallel

speedup

### DIFF
--- a/benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/ParallelWriteBenchmark.java
+++ b/benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/ParallelWriteBenchmark.java
@@ -51,7 +51,7 @@ import static io.github.jbellis.jvector.quantization.KMeansPlusPlusClusterer.UNW
 
 /**
  * JMH benchmark that mirrors the ParallelWriteExample: it builds a graph from vectors, then
- * writes the graph to disk sequentially and in parallel using NVQ + FUSED_ADC features,
+ * writes the graph to disk sequentially and in parallel using NVQ + FUSED_PQ features,
  * and verifies that the outputs are identical.
  */
 @BenchmarkMode(Mode.AverageTime)
@@ -70,12 +70,15 @@ public class ParallelWriteBenchmark {
     @Param({"1024"})
     int dimension;
 
+    @Param({"true", "false"})
+    boolean addHierarchy;
+
     // Graph build parameters
     final int M = 32;
     final int efConstruction = 100;
     final float neighborOverflow = 1.2f;
     final float alpha = 1.2f;
-    final boolean addHierarchy = false;
+    //final boolean addHierarchy = false;
     final boolean refineFinalGraph = true;
 
     // Dataset and index state
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
@@ -16,6 +16,7 @@
 
 package io.github.jbellis.jvector.graph.disk;
 
+import io.github.jbellis.jvector.annotations.Experimental;
 import io.github.jbellis.jvector.disk.BufferedRandomAccessWriter;
 import io.github.jbellis.jvector.disk.RandomAccessWriter;
 import io.github.jbellis.jvector.graph.ImmutableGraphIndex;
@@ -241,6 +242,7 @@ public class OnDiskGraphIndexWriter extends AbstractGraphIndexWriter<RandomAcces
      * @param featureStateSuppliers suppliers for feature state data
      * @throws IOException if an I/O error occurs
      */
+    @Experimental
     private void writeL0RecordsParallel(Map<FeatureId, IntFunction<Feature.State>> featureStateSuppliers) throws IOException {
         if (filePath == null) {
             throw new IllegalStateException("Parallel writes require a file path. Use Builder(ImmutableGraphIndex, Path) constructor.");
@@ -362,7 +364,7 @@ public class OnDiskGraphIndexWriter extends AbstractGraphIndexWriter<RandomAcces
      * to ensure that all parallel write operations have fully completed before writing the footer (e.g., checksum).
      * The footer must only be written after all data has been flushed and no further writes are in progress,
      * to avoid data corruption or incomplete checksums. This class does not currently coordinate or synchronize
-     * footer writing with parallel operations. Parallel writes are experimental and should be used with caution.
+     * footer writing with parallel operations.
      */
     public synchronized long checksum() throws IOException {
         long endOffset = out.position();
@@ -373,8 +375,6 @@ public class OnDiskGraphIndexWriter extends AbstractGraphIndexWriter<RandomAcces
      * Enables parallel writes for L0 records. This can significantly improve throughput
      * for large graphs by parallelizing record building across multiple cores.
      * <p>
-     * Note: This is currently experimental. The sequential path is the default.
-     *
      * @param enabled whether to enable parallel writes
      */
     public void setParallelWrites(boolean enabled) {
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelWriteExample.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelWriteExample.java
@@ -322,7 +322,7 @@ public class ParallelWriteExample {
         int efConstruction = 100;
         float neighborOverflow = 1.2f;
         float alpha = 1.2f;
-        boolean addHierarchy = false;
+        boolean addHierarchy = true;
         boolean refineFinalGraph = true;
 
         System.out.printf("Building graph with PQ-compressed vectors (M=%d, efConstruction=%d)...%n", M, efConstruction);
--- a/jvector-examples/yaml-configs/autoDefault.yml
+++ b/jvector-examples/yaml-configs/autoDefault.yml
@@ -1,4 +1,4 @@
-version: 5
+version: 6
 
 dataset: cohere-english-v3-100k
 


jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -0,0 +1,168 @@
...
+        this.featureStateSuppliers = featureStateSuppliers;
+        this.recordSize = recordSize;
+        this.fileOffset = fileOffset;
+        this.buffer = buffer;
+    }
+
+    @Override
+    public Result call() throws Exception {
+        buffer.clear();
+        var writer = new ByteBufferIndexWriter(buffer);
```
michaeljmarshall: The ownership model and lifecycle for the `buffer` here is a bit ambiguous to me, especially because the `buffer` is passed into this class as a parameter. I wonder if we can make the `ByteBufferIndexWriter` hide some of the implementation details so that we do not have a `buffer.clear()` and the buffer management logic at the end of this method. Instead, we could make the `ByteBufferIndexWriter` manage all of that, by adding a clone and a reset method, perhaps?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,321 @@
...
+                        recordSize,
+                        fileOffset,
+                        buffer
+                );
+
+                return task.call();
+            });
+
+            futures.add(future);
+        }
```
michaeljmarshall: IIUC, this doesn't have any back pressure mechanism, and we need that to prevent excessive memory utilization. I think we might want to consider an implementation that uses semaphores to manage concurrency.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,321 @@
...
+                        return t;
+                    });
+
+            try (var afc = AsynchronousFileChannel.open(filePath, opts, fileWritePool)) {
+                for (Future<NodeRecordTask.Result> future : futures) {
+                    NodeRecordTask.Result result = future.get();
+
+                    // result.data is already a copy made in NodeRecordTask to avoid
+                    // race conditions with thread-local buffer reuse
+                    afc.write(result.data, result.fileOffset).get();
```
michaeljmarshall: The `.get()` here negates the usages of the async file channel.

michaeljmarshall: We could use a sempahore to limit the number of write tasks in flight.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: updated ByteBufferIndexWriter with CloneBuffer (to avoid confusion with Object.clone()) and updated reset to include clearing the buffer. Moved that logic out of NodeRecordTask to clarify ownership in latest commit.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: In the latest commit I've added backpressure mechanisms to both this loop and the file channel write loop. I'm not entirely clear on what solution you had in mind vis-a-vis semaphores, but I am open to re-implementing in another fashion if you feel that the code could be improved.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: this section has been rewritten to remove the get() call at this point in the code and to split the write process into sub-tasks. Again, I'm happy to go with a different implementation if the code can be improved but I'm not sure exactly what you had in mind.

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
```diff
@@ -0,0 +1,252 @@
...
+     */
+    public ByteBufferIndexWriter(ByteBuffer buffer) {
+        this(buffer, true);
+    }
+
+    /**
+     * Creates a writer with a new heap ByteBuffer of the given capacity.
+     * The buffer uses BIG_ENDIAN byte order.
+     */
+    public static ByteBufferIndexWriter allocate(int capacity) {
```
sam-herman: nit: name doesn't sound very standard for creator method.
perhaps `create(int bufferCapcity, boolean offHeap)`?

jvector-examples/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelWriteExample.java
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Main method to run a benchmark test of sequential vs parallel writes.
+     *
+     * Usage: java ParallelWriteExample [dataset-name]
+     *
+     * Example: java ParallelWriteExample cohere-english-v3-100k
+     *
+     * If no dataset is provided, uses "cohere-english-v3-100k" by default.
+     */
+    public static void main(String[] args) throws IOException {
```
sam-herman, line 303: Should we make this a JMH benchmark?
https://github.com/datastax/jvector/tree/main/benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -243,20 +355,47 @@ public synchronized void writeHeader(ImmutableGraphIndex.View view) throws IOExc
         out.flush();
     }
 
+    //TODO: How do the parallel writes work with the footer?
```
sam-herman: Maybe just write footer when all other writes completed? I think I've seen it in the code as well line 221
```Java
        if (version >= 5) {
            writeFooter(view, out.position());
        }
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -155,6 +204,86 @@ public synchronized void write(Map<FeatureId, IntFunction<Feature.State>> featur
...
+            writeL0RecordsSequential(view, featureStateSuppliers);
+        }
+
+        // We will use the abstract method because no random access is needed
+        writeSparseLevels(view);
+
+        // We will use the abstract method because no random access is needed
+        writeSeparatedFeatures(featureStateSuppliers);
+
+        // Write the header again with updated offsets
```
sam-herman: Did you mean to move this comment to line 225, and here say "write footer again"?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,371 @@
...
+         */
+        public static Config defaultConfig() {
+            return new Config(0, false);
+        }
+    }
+
+    /**
+     * Creates a parallel writer.
+     *
+     * @param writer the underlying writer (must be BufferedRandomAccessWriter)
```
sam-herman: `the underlying writer (must be BufferedRandomAccessWriter)` does that mean we need to limit the type to `BufferedRandomAccessWriter`?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,371 @@
...
+     * @param inlineFeatures the inline features to write
+     * @param config parallelization configuration
+     * @param filePath file path for async writes (required, cannot be null)
+     */
+    public ParallelGraphWriter(RandomAccessWriter writer,
+                               ImmutableGraphIndex graph,
+                               List<Feature> inlineFeatures,
+                               Config config,
+                               Path filePath) {
+        if (filePath == null) {
```
sam-herman: nit: can be `Objects.requireNonNull()`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,371 @@
...
+            throw new IllegalArgumentException("filePath is required for parallel writes");
+        }
+        if (!(writer instanceof BufferedRandomAccessWriter)) {
```
sam-herman: why not just limit the type argument taken by the constructor? seems more reasonable than take wrong type and throw exception?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,371 @@
...
+            throw new IllegalArgumentException("writer must be BufferedRandomAccessWriter for parallel writes");
+        }
+
+        this.writer = writer;
+        this.graph = graph;
+        this.filePath = filePath;
+        this.executor = Executors.newFixedThreadPool(config.workerThreads,
+            r -> {
+                Thread t = new Thread(r);
+                t.setName("ParallelGraphWriter-Worker");
```
sam-herman: Would there be any numbering appended to the name? e.g. `ParallelGraphWriter-Worker-1`
Context: will this create all threads with the exact same name or will numbering be appended?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.concurrent.Callable;
+import java.util.function.IntFunction;
+
+/**
+ * A task that builds a single node's L0 record in memory.
+ * <p>
+ * This task is designed to be executed in a thread pool, with each worker thread
+ * owning its own ImmutableGraphIndex.View for thread-safe neighbor iteration.
+ */
+class NodeRecordTask implements Callable<NodeRecordTask.Result> {
```
sam-herman: What would be the memory impact of creating this object for each write operation?

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
MarkWolters: This is meant to align with the ByteBuffer API , i.e. `public static ByteBuffer allocate(int capacity)` and `public static ByteBuffer allocateDirect(int capacity)`

jvector-examples/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelWriteExample.java
MarkWolters: Done in the latest commit, although I want to keep the original test as well, as it offers more flexibility in dataset selection and manual testing.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
MarkWolters: The (original) comments are a little confusing here, because there is a footer that is written by the Writer, but there is also a checksum, which is referred to as the `footer` here, which is not written by the Writer and depends on the client to write it. I've updated the comment here which hopefully clarifies this.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
MarkWolters: This is fro the original code, just shuffled around, so not added in this PR. I'll update the location of the comment on the next commit.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: This was from a different design, I have removed the requirement in the latest commit.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: This was from a different design, I have removed the requirement in the latest commit.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: Done in latest commit

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: Done in latest commit.

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
sam-herman: Hmm.. it is not a standard name however for an index writer constructor... or a FileWriter

jvector-examples/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelWriteExample.java
sam-herman: > although I want to keep the original test as well, as it offers more flexibility in dataset selection and manual testing.


sounds good

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+
+    @Override
+    public void writeUTF(String s) throws IOException {
+        // Use standard DataOutputStream UTF encoding
+        byte[] bytes = s.getBytes("UTF-8");
+        int utflen = bytes.length;
+        
+        if (utflen > 65535) {
```
sam-herman, line 264: Why this number? 65535

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/GraphIndexWriterTypes.java
```diff
@@ -0,0 +1,42 @@
...
+
+package io.github.jbellis.jvector.graph.disk;
+
+/**
+ * Enum defining the available types of graph index writers.
+ * <p>
+ * Different writer types offer different tradeoffs between performance,
+ * compatibility, and features.
+ */
+public enum GraphIndexWriterTypes {
```
sam-herman, line 25: Love the enum types and the explanation here!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -0,0 +1,166 @@
...
+        this.featureStateSuppliers = featureStateSuppliers;
+        this.recordSize = recordSize;
+        this.fileOffset = fileOffset;
+        this.buffer = buffer;
+    }
+
+    @Override
+    public Result call() throws Exception {
+        // Writer automatically clears buffer on construction
+        var writer = new ByteBufferIndexWriter(buffer);
```
sam-herman, line 92: Why not make the `ByteBufferIndexWriter` thread local and avoid recreating it on every call?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,362 @@
...
+     * <p>
+     * The calculation considers:
+     * - Current heap usage and maximum heap size
+     * - Estimated memory per record (data + Future/Result overhead)
+     * - Conservative allocation (20% of available memory)
+     * - Reasonable bounds (min 100, max 1000000 records)
+     *
+     * @return the number of records to buffer before writing
+     */
+    private int calculateOptimalBufferSize() {
```
sam-herman: Should we take graph properties to this calculation to consider alignment with device blocks?
Also if we know the supported parallelism of the hardware, can leverage that as well to reduce the buffer size in the calculation to achieve the minimal queue depth?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,362 @@
...
+        int maxOrdinal = ordinalMapper.maxOrdinal();
+        List<Future<NodeRecordTask.Result>> futures = new ArrayList<>(bufferSize);
+
+        // Build and write records in chunks to manage memory usage
+        for (int newOrdinal = 0; newOrdinal <= maxOrdinal; newOrdinal++) {
+            final int ordinal = newOrdinal;
+            final long fileOffset = baseOffset + (long) ordinal * recordSize;
+
+            // Submit task to build this record
+            Future<NodeRecordTask.Result> future = executor.submit(() -> {
```
sam-herman: This will create a new future object for each ordinal? Should we batch those to prevent GC and memory allocation overhead?

sam-herman (commented):
Added more comments, primarily around two aspects:
1. Calculation of buffer - Seems like we need to consider alignment and hardware parallelism to reduce buffer sizes.
2. Batching and GC and malloc overhead - It seems as if we are creating a task per ordinal, which seems like something that could be somewhat expensive on memory allocation and GC overhead.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: It is batched. The buffer is checked and, if full, written and cleared on lines 190 - 193.
`if (futures.size() >= bufferSize) {
    writeRecordsAsync(futures);
    futures.clear();
}`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: That depends on dataset size and vector dimensionality. But the number of NodeRecordTasks in existence at any time is bounded by a buffer that gets sized based on available memory as a back pressure mechanism.

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
MarkWolters: UTF format stores the string length as a 2-byte (16-bit) unsigned integer prefix, which has a maximum value of 65535

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
MarkWolters: I'll add a public `public static ByteBufferIndexWriter create(int capacity, boolean offHeap)` method and make the allocate/allocateDirect methods private and internal in the next commit.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: I don't think there's a significant difference. The constructor is trivially cheap, it just stores the buffer reference and initial position, and the real cost (the BytBuffer allocation) is already thread local. ByteBufferIndexWrite itself is tiny, only 2 fields, so allocation cost is negligible.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: `Should we take graph properties to this calculation to consider alignment with device blocks?` - I'm afraid I don't really understand what you are asking here, can you point me to an example of what you are referring to, or go into a little more detail?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/GraphIndexWriterTypes.java
MarkWolters: Thanks!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
sam-herman: Please correct me if I'm misunderstanding, but it seems to me that for every ordinal we will (eventually) create a task object? If this is the case, can we avoid that?

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
marianotepper: Probably worth adding a comment about this for future reference.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/GraphIndexWriter.java
```diff
@@ -38,4 +53,62 @@ public interface GraphIndexWriter extends Closeable {
...
+     *   <li>ON_DISK_PARALLEL is not supported via this method (use the Path overload instead)</li>
+     * </ul>
+     *
+     * @param type the type of writer to create
+     * @param graphIndex the graph index to write
+     * @param out the output writer
+     * @return a builder for the specified writer type
+     * @throws IllegalArgumentException if the type requires a specific writer type that wasn't provided
+     */
+    static AbstractGraphIndexWriter.Builder<? extends AbstractGraphIndexWriter<?>, ? extends IndexWriter>
```
marianotepper, line 74: Would it be useful to separate AbstractGraphIndexWriter.Builder into a separate file so that we decouple the implementations from the interface in the code?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/GraphIndexWriterTypes.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    ON_DISK_SEQUENTIAL,
+
+    /**
+     * Parallel on-disk writer that uses asynchronous I/O for improved throughput.
+     * Builds records in parallel across multiple threads and writes them
+     * asynchronously using AsynchronousFileChannel.
+     * Requires a Path to be provided for async file channel access.
+     */
+    ON_DISK_PARALLEL
```
marianotepper, line 41: Do we need the prefix "ON_DISK" in this enum? Additionally, would it be better to use RANDOM instead of PARALLEL, to better indicate the default behavior? Since parallelization is optional...

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -0,0 +1,166 @@
...
+        this.view = view;
+        this.inlineFeatures = inlineFeatures;
+        this.featureStateSuppliers = featureStateSuppliers;
+        this.recordSize = recordSize;
+        this.fileOffset = fileOffset;
+        this.buffer = buffer;
+    }
+
+    @Override
+    public Result call() throws Exception {
```
marianotepper: I think that it would be useful for SequentialWriter and writeL0RecordsSequential to share this implementation somehow, so that we do not have code duplication.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
marianotepper: I'm also wondering about this. If we are wrong, it may be useful to add a comment in the code explaining this point. Others might trip here too in the future.

MarkWolters: You are correct in your summary @sam-herman, but it seemed to me like the natural splitting point. I could assign a range of ordinals to each task (which, I suppose it what you were suggesting by batching, which I misunderstood) instead of only a single ordinal. What do you think of that? cc:@marianotepper

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/GraphIndexWriter.java
MarkWolters: AbstractGraphIndexWriter.Builder is already defined in AbstractGraphIndexWriter, this is just a factory method to get the Builder appropriate for the type of Writer desired. I could move the implementation to AbstractGraphIndexWriter, but I would have to make it non-static, which would defeat the purpose, as you would need to have a GraphIndexWriter to get a Builder for one, or I could make it static in AbstractGraphIndexWriter, but that would defeat the purpose of having an interface.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -0,0 +1,362 @@
...
+import io.github.jbellis.jvector.graph.ImmutableGraphIndex;
+import io.github.jbellis.jvector.graph.disk.feature.Feature;
+import io.github.jbellis.jvector.graph.disk.feature.FeatureId;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.AsynchronousFileChannel;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.*;
```
marianotepper: Is it worth replacing the import `java.util.*` and other similar ones with specific imports?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
marianotepper: I don't have a specific suggestion here, besides the ones suggested above. Not sure if that one would be over tuning the implementation either. We could run a micro benchmark to see if memory usage increases significantly or not.

sam-herman: Hi @MarkWolters,

I've been thinking more about this, and aside from the memory allocation aspect, I have another concern related to disk alignment and throughput (as mentioned in my earlier comment regarding ParallelGraphWriter line 215).

Given a file of length `N` bytes and a disk capable of `k` concurrent writes, the ideal scenario for maximizing throughput would be to minimize the disk queue depth `Q`. In theory, if we have k threads each handling a contiguous buffer of size `B = N/k`, we could keep Q close to zero.

However, in the current implementation, it looks like the write of each node and it's adjacency list, are being distributed randomly among the threads. This seems to reduce the number of contiguous writes, which could increase the queue depth and lower overall throughput.

Is this understanding correct? Also, for the graphs you mentioned in the description, it might be helpful to measure the actual disk queue depth to assess how effectively we're maximizing throughput.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/GraphIndexWriterTypes.java
MarkWolters: Technically of course we can call it whatever we want. The naming is based on the fact that we have two main classes extending `AbstractGraphIndexWriter`: `OnDiskSequentialGraphIndexWriter` and `OnDiskGraphIndexWriter`, which optionally implements parallel writes, but still on disk. So at the moment both choices only write to disk. We could add more types, `ON_HEAP_*` etc. but the naming is meant to convey the message that you are creating a writer that writes to disk. That being said I'm OK with changing the names if they can be clearer.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: I 100% agree. My only question is whether that belongs in this PR or not. Originally the logic was duplicated between OnDiskSequentialGraphIndexWriter and OnDiskGraphIndexWriter, so there's not actually any new duplication being added here. If the consensus is that I should tackle that here I will do so. What do you think @marianotepper and @sam-herman?

jvector-base/src/main/java/io/github/jbellis/jvector/disk/ByteBufferIndexWriter.java
MarkWolters: Done

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: Done

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
marianotepper: True, we can do that in a subsequent PR.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: The eventual solution I settled on is similar to this suggestion, although it uses dataset size in terms of number of vectors rather than absolute file size. Monitoring the disk queue in Java would be tricky as we're talking about writing OS-specific code, and I don't think this enhancement should introduce that level of risk.

benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/ParallelWriteBenchmark.java
```diff
@@ -0,0 +1,287 @@
...
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.IntFunction;
+
+import static io.github.jbellis.jvector.quantization.KMeansPlusPlusClusterer.UNWEIGHTED;
+
+/**
+ * JMH benchmark that mirrors the ParallelWriteExample: it builds a graph from vectors, then
+ * writes the graph to disk sequentially and in parallel using NVQ + FUSED_ADC features,
```
tlwillke: The FUSED_ADC feature is now called FUSED_PQ.

benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/ParallelWriteBenchmark.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @Param({"1024"})
+    int dimension;
+
+    // Graph build parameters
+    final int M = 32;
+    final int efConstruction = 100;
+    final float neighborOverflow = 1.2f;
+    final float alpha = 1.2f;
+    final boolean addHierarchy = false;
```
tlwillke: Did you test parallel writes +/- hierarchy?

benchmarks-jmh/scripts/test_node_setup.sh
```diff
@@ -43,5 +43,5 @@ java --enable-native-access=ALL-UNNAMED \
   --add-modules=jdk.incubator.vector \
   -XX:+HeapDumpOnOutOfMemoryError \
   -Xmx14G -Djvector.experimental.enable_native_vectorization=true \
-  -jar target/benchmarks-jmh-4.0.0-beta.3-SNAPSHOT.jar
+  -jar target/benchmarks-jmh-4.0.0-rc.4-SNAPSHOT.jar
```
tlwillke, line 46: Any reason we can't advance this to the latest release?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.function.IntFunction;
+
+/**
+ * A task that builds L0 records for a range of nodes in memory.
+ * <p>
+ * This task is designed to be executed in a thread pool, with each worker thread
+ * owning its own ImmutableGraphIndex.View for thread-safe neighbor iteration.
+ * Each task processes a contiguous range of ordinals to reduce task creation overhead.
+ */
+class NodeRecordTask implements Callable<List<NodeRecordTask.Result>> {
```
tlwillke, line 38: Let's follow-up with an issue to make the same NodeRecordTask serve all of writers.  Trying to improve code quality.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
```diff
@@ -243,20 +356,53 @@ public synchronized void writeHeader(ImmutableGraphIndex.View view) throws IOExc
...
     }
 
-    /** CRC32 checksum of bytes written since the starting offset */
+    /** CRC32 checksum of bytes written since the starting offset
+     * Note on parallel writes and footer handling:
+     * When parallel writes are enabled (via {@link #setParallelWrites(boolean)}), it is the caller's responsibility
+     * to ensure that all parallel write operations have fully completed before writing the footer (e.g., checksum).
+     * The footer must only be written after all data has been flushed and no further writes are in progress,
+     * to avoid data corruption or incomplete checksums. This class does not currently coordinate or synchronize
+     * footer writing with parallel operations. Parallel writes are experimental and should be used with caution.
```
tlwillke: Let's decorate the experimental.

jvector-examples/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelWriteExample.java
```diff
@@ -0,0 +1,383 @@
...
+        var pq = ProductQuantization.compute(floatVectors, pqM, 256, centerData, UNWEIGHTED);
+        var pqVectors = (PQVectors) pq.encodeAll(floatVectors);
+        System.out.printf("PQ compression: %d subspaces, 256 clusters%n", pqM);
+
+        // Build graph parameters (matching typical benchmark settings)
+        int M = 32;
+        int efConstruction = 100;
+        float neighborOverflow = 1.2f;
+        float alpha = 1.2f;
+        boolean addHierarchy = false;
```
tlwillke: How about we enable hierarchy for the example since there are related checks?

tlwillke (commented):
Please address the comments.  Other LGTM.

benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/ParallelWriteBenchmark.java
MarkWolters: updated comment in latest push

benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/ParallelWriteBenchmark.java
MarkWolters: Yes. Test updated to run with addHierarchy parameterized to true/false

benchmarks-jmh/scripts/test_node_setup.sh
MarkWolters: updated

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexWriter.java
MarkWolters: `writeL0RecordsParallel` decorated as @Experimental, comments regarding parallel writes being experimental removed.

jvector-examples/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelWriteExample.java
MarkWolters: Done

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: See [Issue 579](https://github.com/datastax/jvector/issues/579)

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: See [Issue 579](https://github.com/datastax/jvector/issues/579)

tlwillke (approved):
Thanks for resolving all of the comments.

PR #543, opened 2025-10-15
harden tests for heap graph reconstruction

sam-herman:
# Description

- Harden test for the cases where heap graph reconstruction and ravv to graph ordinal mappings are not the identity mapping.
- Fix bug when graph ord mapping is used with the remmaped RAVV
- Move to remmaped interface for buildAndMerge to avoid bugs of remmpaing ords twice
- Add interface for providing the executor pools for the `buildAndMerge` method  -important for downstream dependencies with centralized resource management

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -439,14 +439,6 @@ public class GraphIndexBuilder implements Closeable {
     }
 
     public ImmutableGraphIndex build(RandomAccessVectorValues ravv) {
-        return reallyBuild(ravv);
-    }
-
-    public ImmutableGraphIndex build(RandomAccessVectorValues ravv, int[] graphToRavvOrdMap) {
-        return reallyBuild(new RemappedRandomAccessVectorValues(ravv, graphToRavvOrdMap));
-    }
-
-    public ImmutableGraphIndex reallyBuild(RandomAccessVectorValues ravv) {
         var vv = ravv.threadLocalSupplier();
         int size = ravv.size();
 
--- a/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/OnHeapGraphIndexTest.java
+++ b/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/OnHeapGraphIndexTest.java
@@ -150,7 +150,8 @@ public class OnHeapGraphIndexTest extends RandomizedTest  {
     public void testGraphConstructionWithNonIdentityOrdinalMapping() throws IOException {
         // create reversed mapping from graph node id to ravv ordinal
         int[] graphToRavvOrdMap = IntStream.range(0, baseVectorsRavv.size()).map(i -> baseVectorsRavv.size() - 1 - i).toArray();
-        var bsp = BuildScoreProvider.randomAccessScoreProvider(baseVectorsRavv, graphToRavvOrdMap, SIMILARITY_FUNCTION);
+        final RemappedRandomAccessVectorValues remappedBaseVectorsRavv = new RemappedRandomAccessVectorValues(baseVectorsRavv, graphToRavvOrdMap);
+        var bsp = BuildScoreProvider.randomAccessScoreProvider(remappedBaseVectorsRavv, SIMILARITY_FUNCTION);
         try (var baseGraphIndexBuilder = new GraphIndexBuilder(bsp,
                 baseVectorsRavv.dimension(),
                 M, // graph degree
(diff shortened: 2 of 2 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -439,6 +439,14 @@ public static GraphIndexBuilder rescore(GraphIndexBuilder other, BuildScoreProvi
     }
 
     public ImmutableGraphIndex build(RandomAccessVectorValues ravv) {
+        return reallyBuild(ravv);
+    }
+
+    public ImmutableGraphIndex build(RandomAccessVectorValues ravv, int[] graphToRavvOrdMap) {
```
marianotepper: We should not add this method here. It should be the responsibility of the caller to use the remapped RAVV, not a conversion that we want done by JV. Just like the change in buildAndMergeNewNodes

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -990,15 +996,47 @@ private void loadV3(RandomAccessReader in, int size) throws IOException {
...
+     * @param addHierarchy whether to add hierarchical structures while building the graph.
+     * @param simdExecutor the ForkJoinPool executor used for SIMD tasks during graph building.
+     * @param parallelExecutor the ForkJoinPool executor used for general parallelization during graph building.
+     *
+     * @return the in-memory representation of the graph index.
+     * @throws IOException if an I/O error occurs during the graph loading or conversion process.
+     */
+    @Experimental
+    public static ImmutableGraphIndex buildAndMergeNewNodes(RandomAccessReader in,
+                                                            RemappedRandomAccessVectorValues newVectors,
```
marianotepper, line 1022: This method should receive RandomAccessVectorValues isntead of RemappedRandomAccessVectorValues

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/OnHeapGraphIndexTest.java
```diff
@@ -166,12 +202,51 @@ public void testReconstructionOfOnHeapGraphIndex() throws IOException {
...
     }
 
+    /**
+     * Create an {@link OnHeapGraphIndex} persist it as a {@link OnDiskGraphIndex} and reconstruct back to a mutable {@link OnHeapGraphIndex}
+     * Using random mapping from graph node id to ravv ordinal
+     * Make sure that both graphs are equivalent
+     * @throws IOException
+     */
+    @Test
+    public void testReconstructionOfOnHeapGraphIndex_withNonIdentityOrdinalMapping() throws IOException {
```
marianotepper, line 213: Remove underscore in name, use camel case

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/OnHeapGraphIndexTest.java
```diff
@@ -129,14 +141,38 @@ public void tearDown() {
...
 
     /**
      * Create an {@link OnHeapGraphIndex} persist it as a {@link OnDiskGraphIndex} and reconstruct back to a mutable {@link OnHeapGraphIndex}
+     * Using identity mapping from graph node id to ravv ordinal
      * Make sure that both graphs are equivalent
      * @throws IOException
      */
     @Test
-    public void testReconstructionOfOnHeapGraphIndex() throws IOException {
+    public void testReconstructionOfOnHeapGraphIndex_withIdentityOrdinalMapping() throws IOException {
```
marianotepper, line 176: Remove underscore in name, use camel case

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/OnHeapGraphIndexTest.java
```diff
@@ -166,12 +202,51 @@ public void testReconstructionOfOnHeapGraphIndex() throws IOException {
...
+        }
+    }
+
     /**
      * Create {@link OnDiskGraphIndex} then append to it via {@link GraphIndexBuilder#buildAndMergeNewNodes}
      * Verify that the resulting OnHeapGraphIndex is equivalent to the graph that would have been alternatively generated by bulk index into a new {@link OnDiskGraphIndex}
      */
     @Test
-    public void testIncrementalInsertionFromOnDiskIndex() throws IOException {
+    public void testIncrementalInsertionFromOnDiskIndex_withIdentityOrdinalMapping() throws IOException {
```
marianotepper, line 250: Remove underscore in name, use camel case

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
sam-herman: Agree, will remove it!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
sam-herman: I think for this one we always take a `RemappedRandomAccessVectorValues` because we have to assume some ordinals between graph nodeIds and RAVV might not fit (remember that the graph was built with different RAVV instance), therefore to prevent user error we force remmaped type. We can use identity mapping in the edge case that there is not difference, but that is highly unlikely.

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/OnHeapGraphIndexTest.java
sam-herman: In Java for testing there is a naming convention of `[MethodUnderTest]_[Scenario]_[ExpectedResult]` for unit tests.
It's especially common when branching several scenarios under the same method being tested. With that being said I don't mind changing to whatever you feel is more readable :)

marianotepper: I did not know that. I learned something :-)
Ok with me then!

PR #548, opened 2025-10-15
javadoc for base [graph]

jshook:
This is the 2nd of the javadoc backfill PRs, covering only minor elisions in the base graph package.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
@@ -34,14 +34,14 @@ import java.util.stream.Collectors;
  * @param <T> the type of the output writer
  */
 public abstract class AbstractGraphIndexWriter<T extends IndexWriter> implements  GraphIndexWriter {
-    /** EOF magic number. */
-    public static final int FOOTER_MAGIC = 0x4a564244; // "EOF magic"
+    /** A magic number to indicate the file footer */
+    public static final int FOOTER_MAGIC = 0x4a564244;
     /** The size of the offset in the footer. */
-    public static final int FOOTER_OFFSET_SIZE = Long.BYTES; // The size of the offset in the footer
+    public static final int FOOTER_OFFSET_SIZE = Long.BYTES;
     /** The size of the magic number in the footer. */
-    public static final int FOOTER_MAGIC_SIZE = Integer.BYTES; // The size of the magic number in the footer
+    public static final int FOOTER_MAGIC_SIZE = Integer.BYTES;
     /** The total size of the footer. */
-    public static final int FOOTER_SIZE = FOOTER_MAGIC_SIZE + FOOTER_OFFSET_SIZE; // The total size of the footer
+    public static final int FOOTER_SIZE = FOOTER_MAGIC_SIZE + FOOTER_OFFSET_SIZE;
     final int version;
     final ImmutableGraphIndex graph;
     final OrdinalMapper ordinalMapper;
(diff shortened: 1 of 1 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
```diff
@@ -29,10 +29,18 @@
 import java.util.function.IntFunction;
 import java.util.stream.Collectors;
 
+/**
+ * Abstract base class for writing graph indexes to disk.
+ * @param <T> the type of the output writer
+ */
 public abstract class AbstractGraphIndexWriter<T extends IndexWriter> implements  GraphIndexWriter {
+    /** EOF magic number. */
```
marianotepper: This magic number does not mark EOF.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
```diff
@@ -150,6 +165,9 @@ void writeFooter(ImmutableGraphIndex.View view, long headerOffset) throws IOExce
      * <p>
      * Public so that you can write the index size (and thus usefully open an OnDiskGraphIndex against the index)
      * to read Features from it before writing the edges.
+     * @param view the graph index view
+     * @param startOffset the start offset
+     * @throws IOException if an I/O error occurs
      */
     public synchronized void writeHeader(ImmutableGraphIndex.View view, long startOffset) throws IOException {
```
marianotepper: Should this methods be made protected? @sam-herman @michaeljmarshall do we really need it public?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
jshook: Is the next line 33, incorrect then?
`public static final int FOOTER_MAGIC = 0x4a564244; // "EOF magic"`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
jshook: @marianotepper can we make the doc improvements anyway, given these two things are orthogonal?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: Yes.

jshook: What would you suggest as a better description here?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: I'd rather not, because the javadoc is trying to explain why the method is public, which may be just a fluke.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: It's a magic marker that is used to check for the presence of a footer.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
michaeljmarshall: CC does not use this method.

sam-herman: Can be protected, not using it directly.

marianotepper: @jshook I think that we can get this PR approved if we remove the lines in the doc of this function that try to justify why it is public.
```
     * <p>
     * Public so that you can write the index size (and thus usefully open an OnDiskGraphIndex against the index)
     * to read Features from it before writing the edges.
```
We can make the function private/protected in a separate PR.

jshook: I didn't see this last comment before I updated the PR to change the access modifier. The latest push contains a slight scope change to include modifying previous javadoc/comments as well as setting this to protected.

PR #550, opened 2025-10-17
Re-add OnDiskGraphIndex::getDimension

michaeljmarshall:
This `getDimension` method was removed from the `OnDiskGraphIndex` class via https://github.com/datastax/jvector/pull/536. We use it in C*, so I propose adding it back in since it is a getter on a final field.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
@@ -227,6 +227,7 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
         return features.keySet();
     }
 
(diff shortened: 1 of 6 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -227,6 +227,10 @@ public Set<FeatureId> getFeatureSet() {
         return features.keySet();
     }
 
+    public int getDimension() {
```
MarkWolters, line 231: How does this differ from dimension() below?

MarkWolters: never mind, that's a part of the view

marianotepper (approved):
LGTM

PR #551, opened 2025-10-22
CNDB-15703: Fix ScoreTracker impl init and reset methods

michaeljmarshall:
This PR returns the behavior of the `ScoreTracker` to before https://github.com/datastax/jvector/pull/501. There are two issues:

1. The `RelaxedMonotonicityTracker` had a fixed size `bestScores` at `new BoundedLongHeap(100)`. The previous behavio used a size that varied based on the `rerankK` for the query. Further, when the reset method is called, we do not update the size, and as such, we end up with a mismatched `bestScores` heap, which can result in an incorrect understanding of the `worstBestScore` in the `shouldStop` method. In cases where the `rerankK` is lower than 100, we will search deeper into the graph while in cases where it is above 100, we may return early.
2. The `ScoreTrackerFactory#getScoreTracker` has enough information to initialize more correct versions of the trackers, so I do that.

### DIFF
--- a/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestLowCardinalityFiltering.java
+++ b/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestLowCardinalityFiltering.java
@@ -1,3 +1,27 @@
+/*
+ * All changes to the original code are Copyright DataStax, Inc.
+ *
+ * Please see the included license file for details.
+ */
+
+/*
+ * Original license:
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


eolivelli:
which unit tests cover this change ?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ScoreTracker.java
```diff
@@ -211,7 +203,12 @@ void reset(int bestScoresTracked) {
             if (this.recentScoresSize > recentScores.length) {
                 recentScores = ArrayUtil.grow(recentScores, this.recentScoresSize);
             }
-            this.bestScores.clear();
+            if (this.recentScoresSize > bestScores.size()) {
+                bestScores = new BoundedLongHeap(this.recentScoresSize);
+            } else {
+                bestScores.clear();
+                bestScores.setMaxSize(this.recentScoresSize);
+            }
```
marianotepper: To match the constructor, bestScores should be resized using `bestScoresTracked` instead of `this.recentScoresSize`. That is:

```
if (this.recentScoresSize > bestScores.size()) {
    bestScores = new BoundedLongHeap(bestScoresTracked);
} else {
    bestScores.clear();
    bestScores.setMaxSize(bestScoresTracked);
 }
```

marianotepper (changes requested):
This is a good catch. Thanks for the contribution!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ScoreTracker.java
michaeljmarshall: Of course. I definitely looked at the variables and thought I picked the right one, but I clearly messed that one up! It's fixed, and it is simplified as well since the `BoundedLongHeap` resizes automatically and we do not need to handle that ourselves here.

marianotepper (approved):
LGTM

PR #553, opened 2025-10-22
memory usage estimator

jshook:
[ This will have an update with several improvements when time permits ]

This is a utility that we can use to get an empirical estimation of how much heap a given index or indexing process will take. It is a part of a larger strategy for managing heap usage safely.

Why? Here are the key reasons:

* We can take an index config and model "what if" heap usage very accurately, to understand how many vectors can reasonably fit on a system ahead of time.
* It is meant to provide advisory limits which we can use as a reliability gate without locking in bad administrative limits which do not size appropriately when things change
* This is a piece of a bigger idea to robustify heap usage.
* You do this once per index and you have a cost model for it without further tracking, counting work.

### DIFF
--- a/devdocs/memory_estimator.md
+++ b/devdocs/memory_estimator.md
@@ -48,13 +48,13 @@ Pick a sample size between 1,000 and 10,000 vectors; larger samples tighten the 
 
 ### Accuracy
 
-`MemoryCostEstimatorAccuracyTest` exercises the estimator against real graph builds with and without PQ. The measured footprint must stay within a 20% tolerance of the prediction. Current coverage includes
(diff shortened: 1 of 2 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/MemoryCostEstimator.java
```diff
@@ -0,0 +1,367 @@
...
+            this.pqSubspaces = pqSubspaces;
+            this.pqClusters = pqClusters;
+            this.pqCenter = pqCenter;
+        }
+
+        public static IndexConfig defaultConfig(int dimension) {
+            return new IndexConfig(
+                dimension,
+                16,
+                1.5f,
```
MarkWolters, line 92: Nit: We generally default to 1.2 for overflowRatio

jshook: noted, I'll follow this PR up with an improved one.

jshook: Just an update:

I have some improvements for this approach, but want to try to convert this as much as possible into a closed-form function which applies to the index config (yes, there are non-deterministic elements). This is very much like it works now except for the computation. There is some related work to anchor this in tests for accuracy bounds which hasn't been completed yet.

PR #555, opened 2025-10-22
Add RemappedRandomAccessVectorValues to fix BuildScoreProvider::randomAccessScoreProvider

michaeljmarshall:
This PR is meant to show an alternative to https://github.com/datastax/jvector/pull/554. The main idea is that we already have a mapping class, and if the caller wants to remap values, they can do it on their own. The primary benefit of this change is that is backwards compatible, while #554 is not.


MarkWolters (approved):
LGTM

PR #559, opened 2025-10-24
Make the thresholds in TestLowCardinalityFiltering tighter

marianotepper:
This PR makes the thresholds tighter in the newly introduced test TestLowCardinalityFiltering (#551). The previous thresholds were too loose. It also reduces the number of vectors as it is enough to monitor the number of visited nodes.


MarkWolters (approved):
LGTM

PR #561, opened 2025-11-03
Bring back the fused graph index 

marianotepper:
This PR does extensive work to bring back the Fused Graph Index (FGI). In a non-fused graph, the PQ codebook of each vector in the index is stored in memory. The memory complexity is the linear in the number of vectors. FGI reduces significantly the amount of heap memory used during search by offloading the PQ codebooks to storage. These PQ codebooks are packed and stored in-line with the graph, to avoid runtime overheads resulting from this offload.

The memory complexity has two cases now:
- When using a non-hierarchical graph, the fused graph reduces the linear memory complexity to a small constant (the number of vectors in the graph does not change this constant).
- When using a hierarchical graph, the upper layers of the hierarchy are kept in memory an the bottom layer is in storage. The PQ codebooks of those vectors in the upper layers are kept in memory. The bottom layer behaves exactly like a non-hierarchical graph. Since the upper graph layers are sampled using a logarithmic distribution, we end up with a logarithmic memory complexity.

These savings come with a very moderate slowdown (reduction in throughput and increase in latency) of about 15%. See the results below for an example.

In this version (and in past versions), FGI only works with PQ through the FUSED_PQ feature. This feature used to be called FUSED_ADC, but to highlight the link with PQ, it has been renamed.

The routine for expanding a node (gathering its out-neighbors and computing their similarities to the query), has been pushed down to the GraphIndex views. This enables having slightly different algorithms depending on the graph layout that may be a little bit more efficient than if abstracted away in the GraphSearcher.

This PR refactors the use of SIMD instructions by FUSED PQ:
- The old algorithm used a transposed layout similar to Quick(er)-ADC. However, that design performed the SIMD parallelization not for each codebook, but across different codebooks (thus the transpose). This parallelization was virtually impossible to combine with the skipping of the computation for previously computed similarities and implied additional computational overhead.
- An analysis of the number of skipped similarity computations yielded about 50% skips. thus, the new algorithm simplifies the layout by storing the vectors in a non-transposed fashion, with no across-codebook parallelization. This makes it compatible with the visited checks. Additionally, it is more efficient than the non-fused approach because we have improved the locality of the PQ codebooks to gather when expanding a given graph node (in the non-fused graph this required random accesses to multiple rows of a very tall and skinny matrix that exhibit poor locality).

These SIMD changes have opened the possibility of deprecating the native vector util backend. Not effecting this deprecation in this PR because there might be another considerations to keep it around.

## Edits:
- To enable the FUSED_PQ feature, we introduced the new version 6 file format for our graph indices.

# Experimental results:

Dataset: ada002-100k
Configuration:
  M                     : 32
  usePruning            : true
  neighborOverflow      : 1.2
  addHierarchy          : true
  efConstruction        : 100

## Results with topK=10

With a non-fused graph:
```
Overquery    Avg QPS (of 3)    ± Std Dev    CV %        Mean Latency (ms)    STD Latency (ms)    p999 Latency (ms)    Avg Visited    Avg Expanded Base Layer    Recall@10   
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1.00         16118.8           285.2        1.8         0.456                0.049               0.770                307.4          12.7                       0.67        
2.00         14100.7           19.5         0.1         0.507                0.064               0.802                421.6          22.9                       0.85        
5.00         11307.2           260.8        2.3         0.640                0.111               1.093                702.5          52.9                       0.94        
10.00        8335.0            88.8         1.1         0.849                0.184               1.587                1108.2         102.1                      0.97        
```

With a fused graph:
```
Overquery    Avg QPS (of 3)    ± Std Dev    CV %        Mean Latency (ms)    STD Latency (ms)    p999 Latency (ms)    Avg Visited    Avg Expanded Base Layer    Recall@10   
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1.00         13768.9           972.6        7.1         0.533                0.045               0.909                307.4          12.7                       0.67        
2.00         12384.5           582.9        4.7         0.586                0.057               0.966                421.6          22.9                       0.85        
5.00         9609.4            151.8        1.6         0.735                0.098               1.254                702.5          52.9                       0.94        
10.00        7135.4            162.5        2.3         0.955                0.163               1.603                1108.2         102.1                      0.97        
```

With the fused graph, the number of queries per second (QPS) is slowed down by less than 15% with an average of 14% and the latency by less than 17% with an average of 15%.

## Results with topK=100

With a non-fused graph:
```
Overquery    Avg QPS (of 3)    ± Std Dev    CV %        Mean Latency (ms)    STD Latency (ms)    p999 Latency (ms)    Avg Visited    Avg Expanded Base Layer    Recall@100   
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1.00         8314.4            307.9        3.7         0.862                0.187               1.570                1108.2         102.1                      0.78         
2.00         5415.1            17.0         0.3         1.294                0.302               2.293                1871.1         198.0                      0.93         
```

With a fused graph:
```
Overquery    Avg QPS (of 3)    ± Std Dev    CV %        Mean Latency (ms)    STD Latency (ms)    p999 Latency (ms)    Avg Visited    Avg Expanded Base Layer    Recall@100   
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1.00         6744.6            303.1        4.5         0.975                0.172               1.800                1108.2         102.1                      0.78         
2.00         4550.6            113.0        2.5         1.401                0.263               2.445                1871.1         198.0                      0.93         
```

With the fused graph, the number of queries per second (QPS) is slowed down by 19% and 16% (overquery=1 and 2, respectively)  with an average and the latency by 13% and 8% (overquery=1 and 2, respectively).

# Experimental results on larger datasets

In the plots below, QPS, latency, and recall are stable (there's run-to-run variability that is intrinsic to the benchmark). Index construction time increased a bit by the process of fusing the graph on disk, which involves multiple random memory accesses for each node, and writing more to disk.

Index_Construction_Time
Mean_Latency
QPS
Recall_at_10

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
@@ -167,20 +167,27 @@ public class Grid {
 
         try {
             for (var cpSupplier : compressionGrid) {
-                var compressor = getCompressor(cpSupplier, ds);
-                CompressedVectors cv;
-                if (compressor == null) {
-                    cv = null;
-                    System.out.format("Uncompressed vectors%n");
-                } else {
-                    long start = System.nanoTime();
-                    cv = compressor.encodeAll(ds.getBaseRavv());
-                    System.out.format("%s encoded %d vectors [%.2f MB] in %.2fs%n", compressor, ds.baseVectors.size(), (cv.ramBytesUsed() / 1024f / 1024f), (System.nanoTime() - start) / 1_000_000_000.0);
-                }
-
                 indexes.forEach((features, index) -> {
-                    try (var cs = new ConfiguredSystem(ds, index, cv,
-                                                       index instanceof OnDiskGraphIndex ? ((OnDiskGraphIndex) index).getFeatureSet() : Set.of())) {
+                    final Set<FeatureId> featureSetForIndex = index instanceof OnDiskGraphIndex ? ((OnDiskGraphIndex) index).getFeatureSet() : Set.of();
+
+                    CompressedVectors cv;
+                    if (featureSetForIndex.contains(FeatureId.FUSED_PQ)) {
+                        cv = null;
+                        System.out.format("Fused graph index%n");
+                    } else {
+                        var compressor = getCompressor(cpSupplier, ds);
+
+                        if (compressor == null) {
+                            cv = null;
+                            System.out.format("Uncompressed vectors%n");
+                        } else {
+                            long start = System.nanoTime();
+                            cv = compressor.encodeAll(ds.getBaseRavv());
+                            System.out.format("%s encoded %d vectors [%.2f MB] in %.2fs%n", compressor, ds.baseVectors.size(), (cv.ramBytesUsed() / 1024f / 1024f), (System.nanoTime() - start) / 1_000_000_000.0);
+                        }
+                    }
+
+                    try (var cs = new ConfiguredSystem(ds, index, cv, featureSetForIndex)) {
                         testConfiguration(cs, topKGrid, usePruningGrid, M, efConstruction, neighborOverflow, addHierarchy, benchmarks);
                     } catch (Exception e) {
                         throw new RuntimeException(e);
@@ -673,16 +680,16 @@ public class Grid {
         }
 
         public SearchScoreProvider scoreProviderFor(VectorFloat<?> queryVector, ImmutableGraphIndex.View view) {
-            // if we're not compressing then just use the exact score function
-            if (cv == null) {
-                return DefaultSearchScoreProvider.exact(queryVector, ds.similarityFunction, ds.getBaseRavv());
-            }
-
             var scoringView = (ImmutableGraphIndex.ScoringView) view;
             ScoreFunction.ApproximateScoreFunction asf;
             if (features.contains(FeatureId.FUSED_PQ)) {
                 asf = scoringView.approximateScoreFunctionFor(queryVector, ds.similarityFunction);
             } else {
+                // if we're not compressing then just use the exact score function
+                if (cv == null) {
+                    return DefaultSearchScoreProvider.exact(queryVector, ds.similarityFunction, ds.getBaseRavv());
+                }
+
                 asf = cv.precomputedScoreFunctionFor(queryVector, ds.similarityFunction);
             }
             var rr = scoringView.rerankerFor(queryVector, ds.similarityFunction);


UPGRADING.md
```diff
@@ -8,7 +8,14 @@
...
   HNSW-like hierarchy resides in memory for quickly seeding the search. This also reduces the need for caching the
   DiskANN graph near the entrypoint. The base layer of the hierarchy is a DiskANN-like index and inherits its
-  properties. This hierarchical structure can be disabled, ending up with just the base DiskANN layer.  
+  properties. This hierarchical structure can be disabled, ending up with just the base DiskANN layer.
+- The feature previously known as Fused ADC has been renamed to Fused PQ. This feature allows to offload the PQ
+  codebooks from memory during search, storing them within the graph in a way that does not slow down the search.
+  Implementation notes: The implementation of this feature has been overhauled to not require native code acceleration.
+  This explores a design space allowing for packed representations of vectors fused into the graph in shapes optimal 
+  for approximate score calculation. This new feature of graph indexes is opt-in but fully functional now. Any graph
+  degree limitations have been lifted. At this time, only 256-cluster ProductQuantization can use fused PQ.
```
michaeljmarshall, line 17: Worth mentioning that on disk format 6 or greater is required?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
```diff
@@ -142,6 +150,12 @@ interface View extends Closeable {
          */
         NodesIterator getNeighborsIterator(int level, int node);
 
+        /**
+         * Iterates over the neighbors of a given node if they have not been visited yet.
+         * For each non-visited neighbor, it computes its similarity and processes it using the given processor.
+         */
+        void processNeighbors(int level, int node, ScoreFunction scoreFunction, Function<Integer, Boolean> visited, NeighborProcessor neighborProcessor);
```
michaeljmarshall: Since we call the visited method on every node, it seems worth avoiding the autoboxing of int and boolean. Let's add an interface similar to `NeighborhoodProcessor`:

```java
@FunctionalInterface
public interface IntMarker {
    boolean mark(int value);
}
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
```diff
@@ -40,18 +49,18 @@ public abstract class AbstractGraphIndexWriter<T extends IndexWriter> implements
     final int dimension;
     // we don't use Map features but EnumMap is the best way to make sure we don't
     // accidentally introduce an ordering bug in the future
-    final EnumMap<FeatureId, Feature> featureMap;
+    final Map<FeatureId, Feature> featureMap;
```
michaeljmarshall, line 50: Is the comment out of date or are we at risk by introducing another map type below (`LinkedHashMap`)?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
```diff
@@ -60,8 +69,24 @@ public abstract class AbstractGraphIndexWriter<T extends IndexWriter> implements
...
-        this.inlineFeatures = features.values().stream().filter(f -> !(f instanceof SeparatedFeature)).collect(Collectors.toList());
+
+        if (version <= 5) {
+            // Versions <= 5 use the old feature ordering, simply provided by the FeatureId
+            this.featureMap = features;
+            this.inlineFeatures = features.values().stream().filter(f -> !(f instanceof SeparatedFeature)).collect(Collectors.toList());
+        } else {
+            // Version 6 uses the new feature ordering to place fused features last in the list
+            var sortedFeatures = features.values().stream().sorted().collect(Collectors.toList());
+            this.featureMap = new LinkedHashMap<>();
```
michaeljmarshall: Can we just use `new EnumMap<>(FeatureId.class)`?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
```diff
@@ -193,6 +218,58 @@ void writeSparseLevels(ImmutableGraphIndex.View view) throws IOException {
...
+                    throw new IllegalStateException("Mismatch between layer 1 size and features written");
+                }
+            } else {
+                // Write the source feature of the entry node
+                final int originalEntryNode = view.entryNode().node;
+                final int entryNode = ordinalMapper.oldToNew(originalEntryNode);
+                out.writeInt(entryNode);
+
+                // There should be only one fused feature per node. This is checked in the class constructor.
+                for (var feature : inlineFeatures) {
```
michaeljmarshall: Having a single `fusedFeature` variable would allow us to skip the iteration here (though I imagine the iteration is extremely fast). The primary benefit is in the class's design explicitly forcing one fused feature.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FeatureId.java
```diff
@@ -32,13 +30,11 @@
  */
 public enum FeatureId {
     INLINE_VECTORS(InlineVectors::load),
-    FUSED_ADC(FusedADC::load),
+    FUSED_PQ(FusedPQ::load),
```
michaeljmarshall, line 33: There is a small, probably zero, chance that this feature was used in a previous version of CC/HCD. Because we are repurposing this enum position, I propose that we fail the load if the graphs version is less than 6 so we can fail predictably and the message will be to recreate the graph.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
```diff
@@ -60,8 +69,24 @@ public abstract class AbstractGraphIndexWriter<T extends IndexWriter> implements
...
+            this.featureMap = new LinkedHashMap<>();
+            for (var feature : sortedFeatures) {
+                this.featureMap.put(feature.id(), feature);
+            }
+            this.inlineFeatures = sortedFeatures.stream().filter(f -> !(f instanceof SeparatedFeature)).sorted().collect(Collectors.toList());
+        }
+
+        if (this.inlineFeatures.stream().filter(Feature::isFused).count() > 1) {
+            throw new IllegalArgumentException("At most one fused feature is allowed");
+        }
```
michaeljmarshall, line 88: Could reduce the complexity here by removing the `fusedFeature` from the `inlineFeatures` list? There are a couple places we might still need the group of fused and inline features, but there are also places where we access the inline features only to get the fused feature.

What you have is already correct. My question really revolves around whether we need to modify the on disk format for the header, how we represent/access the features in this class, and if we can have just one way to order the features.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -152,6 +158,64 @@ private List<Int2ObjectHashMap<int[]>> loadInMemoryLayers(RandomAccessReader in)
...
+            if (current != null) {
+                return current;
+            }
+            try {
+                return loadInMemoryFeatures(in);
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
+            }
+        });
+    }
```
michaeljmarshall, line 178: What is the motivation of waiting to load in memory features until the first access? Seems like we could do this in the constructor, which is not on the query hot path, with the only drawback that if someone wanted to build the graph without searching it, they would "waste" this memory, but as I understand it, that it is a minimal amount of memory to begin with.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -75,13 +78,16 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
...
     final EnumMap<FeatureId, Integer> inlineOffsets;
+
     private final List<CommonHeader.LayerInfo> layerInfo;
     // offset of L0 adjacency data
     private final long neighborsOffset;
-    /** For layers > 0, store adjacency fully in memory. */
+    // For layers > 0, store adjacency fully in memory.
     private final AtomicReference<List<Int2ObjectHashMap<int[]>>> inMemoryNeighbors;
+    // When using fused features, store the features fully in memory for layers > 0
+    private final AtomicReference<Int2ObjectHashMap<FusedFeature.InlineSource>> inMemoryFeatures;
```
michaeljmarshall, line 90: It appears that these objects are not counted in the graph's `ramBytesUsed()` computations.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
```diff
@@ -71,7 +71,7 @@ public static void main(String[] args) throws IOException {
         );
         List<EnumSet<FeatureId>> featureSets = Arrays.asList(
                 EnumSet.of(FeatureId.NVQ_VECTORS),
-//                EnumSet.of(FeatureId.NVQ_VECTORS, FeatureId.FUSED_ADC),
+//                EnumSet.of(FeatureId.NVQ_VECTORS, FeatureId.FUSED_PQ),
                 EnumSet.of(FeatureId.INLINE_VECTORS)
```
michaeljmarshall, line 75: Do we want to uncomment this now that the feature is implemented?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: It is out of date.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FeatureId.java
marianotepper: Wouldn't that deprecate all previous versions? I think I'm missing something in your explanation.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: This is a good question. Let me think about it.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
marianotepper: That is a good point. It was the behavior for the hierarchy prior to this PR and I thought better not to mix that change with this one since there's a lot going on in this one. I'd rather do a different PR with that change.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
marianotepper: Good catch!

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
marianotepper: We should in reality deprecate Bench altogether. BenchYAML has become a much more handy point of entry. But I agree, I will uncomment this.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FeatureId.java
michaeljmarshall: IIUC, the fused adc feature was available before on disk format 6. I am proposing that when we load the features from disk, if we have fused pq and version < 6, we throw an exception.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
michaeljmarshall: Works for me.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
```diff
@@ -193,6 +218,58 @@ void writeSparseLevels(ImmutableGraphIndex.View view) throws IOException {
                 throw new IllegalStateException("Mismatch between layer size and nodes written");
             }
         }
+
+        // In V6, fused features for the in-memory hierarchy are written in a block after the top layers of the graph.
+        // Since everything in level 1 is also contained in the high-levels, we only need to write the fused features for level 1.
+        if (version == 6) {
```
eolivelli, line 226: is this strictly version == 6 or should it be version >= 6 like in another places in the code ?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/Header.java
```diff
@@ -40,30 +40,48 @@
...
         this.features = features;
     }
 
     void write(IndexWriter out) throws IOException {
         common.write(out);
 
-        if (common.version >= 3) {
-            out.writeInt(FeatureId.serialize(EnumSet.copyOf(features.keySet())));
-        }
+        if (common.version >= 6) {
```
eolivelli, line 56: What about creating a class with constants for all the version ?
so that we can document there which features are available in which version, like we do for SAI: 
https://github.com/datastax/cassandra/blob/82064a0255715c23dec03c5a94be122ac1ddc09e/src/java/org/apache/cassandra/index/sai/disk/format/Version.java#L57-L77

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: It is a good observation. The rule I followed is that subsequent versions might change this behavior anyway, so there's no benefit in expressing expectations over future versions in the code.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/Header.java
marianotepper: I'm not sure I fully get the benefit of such a change in JVector. Looks like we are dealing with a less combinatorial problem here. Can elaborate on the potential benefits of this approach? I'm open to consider it.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: There is only one place in the code where we need that unique fusedFeature (when in AbstractGraphIndexWriter.writeSparseLevels). I refactored the function so that the iteration happens in an outer loop and we create a local fusedFeature variable. I also added a comment saying that this local variable should be turned into a class member if needed elsewhere.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: Linked with this https://github.com/datastax/jvector/pull/561#discussion_r2498646825

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: Conceptually, I like the idea of introducing an ordering of the features. Of course, this is coupled with how things are handled in OnDiskGraphIndex.View.getPackedNeighbors, for example. I see that this coupling is somewhat unfortunate and it looks like the immediate solution is to not have that intrinsic ordering and just hardcode it in AbstractGraphIndexWriter. I am somewhat torn between a design that has more potential but may not be fully realized and one that is more constrained.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
marianotepper: This comment is linked with https://github.com/datastax/jvector/pull/561#discussion_r2495627475. In order to give a correct estimation of the heap usage, we need to either load these structures at initialization or get the information from the header. I think it just males sense to to the former and solve this once and for all.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
marianotepper: Linked to https://github.com/datastax/jvector/pull/561#discussion_r2498701784

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java
marianotepper: Done!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
marianotepper: Done!

UPGRADING.md
marianotepper: Done!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FeatureId.java
marianotepper: Done!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
marianotepper: Done!

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
marianotepper: Done!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -123,8 +130,7 @@ private List<Int2ObjectHashMap<int[]>> loadInMemoryLayers(RandomAccessReader in)
         var imn = new ArrayList<Int2ObjectHashMap<int[]>>(layerInfo.size());
         // For levels > 0, we load adjacency into memory
         imn.add(null); // L0 placeholder so we don't have to mangle indexing
-        long L0size = 0;
-        L0size = idUpperBound * (inlineBlockSize + Integer.BYTES * (1L + 1L + layerInfo.get(0).degree));
+        long L0size = idUpperBound * (inlineBlockSize + Integer.BYTES * (1L + 1L + layerInfo.get(0).degree));
```
michaeljmarshall, line 133: Historically, the `idUpperBound` has not been the same as the size of level 0. It would seem safer to use `layerInfo.get(0).size` here.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -152,6 +158,70 @@ private List<Int2ObjectHashMap<int[]>> loadInMemoryLayers(RandomAccessReader in)
...
+                }
+            }
+            return null;
+        });
+    }
+
+    private Int2ObjectHashMap<FusedFeature.InlineSource> loadInMemoryFeatures(RandomAccessReader in) throws IOException {
+        Int2ObjectHashMap<FusedFeature.InlineSource> hierarchyFeatures = new Int2ObjectHashMap<>();
+
+        long L0size = idUpperBound * (inlineBlockSize + Integer.BYTES * (1L + 1L + layerInfo.get(0).degree));
```
michaeljmarshall, line 183: I suggest we use `layerInfo.get(0).size` since `idUpperBound` is not necessarily correct here and we already have files written with invalid values there, but the size is correct in all cases.

michaeljmarshall: This was an invalid suggestion, the issue was a temporary value stored in CC. I have a PR that is already going to remove that logic, so no action is needed.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
michaeljmarshall: This was an invalid suggestion, the issue was a temporary value stored in CC. I have a PR that is already going to remove that logic, so no action is needed.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/Header.java
jshook: I would recommend using an enum here if we followed this pattern. It's a good idea in terms of type-safe and idiomatic Java. @marianotepper If you like I can take this as a follow-on change and we can decide on it once I have a PR to demonstrate.

jshook (approved):
@marianotepper It would be good to see the level of coverage of new/changed code here. Many of the changes are absolutely dependent on numerical and functional unit tests. It's non-trivial to see this overlap here.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/Header.java
marianotepper: Sounds good. we should move away from branches, I agree. I also agree adding this to a subsequent PR.

README.md
```diff
@@ -25,7 +25,7 @@ The upper layers of the hierarchy are represented by an in-memory adjacency list
 The bottom layer of the graph is represented by an on-disk adjacency list per node. JVector uses additional data stored inline to support two-pass searches, with the first pass powered by lossily compressed representations of the vectors kept in memory, and the second by a more accurate representation read from disk.  The first pass can be performed with
 * Product quantization (PQ), optionally with [anisotropic weighting](https://arxiv.org/abs/1908.10396)
 * [Binary quantization](https://huggingface.co/blog/embedding-quantization) (BQ)
-* Fused ADC, where PQ codebooks are transposed and written inline with the graph adjacency list
+* Fused PQ, where PQ codebooks are written inline with the graph adjacency list
 
 The second pass can be performed with
 * Full resolution float32 vectors
```
tlwillke, line 31: Shouldn't this also list NVQ as an option?

tlwillke (approved):
LGTM.  One comment added about a possible omission in README.md of NVQ as a second pass option.  Also would be great to see some empirical data on the memory savings.  The other performance data is thorough enough.

Thanks for this major contribution!

README.md
marianotepper: I will open a new PR about this.

marianotepper:
Slight refactor in Grid.runOneGraph so that the compressed vectors are loaded only when needed.

PR #563, opened 2025-11-07
Add ImmutableGraphIndex.isHierarchical

marianotepper:
Adds a method to the interface ImmutableGraphIndex that returns true if the index is hierarchical and false otherwise.

Note that a graph can be hierarchical even if it has a single layer, i.e., getMaxLevel() == 0. For example, while building a new hierarchical graph, we may temporarily only have nodes at level 0 because of the random assignment of nodes to levels.

This change enables to make public the GraphIndexBuilder constructor that takes an existing graph as an input, as we can infer the addHierarchy member from the provided graph.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
@@ -991,7 +991,7 @@ public class GraphIndexBuilder implements Closeable {
                                                             float overflowRatio,
                                                             float alpha) throws IOException {
 
-            return buildAndMergeNewNodes(in, newVectors, buildScoreProvider, startingNodeOffset, beamWidth, overflowRatio, alpha, addHierarchy, PhysicalCoreExecutor.pool(), ForkJoinPool.commonPool());
+            return buildAndMergeNewNodes(in, newVectors, buildScoreProvider, startingNodeOffset, beamWidth, overflowRatio, alpha, PhysicalCoreExecutor.pool(), ForkJoinPool.commonPool());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -568,7 +576,8 @@ public static OnHeapGraphIndex load(RandomAccessReader in, int dimension, double
 
         int entryNode = in.readInt();
 
-        var graph = new OnHeapGraphIndex(layerDegrees, dimension, overflowRatio, diversityProvider);
+        var isHierarchical = layerCount > 1;
```
akash-shankaran: nit. boolean?

akash-shankaran (approved):

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
akash-shankaran: maybe log isHierarchical somewhere as a debug message? That'll help with debugging.

jshook: This property being exposed as a test of layer count is inconsistent with the idea that a graph may be hierarchical during construction. If it is important for callers to know the intended graph mode during and after construction, then the property should be supplied as a ctor argument and returned instead of this structural test.

jshook: Or maybe this version of the index never exists as a reference type until it is constructed, which is ok.

jshook: ( I noticed that this was just for deriving the count )

jvector-tests/src/test/java/io/github/jbellis/jvector/TestUtil.java
```diff
@@ -249,6 +249,11 @@ public FullyConnectedGraphIndex(int entryNode, List<Integer> layerSizes) {
             this.layerSizes = layerSizes;
         }
 
+        @Override
+        public boolean isHierarchical() {
+            return layerSizes.size() > 1;
```
MarkWolters, line 254: based on the description of this PR layerSizes may have 1 or more layers that are currently 0 but will still be counted, correct?

marianotepper: Yes! In general, we can have graphs that have a single layer but that may eventually become multi-layered. This is because the assignment of vectors to layers happens randomly, and only a small subset ends up in the upper layers (all layers except the zero-th one). For the tests, just using the number of layers is fine in the vast majority of cases.

jshook (approved):
LGTM

Issue #565, opened 2025-11-12
Dynamic Product Re-Quantization


marianotepper:
## Current issues with dynamic updates to PQ quantization

The support in JVector for updating a Product Quantizer (PQ) is quite limited today. JVector only supports PQ updates through the `ProductQuantization.refine`. This is less than ideal for multiple reasons:
- The current implementation, applying one iteration of Lloyd's algorithm, provides no guarantees of working. In theory, each iteration of Lloyd's algorithm should lower the quantization error, but this is only true under the assumption that we eep the underlying data constant.
- The ability to update the quantizer is limited, because we do not have a way of removing deleted vectors (the centroids in Lloyd's algorithm carry "memory" of past iterations).

Additionally, there is not a real way of tracking the accuracy of the quantization in JVector.

## Proposal

- Change the k-means implementation from Lloyd's algorithm to the [mini-batch variant](https://citeseerx.ist.psu.edu/document?repid=rep1type=pdf&doi=b452a856a3e3d4d37b1de837996aa6813bedfdcf). This will allow to update the PQ codebook efficiently and with a small memory footprint.
- The prior change enables a Dynamic Product Quantization (DPQ) scheme, which is a more flexible way of updating the PQ codebook, supporting insertions and deletions, which are hard to handle with a partial update using Lloyd's algorithm. This would require adding a new deletion capability to the mini-batch algorithm. The change in the code is straightforward.
- Add mechanisms in JVector to track the accuracy of the quantizer. The only natural solution for this is to use the quantization (or reconstruction) error.
- Using the quantization error, we can implement a two quantizers approach. The vectors are quantized with a reference PQ. Additionally, we constantly update a DPQ over time (by applying insertions and deletions to it). The DPQ carries the most up-to-date representation of the data.
  - When we detect a meaningful difference between the quantization error of the reference PQ and the DPQ, the reference PQ has become "stale" and is discarded. Then, the DPQ becomes the reference PQ (i.e., we stop updating it), we re-quantize the vectors, and the process starts over.
  - To compute the quantization error, we need a sample of vectors. The sampling strategy is TBD. A possible solution is to additionally track the standard deviation of the k-means clusters as aprt of the mini-batch update and directly sample from the hard mixture of Gaussians.

## Implementation

- Add a MiniBatchKMeans implementation that has functions to add and remove vectors. Except for these new functions, this implementation should be a drop-in replacement for the current k-means implementation.
[...]

PR #567, opened 2025-11-17
Computation of reconstruction errors for vector compressors

marianotepper:
Given a vector quantizer, this PR introduces mechanisms to compute the reconstruction error for samples of vectors. The reconstruction error is basically the mean squared error (MSE) for each vector. This PR is part of #565.

For using the MSE, we need to consider that the norm of the vector will influence the MSE, so users might want to consider normalizing the errors.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/VectorCompressor.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/VectorCompressor.java
@@ -80,14 +80,18 @@ public interface VectorCompressor<T> {
     int compressedVectorSize();
 
     /**
-     * Compute the mean squared error for the given vector
+     * Compute the mean squared error (MSE) for the given vector.
+     * </p>
+     * MSE = (sum of squared errors over all dimensions) / (number of dimensions)
      * @param vector the vector to compute the reconstruction error for
      * @return the reconstruction error for the given vector
      */
     double reconstructionError(VectorFloat<?> vector);
 
     /**
-     * Compute the mean squared error for each vector in the stream.
+     * Compute the mean squared error (MSE) for each vector in the stream.
+     * </p>
+     * MSE = (sum of squared errors over all dimensions) / (number of dimensions)
      * @param ravv the vectors to compute the reconstruction error for
      * @return the reconstruction error for each vector
      */
@@ -96,7 +100,9 @@ public interface VectorCompressor<T> {
     }
 
     /**
-     * Compute the mean squared error for each vector in the stream in parallel.
+     * Compute the mean squared error (MSE) for each vector in the stream in parallel.
+     * </p>
+     * MSE = (sum of squared errors over all dimensions) / (number of dimensions)
      * @param ravv the vectors to compute the reconstruction error for
      * @param simdExecutor the ForkJoinPool to use for SIMD operations
      * @return the reconstruction error for each vector


jvector-base/src/main/java/io/github/jbellis/jvector/quantization/NVQuantization.java
```diff
@@ -355,6 +360,37 @@ public String toString() {
         return String.format("NVQuantization(sub-vectors=%d)", subvectorSizesAndOffsets.length);
     }
 
+    @Override
+    public double reconstructionError(VectorFloat<?> vector) {
+        var encodedVector = QuantizedVector.createEmpty(subvectorSizesAndOffsets, bitsPerDimension);
```
sam-herman: nit: just a standard practice to mark as `final` variables that are not to be re-assigned (same convention as in c++ with `const`)

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/BinaryQuantization.java
```diff
@@ -157,4 +157,15 @@ public int hashCode() {
     public String toString() {
         return "BinaryQuantization";
     }
+
+    @Override
+    public double reconstructionError(VectorFloat<?> vector) {
+        double sum = 0;
+        for (int i = 0; i < vector.length(); i++) {
```
sam-herman, line 164: should we use SIMD instructions to unroll and optimize this loop?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/VectorCompressor.java
```diff
@@ -74,4 +78,32 @@ default void write(DataOutput out) throws IOException {
...
+        return IntStream.range(0, ravv.size()).mapToDouble(i -> reconstructionError(ravv.getVector(i))).toArray();
+    }
+
+    /**
+     * Compute the mean squared error for each vector in the stream in parallel.
+     * @param ravv the vectors to compute the reconstruction error for
+     * @param simdExecutor the ForkJoinPool to use for SIMD operations
+     * @return the reconstruction error for each vector
+     */
+    default double[] reconstructionErrors(RandomAccessVectorValues ravv, ForkJoinPool simdExecutor) {
```
sam-herman, line 110: I would imagine we may not always want to do the full reconstruction error, for example in the case of large data sets (e.g. 100M) we would we prefer to do some sampling. Should we add a default interface method around that? Or we prefer to leave that for the downstream services to experiment?

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/VectorCompressor.java
```diff
@@ -74,4 +78,32 @@ default void write(DataOutput out) throws IOException {
 
     /** the size of a compressed vector */
     int compressedVectorSize();
+
+    /**
+     * Compute the mean squared error for the given vector
+     * @param vector the vector to compute the reconstruction error for
+     * @return the reconstruction error for the given vector
```
sam-herman, line 87: nit: Might be good to provide the formula in the comment to make the return value clear: 
```
MSE = (sum of squared errors over all dimensions) / (number of dimensions)
```

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/BinaryQuantization.java
marianotepper: We could do that. I have yet to see a problem instance where our current BinaryQuantization is better than PQ. I added the implementation for completeness, but I don't think it's worth devoting too much time to it.

jvector-base/src/main/java/io/github/jbellis/jvector/quantization/VectorCompressor.java
marianotepper: The RandomAccessVectorValues instance here can be a sample (a view over a larger RandomAccessVectorValues). That's how I was thinking about it.

Issue #586, opened 2025-12-03
NPE when calling OnDiskGraphIndex#ramBytesUsed


michaeljmarshall:
Looks like http://github.com/datastax/jvector/pull/561 might have introduced a regression for the `OnDiskGraphIndex#ramBytesUsed` method. I am getting this NPE:

```
java.lang.NullPointerException: Cannot invoke "org.agrona.collections.Int2ObjectHashMap.values()" because "neighbors" is null
    at io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex.ramBytesUsed(OnDiskGraphIndex.java:406)
```

PR #588, opened 2025-12-03
Fix OnDiskGraphIndex#ramBytesUsed NPE

michaeljmarshall:
Fixes: https://github.com/datastax/jvector/issues/586

Now that we lazily load the `inMemoryNeighbors` and the `inMemoryFeatures`, we need to handle the case where they are `null` or have values that are `null`. The added test fails without the change.


MarkWolters (approved):
LGTM

PR #589, opened 2025-12-03
virtualize DataSet

jshook:
This is part of a series of incremental changes to improve testing flexibility in JVector.
This one deals only with virtualizing the DataSet type, so that dataset identification, access and loading can more streamlined without breaking existing test logic.

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -132,7 +132,7 @@ jobs:
         uses: actions/upload-artifact@v4
         if: always() # ensures upload happens even if tests fail
         with:
-          name: surefire-results--${{ matrix.isa }}-${{ matrix.jdk }}
+          name: surefire-results--${{ matrix.os }}-${{ matrix.jdk }}
           path: "**/target/surefire-reports/**"
 
 


jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/DataSet.java
```diff
@@ -16,151 +16,61 @@
...
+ */
+public interface DataSet {
 
     /**
-     * Return a dataset containing the given vectors, scrubbed free from zero vectors and normalized to unit length.
-     * Note: This only scrubs and normalizes for dot product similarity.
+     * Get dimensions of the vectors in this dataset.
+     * @return the dimensionality
      */
-    public static DataSet getScrubbedDataSet(String pathStr,
```
MarkWolters, line 78: We will no longer have getScrubbedDataset on the presumption that all datasets should already be scrubbed during prep?

MarkWolters: Disregard, I see it's just been moved to the utils

PR #593, opened 2025-12-11
Virtualize and Modularize DataSetLoader logic

jshook:
This is the next step-wise change to streamline dataset loading and usage. It does the following:
- Virtualizes DataSetLoader, adds a central DataSets access point to simplify loading.
  - use Optional instead of checked IOExceptions
- Separates and modularizes HDF5, MFD loaders and consolidate helper methods.
- Organize dataset types into a package together.
- Update callers, with minor cleanups.
- Removes bench2d and its loader.

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
@@ -131,7 +131,7 @@ public class AutoBenchYAML {
                 logger.info("Loading dataset: {}", datasetName);
                 try {
                     DataSet ds = DataSets.loadDataSet(datasetName).orElseThrow(
-                            () -> new IllegalStateException("Dataset " + datasetName + " not found")
+                            () -> new RuntimeException("Dataset " + datasetName + " not found")
                     );
                     logger.info("Dataset loaded: {} with {} vectors", datasetName, ds.getBaseVectors().size());
 


jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
```diff
@@ -130,7 +130,9 @@ public static void main(String[] args) throws IOException {
 
                 logger.info("Loading dataset: {}", datasetName);
                 try {
-                    DataSet ds = DataSetLoader.loadDataSet(datasetName);
+                    DataSet ds = DataSets.loadDataSet(datasetName).orElseThrow(
+                            () -> new IllegalStateException("Dataset " + datasetName + " not found")
```
MarkWolters: Why IllegalStateException here? I believe all the other checks just use RuntimeException

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderMFD.java
```diff
@@ -0,0 +1,244 @@
...
+ */
+public class DataSetLoaderMFD implements DataSetLoader {
+
+    private static final Logger logger = LoggerFactory.getLogger(DataSetLoaderMFD.class);
+
+    private final static Set<String> infraDatasets = Set.of("dpr-1M", "dpr-10M", "cap-1M", "cap-6M", "cohere-english-v3-1M", "cohere-english-v3-10M");
+    private static final String infraBucketName = "jvector-datasets-infratest";
+    private static final String fvecDir = "fvec";
+    private static final String bucketName = "astra-vector";
+    private static final List<String> bucketNames = List.of(bucketName, infraBucketName);
```
MarkWolters, line 52: We should probably just move everything to the infratest bucket to avoid having to use multiple buckets here, although that's not a comment on this PR. This is fine, just follows the existing logic.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderMFD.java
```diff
@@ -0,0 +1,244 @@
...
+                    FileDownload downloadFile = tm.downloadFile(downloadFileRequest);
+                    CompletedFileDownload downloadResult = downloadFile.completionFuture().join();
+                    long downloadedSize = Files.size(localPath);
+
+                    // Check if downloaded file size matches the expected size
+                    if (downloadedSize == downloadResult.response().contentLength()) {
+                        logger.info("Downloaded file of length " + downloadedSize);
+                        break;  // Successfully downloaded
+                    } else {
+                        logger.error("Incomplete download. Retrying...");
```
MarkWolters, line 107: I wonder if we should delete the partial file if it can't get a complete download in 3 tries

jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
jshook: good catch, fixing

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderMFD.java
```diff
@@ -0,0 +1,244 @@
...
+import software.amazon.awssdk.transfer.s3.model.FileDownload;
+import software.amazon.awssdk.transfer.s3.progress.LoggingTransferListener;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
+
+/**
```
marianotepper, line 40: I suggest making this name clearer. Or at least the description of the class. The main use case for this class is to have support for fvec/ivec formats. It should probably have a link to http://corpus-texmex.irisa.fr/ where the format is explained.

jshook: I'll take a look and see what I can do to clarify this.

PR #596, opened 2026-01-15
add avx512 to panama provider

akash-shankaran:
Add AVX512 support for Dot Product


ashkrisk:
AVX512 should automatically be picked up by `dotProductPreferred`, since the preferred vector species will already be `SPECIES_512` on supported platforms. If I'm not mistaken, this PR is primarily useful for future-proofing in cases where a bigger vector species might exist (say 1024-bits), but the vector length is in-between 512 and 1024 bits. As a suggestion, how about changing the PR title to reflect that?

PR #597, opened 2026-01-20
Add introductory documentation

ashkrisk:
Add better documentation for anyone getting started with JVector.


docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
+# How to run benchmarks
+
+JVector comes with a built-in benchmarking system in `jvector-examples/.../BenchYAML.java`.
+
+To run a benchmark
+- Decide which dataset(s) you want to benchmark.
+- Configure the parameters combinations for which you want to run the benchmark. This includes graph index parameters, quantization parameters and search parameters.
+
+To describe a dataset you will need to specify the base vectors used to construct the index, the query vectors, and the "ground truth" results which will be used to compute accuracy metrics.
```
marianotepper: This sentence is very vague. What about we patch the previous list as follows?
- Decide which dataset(s) you want to benchmark. A dataset is specified by the vectors to be indexed (usually called base or target vectors), the query vectors, the "ground truth" results that are used to compute accuracy metrics, and the similarity metric (which should have been used to compute the ground truth).

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+
+JVector supports two types of datasets:
+- **Fvec/Ivec**: The dataset consists of three files, for example `base.fvec`, `queries.fvec` and `neighbors.ivec`.
```
marianotepper: Add link to http://corpus-texmex.irisa.fr/ where these formats are described

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+- **HDF5**: The dataset consists of a single HDF5 file with three datasets labelled `train`, `test` and `neighbors`, representing the base vectors, query vectors and the ground truth.
+
+General procedure for running benchmarks:
+- Specify the dataset names to benchmark in `datasets.yml`.
```
marianotepper: This is unclear. Do we need to modify datasets.yml, or chose from it? If the latter, where do we specify the chosen ones?

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+- Certain datasets will be downloaded automatically. If using a different datasets, make sure the dataset files are downloaded and made available (refer the section on [using datasets](#using-datasets)).
```
marianotepper: Typo: a different datasets -> a different dataset

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+You can run the configured benchmark with maven:
+```sh
+mvn clean compile exec:exec@bench -pl jvector-examples -am
+```
+
+## Using Datasets
+
+### Using Fvec/Ivec datasets
+
+Using fvec/ivec datasets requires them to be configured in `MultiFileDatasource.java`. Some datasets are already pre-configured; these will be downloaded and used automatically on running the benchmark.
```
marianotepper, line 86: This section needs some tweaking following the latest changes from @jshook. Maybe @jshook  can help to adjust this section?

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+- Save all three files somewhere in the `fvec` directory in the root of the `jvector` repo (if it doesn't exist, create it). It's recommended to create at least one sub-folder with the name of the dataset and copy or move all three files there.
+- Edit `MultiFileDatasource.java` to configure a new dataset and it's associated files:
+    ```java
+    put("cust-ds", new MultiFileDatasource("cust-ds",
+            "/cust-ds/base.fvec",
+            "/cust-ds/query.fvec",
+            "/cust-ds/neighbors.ivec"));
+    ```
+    The file paths are resolved relative to the `fvec` directory. `cust-ds` is the name of the dataset and can be changed to whatever is appropriate.
+- In `jvector-examples/yaml-configs/datasets.yml`, add an entry corresponding to your custom dataset. Comment out other datasets which you don't want to benchmark.
```
marianotepper: don't -> do not

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+    ```
+
+### Using HDF5 datasets
+
+HDF5 datasets consist of a single file. The Hdf5Loader looks for three HDF5 datasets within the file, `train`, `test` and `neighbors`. These correspond to the base, query and neighbors vectors described above for fvec/ivec files.
+
+To use an HDF5 dataset, edit `jvector-examples/yaml-configs/datasets.yml` to add an entry like the following:
+```yaml
+category:
+  - dataset-name.hdf5
```
marianotepper: Should we use a convention like [your dataset name].hdf5 for extra clarity? Or is ti clear enough already?

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+- `default.yml` which defines the default parameter sets to be used for all datasets.
+- `dataset-name.yml` which specifies the parameter sets for a single dataset.
+
+### datasets.yml
+
+This file specifies the datasets to be used when running `BenchYAML`. Datasets are grouped into categories. The categories can be arbitrarily chosen for convenience and are not currently considered by the benchmarking system.
+
+### default.yml / \<dataset-name\>.yml
+
+These files define the parameters to be used by `BenchYAML`. The settings in the `default.yml` file apply to all datasets, except ones which have a custom configuration defined in `<dataset-name>.yml`.
```
marianotepper: except ones which have a -> except the ones with a

docs/draft-hello.md
```diff
@@ -0,0 +1,150 @@
...
+    vts.createFloatVector(new float[]{0.01f, 0.15f, -0.3f}),
+    vts.createFloatVector(new float[]{-0.2f, 0.1f, 0.35f})
+);
+
+// RAVV or `ravv` is convenient shorthand for a RandomAccessVectorValues instance
+RandomAccessVectorValues ravv = new ListRandomAccessVectorValues(baseVectors, dimension /* 3 */);
+```
+
+> [!TIP]
+> In this example, all vectors are loaded in-memory, but RAAVs are quite versatile. For example, you might have a RAVV backed by disk (check out `MMapRandomAccessVectorValues.java`) or write your own custom RAAV that transfers data over a network interface.
```
marianotepper: RAAVs -> RAVVs

docs/draft-hello.md
```diff
@@ -0,0 +1,150 @@
...
+
+```java
+// The type of similarity score to use. JVector supports EUCLIDEAN (L2 distance), DOT_PRODUCT and COSINE.
+VectorSimilarityFunction similarityFunction = VectorSimilarityFunction.EUCLIDEAN;
+
+// A simple score provider which can compute exact similarity scores by holding a reference to all the base vectors.
+BuildScoreProvider bsp = BuildScoreProvider.randomAccessScoreProvider(ravv, similarityFunction);
+```
+
+Let's also initialize the graph parameters. For now we won't worry about the exact function of the parameters, except to note that the below is a reasonable set of defaults. Refer the DiskANN and HNSW papers for more details.
```
marianotepper: won't -> will not
except to note that the below is a reasonable set of defaults -> except to note that these are reasonable defaults

docs/draft-hello.md
```diff
@@ -0,0 +1,150 @@
...
+                                                  alpha,
+                                                  addHierarchy,
+                                                  refineFinalGraph);
+ImmutableGraphIndex graph = builder.build(ravv);
+```
+
+> [!NOTE]
+> You may notice that we supplied the same `ravv` to `builder.build`, even though we'd already passed in the RAVV while creating the `BuildScoreProvider`. This is necessary since generally speaking, the `BuildScoreProvider` won't keep a reference to the actual base vectors, it just so happens that we're using an "exact" score provider that does so.
+
+At this point, you have a completed Graph Index that's entirely fully in-memory.
```
marianotepper: that's entirely fully in-memory -> that resides in-memory

docs/draft-hello.md
```diff
@@ -0,0 +1,150 @@
...
+
+To perform a search operation, you need to first create a `GraphSearcher`.
+
+> [!IMPORTANT]
+> The graph index itself can be shared between threads, but `GraphSearcher`s maintain internal state and are therefore NOT thread-safe.
```
marianotepper: Add something like: You can still run search concurrently, but need to create one GraphSearcher for each thread.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/VectorIntro.java
```diff
@@ -0,0 +1,97 @@
...
+import io.github.jbellis.jvector.graph.similarity.BuildScoreProvider;
+import io.github.jbellis.jvector.graph.similarity.DefaultSearchScoreProvider;
+import io.github.jbellis.jvector.graph.similarity.SearchScoreProvider;
+import io.github.jbellis.jvector.util.Bits;
+import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
+import io.github.jbellis.jvector.vector.VectorizationProvider;
+import io.github.jbellis.jvector.vector.types.VectorFloat;
+import io.github.jbellis.jvector.vector.types.VectorTypeSupport;
+
+public class VectorIntro {
```
marianotepper, line 42: There should be a way to grab sections of this file from draft-hello.md, so that we do not have to manually sync both files.

marianotepper (commented):
Looks good in general. Some minor comments here and there.
There is one section that may need an update following some recent changes by @jshook.

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
+# How to run benchmarks
```
MarkWolters, line 1: Needs license or needs to be added to rat-excludes

docs/benchmarking.md
```diff
@@ -0,0 +1,90 @@
...
+Using fvec/ivec datasets requires them to be configured in `MultiFileDatasource.java`. Some datasets are already pre-configured; these will be downloaded and used automatically on running the benchmark.
+
+To use a custom dataset consisting of files `base.fvec`, `queries.fvec` and `neighbors.ivec`, do the following:
+- Ensure that you have three files:
+    - `base.fvec` containing N D-dimensional float vectors. These are used to build the index.
+    - `queries.fvec` containing Q D-dimensional float vectors. These are used for querying the built index.
+    - `neighbors.ivec` containing Q K-dimensional integer vectors, one for each query vector, representing the exact K-nearest neighbors for that query among the base vectors.
+    The files can be named however you like.
+- Save all three files somewhere in the `fvec` directory in the root of the `jvector` repo (if it doesn't exist, create it). It's recommended to create at least one sub-folder with the name of the dataset and copy or move all three files there.
+- Edit `MultiFileDatasource.java` to configure a new dataset and it's associated files:
```
MarkWolters, line 95: This section also needs to describe how we are using the environment variable DATASET_HASH and how it needs to be set on the target system or the downloads will fail

docs/tutorials/1-intro-tutorial.md
```diff
@@ -0,0 +1,150 @@
+# JVector Tutorial
```
MarkWolters, line 1: Needs license or needs to be added to rat-excludes

docs/tutorials/1-intro-tutorial.md
```diff
@@ -0,0 +1,150 @@
...
+SearchScoreProvider ssp = DefaultSearchScoreProvider.exact(queryVector, similarityFunction, ravv);
+```
+
+Now we can run a search
+
+```java
+int topK = 10;  // number of approximate nearest neighbors to fetch
+// You can provide a filter to the query as a bit mask.
+// In this case we want the actual topK neighbors without filtering,
+// so we pass in a virtual bit mask representing all ones.
```
MarkWolters, line 133: An example of what is meant by filtering when something other than Bits.ALL is used might be useful here.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/VectorIntro.java
```diff
@@ -0,0 +1,97 @@
+package io.github.jbellis.jvector.example;
```
MarkWolters, line 17: Needs license or needs to be added to rat-excludes

MarkWolters (changes requested):

docs/benchmarking.md
ashkrisk: Folded this into the previous list.

docs/benchmarking.md
ashkrisk: done

docs/benchmarking.md
ashkrisk: Added a link to a following section which describes how "specifying datasets" works, and also reworked the section to make it more clear. Let me know if you'd prefer I elaborate a bit here.

docs/benchmarking.md
ashkrisk: done

docs/benchmarking.md
ashkrisk: done

docs/benchmarking.md
ashkrisk: Added angle brackets `<...>` to make it more obvious. This coupled with the expansion of the section on `datasets.yml` should make things quite clear. I also moved that section up above this.

docs/benchmarking.md
ashkrisk: done

docs/benchmarking.md
ashkrisk: This would be applicable to specific pre-configured datasets (cohere 1M-10M, dpr etc) that are defined in `MultiFileDatasource`. If I'm not mistaken, accessing these datasets requires not only the dataset hash, but also the credentials of the corresponding S3 bucket. I think it's alright to skip it for now considering that the affected datasets are not public.

Note that this won't prevent the user from externally downloading and using any dataset they have access to.

docs/draft-hello.md
ashkrisk: done

docs/draft-hello.md
ashkrisk: done

docs/draft-hello.md
ashkrisk: done

jvector-examples/src/main/java/io/github/jbellis/jvector/example/VectorIntro.java
ashkrisk: A quick couple of searches didn't turn up a simple way to do this such that it works in all contexts. I don't think we should spend too much energy on this, because anyone changing `VectorIntro.java` _should_ be making sure that `intro-tutorial.md` is also in sync, considering that it's content is heavily reliant on `VectorIntro`. I'll add comment to `VectorIntro.java` making the link more clear.

docs/tutorials/1-intro-tutorial.md
ashkrisk: I wanted to avoid digressing too much at this point. I've only mentioned filtering since I can't run a search without passing in _some_ `Bits` instance, and I didn't want to do that with zero commentary. I was planning to elaborate further in other docs, but do you think it's too confusing without it?

docs/benchmarking.md
ashkrisk: done (added docs to rat-excludes)

docs/tutorials/1-intro-tutorial.md
ashkrisk: done

jvector-examples/src/main/java/io/github/jbellis/jvector/example/VectorIntro.java
ashkrisk: done

docs/draft-hello.md
ashkrisk: done

docs/benchmarking.md
ashkrisk: As far as I can tell there's no change required in the content of this section, @jshook would if you notice something please let me know.

PR #602, opened 2026-01-27
add protection against bad ordinal mappings

MarkWolters:
Adding protections against bad values for ordinal mappings being passed in to PQVectors


tlwillke (approved):
LGTM.

PR #608, opened 2026-02-02
Refactor parallel writer

MarkWolters:
Sequence diagram of current OnDiskGraphIndexWriter usage by Cassandra:
[Cassandra_OnDiskGraphIndexWriter_CurrentState_SequenceDiagram.md](https://github.com/user-attachments/files/24675240/Cassandra_OnDiskGraphIndexWriter_CurrentState_SequenceDiagram.md)

Sequence diagram of proposed future OnDiskParallelGraphIndexWriter usage:
[OnDiskParallelGraphIndexWriter_SequenceDiagram.md](https://github.com/user-attachments/files/24675241/OnDiskParallelGraphIndexWriter_SequenceDiagram.md)

Perf test results:
[refactor_parallel.tar.gz](https://github.com/user-attachments/files/24674906/refactor_parallel.tar.gz)

Refactoring of the parallelization of graph index writer.

This PR splits the parallel writer into a separate class rather than maintaining if-based branches throughout a single class (OnDiskGraphIndexWriter). A large amount of common code has been abstracted into the new RandomAccessOnDiskGraphIndexWriter making the hierarchy cleaner and easier to understand and maintain.

Previously it was discovered that calling write() after calling writeInline() would results in the features from writeInline() being overwritten with zeroes. This is resolved in this case by checking for feature provider being null, emulating how it is done in sequential writes.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/ProductQuantization.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/quantization/ProductQuantization.java
@@ -29,27 +29,24 @@ import io.github.jbellis.jvector.vector.VectorizationProvider;
 import io.github.jbellis.jvector.vector.types.ByteSequence;
 import io.github.jbellis.jvector.vector.types.VectorFloat;
 import io.github.jbellis.jvector.vector.types.VectorTypeSupport;
+import org.agrona.collections.IntHashSet;
 
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.SplittableRandom;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Supplier;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
-import java.util.stream.DoubleStream;
 import java.util.stream.IntStream;
-import java.util.stream.Stream;
 
 import static io.github.jbellis.jvector.quantization.KMeansPlusPlusClusterer.UNWEIGHTED;
 import static io.github.jbellis.jvector.util.MathUtil.square;
 import static io.github.jbellis.jvector.vector.VectorUtil.dotProduct;
 import static io.github.jbellis.jvector.vector.VectorUtil.sub;
-import static java.lang.Math.min;
 import static java.lang.Math.sqrt;
 
 /**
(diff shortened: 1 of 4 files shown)


michaeljmarshall:
Any reason not to merge my PR first since it is a discrete bug needing fixing? Not a big deal either way, just curious.

MarkWolters:
@michaeljmarshall no, merging yours first would be fine, I just wanted you to be aware that I needed your fix in my branch or I couldn't merge it as it wouldn't pass the gate for the automated tests.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -81,41 +74,75 @@ static class Result {
...
+     * @param position the file position to write at
+     * @throws IOException if an I/O error occurs
+     * @throws ExecutionException if the write operation fails
+     * @throws InterruptedException if interrupted while waiting for write completion
+     */
+    private static void writeFully(AsynchronousFileChannel channel, ByteBuffer buffer, long position)
+            throws IOException, ExecutionException, InterruptedException {
+        long currentPosition = position;
+        while (buffer.hasRemaining()) {
+            int written = channel.write(buffer, currentPosition).get();
```
michaeljmarshall, line 97: If we're going to call `.get()` on the `AsynchronousFileChannel`'s write future, we can skip the overhead and use a synchronous file writer via a `FileChannel`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -81,41 +74,75 @@ static class Result {
...
+            long currentPosition = recordOffset;
 
             // Reset buffer for this ordinal
             writer.reset();
 
-            var originalOrdinal = ordinalMapper.newToOld(newOrdinal);
-
             // Write node ordinal
             writer.writeInt(newOrdinal);
+            ByteBuffer ordinalData = writer.cloneBuffer();
```
michaeljmarshall, line 120: IIUC, if we're not actually using async io, we can skip cloning the buffer.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -81,41 +74,75 @@ static class Result {
...
+            writeFully(channel, ordinalData, currentPosition);
```
michaeljmarshall, line 121: I thought we sized the write buffer to be the size of a node? If that's the case, can't we defer writing until the end of the `call` method?

MarkWolters: No. That approach doesn't work because of the possibility that writeInline() was already called, in which case there will already be data on disk in the file. That would necessitate reading from the file in order to insert it into the buffer before writing, which is not something a writer should be doing.

michaeljmarshall: Good point, we used to have that pattern in CC until a few months ago, and that is definitely what ODGIW is doing. CC moved away from it to support NVQ.

In the long run, I think we'll see a significant performance boost if we can buffer the whole node record task and then write it once. Maybe it'd be possible to add a boolean flag to branch behaviors or to have two different `NodeRecordTask` implementations?

PR #612, opened 2026-02-05
On-disk index cache for the Grid benchmark harness

tlwillke:
This PR adds a deterministic on-disk index cache for the Grid benchmark harness and wires it in end-to-end so repeated runs can reuse previously-built graph indexes to save time.

Key changes

1. Introduced `OnDiskGraphIndexCache` (flat directory cache, one file per index) keyed by a stable signature derived from:

-  dataset base name
-  feature set (per-index; not dependent on the list of feature sets)
-  build params (M, efConstruction, neighborOverflow, addHierarchy, refineFinalGraph)
-  build compressor identity

2. Cache filenames are derived from the signature (sanitized for filesystem safety). This allows multiple cached indexes to coexist in a single flat cache directory without collisions.

3. Updated `Grid.runOneGraph` to treat caching per-index (per feature set):

- load cached indexes when present
- build only the missing ones and merge results
- keep non-cached builds writing into the temp work directory using the original `graphN` naming

4. Refactored `buildOnDisk` so it can write either:

- the original `graph0..graphN` temp files (cache disabled), or
- signature-named files in the cache directory (cache enabled), while preserving existing build behavior and minimizing churn.

5. Updated Bench, BenchYAML dataset files, and HelloVectorWorld

- disabled the index cache by default
- easy enabling of the cache if desired (useSavedIndexIfExists or enableIndexCache)

6. Improved logging so it’s obvious when a cached index is used vs built from scratch, and added a cache-enabled startup message pointing to the cache directory and how to reclaim disk.

Behavior

- Cache enabled: reuse cached indexes when signatures match; build only missing feature sets; artifacts persist in the cache directory.
- Cache disabled: no cache reads/writes; build into temp work directory and clean up as before.  NOTE: cached files indeed persist and must be explicitly deleted.

Notes

- Signatures are per-index (per feature set) and include all build-defining params to prevent accidental reuse across incompatible builds.
- Filenames are sanitized to avoid filesystem issues from compressor IDs or other signature components.
- No impact on graph construction or other benchmarking performance
- Cleaned up a few minor bugs in the existing code
- Does not address disk use metrics (disk space or file count).  They report 0 for cached indexes (a TODO)

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
@@ -154,12 +154,12 @@ public class AutoBenchYAML {
                     logger.info("Using configuration: {}", config);
 
                     List<BenchResult> datasetResults = Grid.runAllAndCollectResults(ds,
-                            false,
-                            null,
+                            config.construction.useSavedIndexIfExists,
                             config.construction.outDegree, 
                             config.construction.efConstruction,
                             config.construction.neighborOverflow, 
                             config.construction.addHierarchy,
+                            config.construction.refineFinalGraph,
                             config.construction.getFeatureSets(), 
                             config.construction.getCompressorParameters(),
                             config.search.getCompressorParameters(), 
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
@@ -63,14 +63,7 @@ import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Comparator;
-import java.util.EnumMap;
-import java.util.HashMap;
-import java.util.IdentityHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Function;
 import java.util.function.IntFunction;
 import java.util.stream.IntStream;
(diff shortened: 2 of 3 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
```diff
@@ -153,7 +153,9 @@ public static void main(String[] args) throws IOException {
                     }
                     logger.info("Using configuration: {}", config);
 
-                    List<BenchResult> datasetResults = Grid.runAllAndCollectResults(ds, 
+                    List<BenchResult> datasetResults = Grid.runAllAndCollectResults(ds,
```
MarkWolters, line 156: The signature of this method in Grid is changed to include the new variables for enableIndexCache and indexCacheDirectory but they are never used. Elsewhere in BenchYAML the existing MultiConfig property useSavedIndexIfExists is used as a stand-in for enableIndexCache with the Grid class level index cache directory used as the index cache directory. It might make sense to use the same pattern here.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
```diff
@@ -0,0 +1,349 @@
...
+    public static final class CacheKey {
+        public final String datasetName;
+        public final Set<FeatureId> featureSet;
+        public final int M;
+        public final int efConstruction;
+        public final float neighborOverflow;
+        public final boolean addHierarchy;
+        public final boolean refineFinalGraph;
+        public final String compressorId;
+
```
jshook, line 132: Do you want to add a major or minor jvector version indicator here? Could the formats change between versions and invalidate the cache format without this code knowing?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
```diff
@@ -42,6 +39,10 @@ public class Bench {
     public static void main(String[] args) throws IOException {
         System.out.println("Heap space available is " + Runtime.getRuntime().maxMemory());
 
+        // When enabled, caches built indices for reuse in future runs.
+        // Useful for large indexes and repeated testing.
+        boolean enableIndexCache = false;
```
jshook, line 44: What's the harm in enabling this by default? Is there a simple heuristic we can use to enable it in most cases it would be useful?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -116,42 +128,31 @@ static void runAll(DataSet ds,
...
-            try
-            {
-                Files.delete(testDirectory);
+            // Always delete the work directory (never the cache)
+            try {
+                Files.delete(workDir);
             } catch (DirectoryNotEmptyException e) {
-                // something broke, we're almost certainly in the middle of another exception being thrown,
-                // so if we don't swallow this one it will mask the original exception
+                // swallow to avoid masking original exception
```
jshook: This doesn't look safe. The user should at least be told that there was an unexpected condition here.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
tlwillke: I was originally planning to leave caching out of this automation method.  But I'll wire it in anyway and we can leave it disabled.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
tlwillke: Yes!  I think this is a great idea.  @MarkWolters and I had discussed the same a couple hours ago.  I'll give it a shot.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
tlwillke: Let's consider it a preview feature for now and use it for a release cycle before we change the defaults.

tlwillke:
> I think the invalidation / versioning would be well handled by maintaining a properties file that get updated as necessary with the current version from the pom file when a build is run and then can be read in by the java process. Looks good though, should be a good time saver!

Thanks.  I think including the JV revision in the key is a good idea, but I'm afraid it's insufficient since changes to main happen between releases as well.

I think we're headed toward invalidation based on:
1. Any change to the base vectors in the dataset
2. Any change to the index construction algorithm
3. Any change to the featureSet (quantization math, etc.)

Essentially, we need something like version numbers that reflect logic / math changes to indexing, quantization, etc., and cryptographic hashing to monitor dataset changes.

I'll open an issue.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
tlwillke: There doesn't seem to be a non-flaky way to get the current version before a maven build.  IDEs can bypass this.  So, I will include this in an issue to revisit.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
```diff
@@ -0,0 +1,349 @@
...
+        private CacheKey(String datasetName,
+                         Set<FeatureId> featureSet,
+                         int M,
+                         int efConstruction,
+                         float neighborOverflow,
+                         boolean addHierarchy,
+                         boolean refineFinalGraph,
+                         String compressorId) {
```
ashkrisk, line 141: Should the cache key keep track of the `alpha` parameter as well? Currently Grid doesn't expose this (it's hard-coded to `1.2f`), but it might prevent a surprise or two sometime down the line.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
```diff
@@ -0,0 +1,349 @@
...
+    public Optional<ImmutableGraphIndex> tryLoad(CacheKey key) throws IOException {
+        if (!enabled) return Optional.empty();
+        Entry e = resolve(key);
+        if (!Files.exists(e.finalPath)) return Optional.empty();
+
+        try {
+            return Optional.of(OnDiskGraphIndex.load(ReaderSupplierFactory.open(e.finalPath)));
+        } catch (RuntimeException | IOException ex) {
+            // treat as miss; leave cleanup policy to the caller (or call invalidate(key) if desired)
+            return Optional.empty();
```
ashkrisk, line 207: It might help with debugging if we log the message before swallowing the exception.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
```diff
@@ -0,0 +1,349 @@
...
+        if (overwrite == Overwrite.DENY && Files.exists(e.finalPath)) {
+            throw new IllegalStateException("Cache entry already exists: " + e.finalPath);
+        }
+
+        // Ensure any previous temp is gone
+        Files.deleteIfExists(e.tmpPath);
+
+        // If overwriting, remove the final first (best-effort)
+        if (overwrite == Overwrite.ALLOW) {
+            Files.deleteIfExists(e.finalPath);
```
ashkrisk: I've done this thing a few times where I run something that deletes the existing cache as a first step, then I change my mind as it's running and have the urge to hit Ctrl-C, but suddenly I have no cache :)

Maybe we should skip the eager deletion and keep the old cache file till the new one is ready to replace it?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
```diff
@@ -0,0 +1,349 @@
...
+         * will see this entry.
+         */
+        public void commit() throws IOException {
+            if (committed) return;
+
+            try {
+                Files.move(entry.tmpPath, entry.finalPath,
+                        StandardCopyOption.REPLACE_EXISTING,
+                        StandardCopyOption.ATOMIC_MOVE);
+            } catch (AtomicMoveNotSupportedException e) {
```
ashkrisk: I took a peek at the [Javadoc](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/nio/file/Files.html#move(java.nio.file.Path,java.nio.file.Path,java.nio.file.CopyOption...)) for `Files.move`:

> The move is performed as an atomic file system operation and all other options are ignored. If the target file exists then it is implementation specific if the existing file is replaced or this method fails by throwing an [IOException](https://docs.oracle.com/javase/8/docs/api/java/io/IOException.html).

Sounds like we might want to catch `IOException` as well, especially if we do away with the eager delete.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
tlwillke: I have improved the Exception handling and reporting.  We will now get a clear failure with diagnostic reporting.  Something like:

Cleanup Failed: Could not delete graph0 -> /var/folders/rn/w2zr1jy51t7_7bnx75dj429c0000gp/T/BenchGraphDir2594869846815234048/graph0
Cleanup Failed: Could not delete graph1 -> /var/folders/rn/w2zr1jy51t7_7bnx75dj429c0000gp/T/BenchGraphDir2594869846815234048/graph1
Exception in thread "main" java.io.IOException: Run succeeded, but cleanup failed: /var/folders/rn/w2zr1jy51t7_7bnx75dj429c0000gp/T/BenchGraphDir2594869846815234048

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
tlwillke: Good suggestion.  I've added alpha to the key.  The value of alpha is still hardcoded in Grid.  Wiring alpha into the benchmarks is beyond the scope of this PR.  But the cache is now ready for this.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
tlwillke: Good suggestion.

            // File exists but is corrupted or partial
            System.err.println("Cache Load Failed: Corrupted File=" + e.finalPath.getFileName());

It will rebuild the cache entry and continue (self-healing), but the error will be prominent.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
tlwillke: I did away with the eager delete, although most of the time it will not matter since currently invalidating a cache entry and deleting the file are one and the same.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/OnDiskGraphIndexCache.java
tlwillke: Thanks for looking into this.  I agree.  Now if an IOException is thrown on the atomic move, it will try a standard move as a fallback.  If that works, it will warn but continue the run (since some platforms simply don't support atomic moves).  If the fallback fails, it will throw a new fatal IOException.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/AutoBenchYAML.java
tlwillke: Done.  It is disabled by default.  I also wired in refineFinalGraph while I was at it.

PR #617, opened 2026-02-09
Add introductory documentation

ashkrisk:
This includes:
- A set of tutorials covering basic JVector concepts
- A guide for running benchmarks

Supersedes https://github.com/datastax/jvector/pull/597

### DIFF
--- a/docs/tutorials/2-disk-tutorial.md
+++ b/docs/tutorials/2-disk-tutorial.md
@@ -45,7 +45,7 @@ Path graphPath = Files.createTempFile("jvector-example-graph", null);  // or whe
 
 // Create a writer for the on-heap graph we just built.
 // Remember to close when done.
-GraphIndexWriter writer = GraphIndexWriter.getBuilderFor(GraphIndexWriterTypes.ON_DISK_PARALLEL, heapGraph, graphPath)
+GraphIndexWriter writer = GraphIndexWriter.getBuilderFor(GraphIndexWriterTypes.RANDOM_ACCESS_PARALLEL, heapGraph, graphPath)
     // Let the writer know that we'll also be passing in the actual vector data
     // to be saved "inline" with the data for each corresponding graph node.
     .with(new InlineVectors(dim))
--- a/docs/tutorials/3-larger-than-memory-tutorial.md
+++ b/docs/tutorials/3-larger-than-memory-tutorial.md
@@ -100,7 +100,7 @@ for (int ordinal = 0; ordinal < ravv.size(); ordinal++) {
     pqVectors.encodeAndSet(ordinal, v);
 
     // Write the feature (full-resolution vector) for a single vector instead of all at once.
-    writer.writeInline(ordinal, Feature.singleState(FeatureId.INLINE_VECTORS, new InlineVectors.State(v)));
+    writer.writeFeaturesInline(ordinal, Feature.singleState(FeatureId.INLINE_VECTORS, new InlineVectors.State(v)));
 
     builder.addGraphNode(ordinal, v);
 }
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/tutorial/DiskIntro.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/tutorial/DiskIntro.java
@@ -79,7 +79,7 @@ public class DiskIntro {
         Path graphPath = Files.createTempFile("jvector-example-graph", null);  // or wherever you want to save the graph
         try (
             // Create a writer for the on-heap graph we just built.
-            GraphIndexWriter writer = GraphIndexWriter.getBuilderFor(GraphIndexWriterTypes.ON_DISK_PARALLEL, heapGraph, graphPath)
+            GraphIndexWriter writer = GraphIndexWriter.getBuilderFor(GraphIndexWriterTypes.RANDOM_ACCESS_PARALLEL, heapGraph, graphPath)
                 // Let the writer know that we'll also be passing in the actual vector data
                 // to be saved "inline" with the data for each corresponding graph node.
                 .with(new InlineVectors(dim))
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/tutorial/LargerThanMemory.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/tutorial/LargerThanMemory.java
@@ -127,7 +127,7 @@ public class LargerThanMemory {
 
                     // Write the feature (full-resolution vector) for a single vector instead of all at once.
                     try {
-                        writer.writeInline(ordinal, Feature.singleState(FeatureId.INLINE_VECTORS, new InlineVectors.State(v)));
+                        writer.writeFeaturesInline(ordinal, Feature.singleState(FeatureId.INLINE_VECTORS, new InlineVectors.State(v)));
                     } catch (IOException e) {
                         throw new UncheckedIOException(e);
                     }


docs/tutorials/1-intro-tutorial.md
```diff
@@ -0,0 +1,149 @@
...
+int dimension = 3;
+
+// Create a `VectorFloat` from a `float[]`.
+// The types that can be converted to a VectorFloat are technically dependent on which VectorizationProvider is picked,
+// but `float[]` is generally a safe bet.
+float[] vector0Array = new float[]{0.1f, 0.2f, 0.3f};
+VectorFloat<?> vector0 = vts.createFloatVector(vector0Array);
+```
+
+> [!TIP]
```
jshook, line 22: For these types of visual callouts, like `[!TIP]` or `[!NOTE]`, you might just use the unicode versions which do not require special markdown support, like one of 💡,👉, ☞, 📝.

docs/tutorials/3-larger-than-memory-tutorial.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Loading the Dataset
+
+We'll change things up with a different dataset:
+
+```java
+// The DataSet provided by loadDataSet is in-memory,
+// but you can apply the same technique even when you don't have
+// the base vectors in-memory.
+DataSet dataset = DataSets.loadDataSet("e5-small-v2-100k").orElseThrow();
```
jshook: Unfortunately, `.orElseThrow()` often yields crypic errors. If you add the exception provider as in `.orElseThrow(() -> new RuntimeException("...` then you will contextualize any such errors as well as narrate the call site more directly.

I know this is an example, but if someone uses this as copypasta, then their error will not be helpful.

docs/tutorials/2-disk-tutorial.md
```diff
@@ -0,0 +1,107 @@
...
+ReaderSupplier readerSupplier = ReaderSupplierFactory.open(graphPath);
+OnDiskGraphIndex graph = OnDiskGraphIndex.load(readerSupplier);
+```
+
+Now we can perform searches exactly as with the in-memory index, with one minor difference: for the in-memory index, we had to keep track of the RAVV we used to build the graph in order to create an exact `SearchScoreProvider` for each query. In this case, since Inline vectors are available as a feature, we can acquire a RAVV from the index itself.
+
+```java
+GraphSearcher searcher = new GraphSearcher(graph);
+// Views of an OnDiskGraphIndex with inline or separated vectors can be used as RAVVs!
+// Create a new searcher and extract a separate RAVV for each thread
```
jshook: The connection with "threads" feels a bit tenuous here. Can we make a more obvious connection?

docs/tutorials/2-disk-tutorial.md
```diff
@@ -0,0 +1,107 @@
...
+// use the RAVV from the graph instead of the one from the original dataSet
+SearchScoreProvider ssp = DefaultSearchScoreProvider.exact(query, vsf, graphRavv);
+// A slightly more complex overload of `search` which adds three extra parameters.
+// Right now we only care about `rerankK`.
+SearchResult sr = searcher.search(ssp, topK, rerankK, 0.0f, 0.0f, Bits.ALL);
+```
+
+The code from this tutorial is available in `DiskIntro.java`. Run it from the root of this repo using
+```sh
+mvn compile exec:exec@tutorial -Dtutorial=disk -pl jvector-examples -am
```
jshook, line 105: As a side note, we may want to start using exec:java here. The trade-offs are meaningful for new users. The main one being: with exec:exec you have full process isolation and you can run anything but with exec:java you get to carry the classpath and process hierarchy with you which simplifies debugging and digging deeper. This can be a separate discussion, but wanted to plant a seed here.

docs/tutorials/3-larger-than-memory-tutorial.md
ashkrisk: Added exceptions

docs/tutorials/2-disk-tutorial.md
ashkrisk: I updated the wording like this:
```java
// Views of an OnDiskGraphIndex with inline or separated vectors can be used as RAVVs!
// In multi-threaded scenarios you should have one searcher per thread
// and extract a view for each thread from the associated searcher.
```
Is that the kind of thing you meant?

docs/tutorials/1-intro-tutorial.md
ashkrisk: The advantage with these versions is that it displays the content as a nice aside that is visually separate from the main flow:

---
Screenshot 2026-02-11 at 11 02 09 AM

---

The unicode versions will have the icon, but won't demarcate the section as well. In other markdown renderers, it shows up a a standard quotation block, which still manages to demarcate the content even though it doesn't have the icon or the coloring:

---

Screenshot 2026-02-11 at 11 05 30 AM

---

Given that most people will be reading this on Github anyway I'd prefer to leave it as is for the time being.

docs/tutorials/2-disk-tutorial.md
ashkrisk: The main problem I see with exec:java is that you can't pre-configure VM options like `--add-modules=jdk.incubator.vector` in the pom.xml. This has to be done during runtime while starting maven. For these tutorials I think the better approach is to stick with `exec:exec` since
- (a) It gets rid of the warnings about the vector API not being enabled
- (b) Users can go check the "recommended" config options in the pom.xml

docs/tutorials/1-intro-tutorial.md
```diff
@@ -0,0 +1,149 @@
...
+
+```java
+// The type of similarity score to use. JVector supports EUCLIDEAN (L2 distance), DOT_PRODUCT and COSINE.
+VectorSimilarityFunction similarityFunction = VectorSimilarityFunction.EUCLIDEAN;
+
+// A simple score provider which can compute exact similarity scores by holding a reference to all the base vectors.
+BuildScoreProvider bsp = BuildScoreProvider.randomAccessScoreProvider(ravv, similarityFunction);
+```
+
+Let's also initialize the graph parameters. For now we will not worry about the exact function of the parameters, except to note that these are reasonable defaults. Refer the DiskANN and HNSW papers for more details.
```
tlwillke: Refer to the DiskANN and HNSW papers for more details.

docs/tutorials/1-intro-tutorial.md
```diff
@@ -0,0 +1,149 @@
...
+// RAVV or `ravv` is convenient shorthand for a RandomAccessVectorValues instance
+RandomAccessVectorValues ravv = new ListRandomAccessVectorValues(baseVectors, dimension /* 3 */);
+```
+
+> [!TIP]
+> In this example, all vectors are loaded in-memory, but RAVVs are quite versatile. For example, you might have a RAVV backed by disk (check out `MMapRandomAccessVectorValues.java`) or write your own custom RAVV that transfers data over a network interface.
+
+> [!NOTE]
+> A note on terminology:
+> - "Base" vectors are the vectors used to build the index. Each vector becomes a node in the graph. May also be referred to as the "train" set.
```
tlwillke: Let's move away from using the terms train and test.  It's a misuse of the terms and not typical similarity search jargon.

docs/tutorials/1-intro-tutorial.md
```diff
@@ -0,0 +1,149 @@
...
+                                                  efConstruction,
+                                                  neighborOverflow,
+                                                  alpha,
+                                                  addHierarchy,
+                                                  refineFinalGraph);
+ImmutableGraphIndex graph = builder.build(ravv);
+```
+
+> [!NOTE]
+> You may notice that we supplied the same `ravv` to `builder.build`, even though we'd already passed in the RAVV while creating the `BuildScoreProvider`. This is necessary since generally speaking, the `BuildScoreProvider` won't keep a reference to the actual base vectors, it just so happens that we're using an "exact" score provider that does so.
```
tlwillke: ... but it just so happens that we're using an "exact" score provider that does so.

docs/benchmarking.md
```diff
@@ -0,0 +1,119 @@
...
+
+JVector supports two types of datasets:
+- **Fvec/Ivec**: The dataset consists of three files, for example `base.fvec`, `queries.fvec` and `neighbors.ivec` containing the base vectors, query vectors, and ground truth. (`fvec` and `ivec` file formats are described [here](http://corpus-texmex.irisa.fr/))
+- **HDF5**: The dataset consists of a single HDF5 file with three datasets labelled `train`, `test` and `neighbors`, representing the base vectors, query vectors and the ground truth.
+
+The general procedure for running benchmarks is mentioned below. The following sections describe the process in more detail.
+- [Specify the dataset](#specifying-datasets) names to benchmark in `datasets.yml`.
+- Certain datasets will be downloaded automatically. If using a different dataset, make sure the dataset files are downloaded and made available (refer the section on [Custom datasets](#custom-datasets)).
+- Adjust the benchmark parameters in `default.yml`. This will affect the parameters for all datasets to be benchmarked. You can specify custom parameters for a specific dataset by creating a file called `<your-dataset-name>.yml` in the same folder.
```
tlwillke, line 20: With the logging PR merged, the benchmarking parameters, console output, and logging configuration are run-level settings.  Thus, they do not reside in the dataset yaml files.

docs/benchmarking.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Setting benchmark parameters
+
+### default.yml / \<dataset-name\>.yml
+
+`jvector-examples/yaml-configs/default.yml` specifies the default index construction and search parameters to be used by `bench` for all datasets.
+
+You can specify a custom set of a parameters for any given dataset by creating a file called `<dataset-name>.yml`, with `<dataset-name>` replaced by the actual name of the dataset. This is the same as the identifier used in `datasets.yml`, but without the `.hdf5` suffix for hdf5 datasets. The format of this file is exactly the same as `default.yml`.
+
+Refer `default.yml` for a list of all options.
```
tlwillke: Refer to....

docs/tutorials/1-intro-tutorial.md
ashkrisk: done

docs/tutorials/1-intro-tutorial.md
ashkrisk: done

docs/tutorials/1-intro-tutorial.md
ashkrisk: done

docs/benchmarking.md
ashkrisk: Added a fresh point for the run yaml. All the benchmark parameters are still `default.yaml`, so I'm leaving this line unchanged.

docs/benchmarking.md
ashkrisk: done

docs/benchmarking.md
tlwillke: The term 'benchmark parameters' is somewhat misleading.  The benchmarking specs are now in run.yml.  The index construction parameters and search parameters are in the dataset yamls.  No big deal though.

PR #618, opened 2026-02-10
Remap heap graph ordinals on serialization

ashkrisk:
There is currently no way to alter the ordinals of the vectors in an `OnHeapGraphIndex`. This limitation extends to the `save` and `load` methods used for serializing and de-serializing the heap graph.

In particular, this means that any ordinal "holes" created by deleting nodes from the `OnHeapGraphIndex` can never be compacted away. This is different from the behavior of the disk graph writer, which not only provides a way to remap ordinals, it automatically compacts ordinals by default.

This PR adds the option to remap ordinals while serializing the graph, preventing the accumulation of ordinal holes.

Note that this is only useful for those systems which rely on repeatedly loading, mutating, and then saving the `OnHeapGraphIndex`, which is arguably an anti-pattern.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
@@ -27,6 +27,8 @@ package io.github.jbellis.jvector.graph;
 import io.github.jbellis.jvector.annotations.Experimental;
 import io.github.jbellis.jvector.disk.RandomAccessReader;
 import io.github.jbellis.jvector.graph.ConcurrentNeighborMap.Neighbors;
+import io.github.jbellis.jvector.graph.disk.AbstractGraphIndexWriter;
+import io.github.jbellis.jvector.graph.disk.OrdinalMapper;
 import io.github.jbellis.jvector.graph.diversity.DiversityProvider;
 import io.github.jbellis.jvector.graph.similarity.ScoreFunction;
 import io.github.jbellis.jvector.util.Accountable;
@@ -49,7 +51,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerArray;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.StampedLock;
-import java.util.function.Function;
 import java.util.stream.IntStream;
 
 /**
@@ -62,6 +63,7 @@ import java.util.stream.IntStream;
 public class OnHeapGraphIndex implements MutableGraphIndex {
     // Used for saving and loading OnHeapGraphIndex
     public static final int MAGIC = 0x75EC4012; // JVECTOR, with some imagination
+    public static final int SERIALIZE_VERSION = 4;
 
     // The current entry node for searches
     private final AtomicReference<NodeAtLevel> entryPoint;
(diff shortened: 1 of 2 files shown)


jshook:
Does this pose a risk for stable ordinal mapping with external system which retain their own ordinal maps? We need to identify how the through line from non-vector data remains connected to the associated vectors/embeddings in jvector. If this is already handled effectively here, we should make it *very* clear in documentation with examples.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -62,6 +63,7 @@
 public class OnHeapGraphIndex implements MutableGraphIndex {
     // Used for saving and loading OnHeapGraphIndex
     public static final int MAGIC = 0x75EC4012; // JVECTOR, with some imagination
+    public static final int SERIALIZE_VERSION = 4;
```
akash-shankaran, line 66: If this is JVector version, consider writing out the entire version (e.g. 4.0.0-rc4)?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -523,17 +525,43 @@ public String toString() {
...
+    @Experimental
+    @Deprecated
+    public void save(DataOutput out, OrdinalMapper mapper) throws IOException {
         if (!allMutationsCompleted()) {
             throw new IllegalStateException("Cannot save a graph with pending mutations. Call cleanup() first");
         }
 
         out.writeInt(OnHeapGraphIndex.MAGIC); // the magic number
-        out.writeInt(4); // The version
+        out.writeInt(SERIALIZE_VERSION); // The version
```
MarkWolters, line 564: Our versioning is frankly a mess, and needs to be refactored completely. That said, we are currently up to [version 6 ](https://github.com/datastax/jvector/blob/main/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java#L72), so this may need to be updated here as well.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -576,8 +604,8 @@ public static OnHeapGraphIndex load(RandomAccessReader in, int dimension, double
         }
 
         int version = in.readInt(); // The version
-        if (version != 4) {
-            throw new IOException("Unsupported version: " + version);
+        if (version != SERIALIZE_VERSION) {
```
MarkWolters, line 607: same as above

jvector-base/src/main/java/io/github/jbellis/jvector/graph/OnHeapGraphIndex.java
```diff
@@ -523,17 +525,43 @@ public String toString() {
...
 
     /**
-     * Saves the graph to the given DataOutput for reloading into memory later
+     * <p>Saves the graph to the given DataOutput for reloading into memory later.
+     *
+     * <p>
+     * Does not alter existing ordinals even if the ordinals are not compact,
+     * which can happen if some nodes were deleted.
      */
     @Experimental
```
MarkWolters, line 534: I know this is neither the point of this PR, nor was it introduced by this PR but these @Experimental and @Deprecated annotations have been here since 10/10/25. It's nonsensical to combine the 2 annotations to begin with, and based on the fact that the code in question is still in use I would argue that it is neither experimental nor deprecated at this point. Can we just get rid of them?

MarkWolters (changes requested):
I think the version needs to be updated. If it's intentionally being set to a value < the current version used elsewhere explain why and I'll approve once that's settled.

PR #623, opened 2026-02-12
Optimizes reducelanes in diversityCalculation of PQVectors, for Euclidean function

MarkWolters:
This pull request introduces improvement to Euclidean similarity function in PQVectors.diversityFunctionFor. From flamegraph, it is observed that considerable amount of time is spent in jdk/incubator/vector/FloatVector.reducelanesTemplate. This is mainly because FloatVector.reducelanes() is expensive and it is being called inside a for loop (via VectorUtil.squareL2Distance). Modification in this pull request moves call to reduceLanes() outside the for loop.

Change proposed here was tested with the benchmark, PQDistanceCalculationBenchmark.diversityCalculation
With this benchmark, ~18% reduction in time was observed when M=64 and ~22% when M=192.

Code modifications:

Added a new function pqDiversityEuclidean in VectorUtilSupport and its corresponding implementations
Removed for loop in PQVectors.diversityFunctionFor and moved it into pqDiversityEuclidean
Moved FloatVector.reducelanes() outside the for loop
Test setup:
Jvector version : main branch (as of 2025-08-28)
JDK version : openjdk version "24.0.2" 2025-07-15
Platform : INTEL(R) XEON(R) PLATINUM 8592+
Benchmark : PQDistanceCalculationBenchmark.diversityCalculation

New changes
I have modified the code to include dot product & cosine functions and implemented similar changes for scoreFunctionFor.
With the changes applied to scoreFunctionFor, when M=64: dot product shows ~30% reduction in latency & cosine shows ~43% reduction.
I can add data points for other subspace counts, if required.

Code modifications:

Added new functions pqScoreEuclidean, pqScoreDotProduct, pqScoreCosine in VectorUtilSupport and its corresponding implementations for diversityFunctionFor
Added overloaded version of above functions for scoreFunctionFor
Removed for loop in PQVectors.diversityFunctionFor and PQVectors.scoreFunctionFor and moved them into respective functions in PanamaVectorUtilSupport
Moved FloatVector.reducelanes() outside the for loop
Added a new benchmark which uses MutablePQVectors to test this
Test setup:
Jvector version : main branch (as of 2025-10-22)
JDK version : openjdk version "25.0.1" 2025-10-21
Platform : Intel(R) Xeon(R) 6979P
Benchmark : PQDistanceCalculationMutableVectorBenchmark (Added this by replicating PQDistanceCalculationBenchmark to measure performance for MutablePQVectors)

### DIFF
--- a/jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
+++ b/jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
@@ -1543,4 +1543,1025 @@ class PanamaVectorUtilSupport implements VectorUtilSupport {
     public float pqDecodedCosineSimilarity(ByteSequence<?> encoded, int clusterCount, VectorFloat<?> partialSums, VectorFloat<?> aMagnitude, float bMagnitude) {
         return pqDecodedCosineSimilarity(encoded, 0, encoded.length(),  clusterCount, partialSums, aMagnitude, bMagnitude);
     }
+
+    /*-------------------- Score functions--------------------*/
+    //adding SPECIES_64 & SPECIES_128 for completeness, will it get there?
+    float pqScoreEuclidean_512(VectorFloat<?>[] codebooks, int[][] subvectorSizesAndOffsets, ByteSequence<?> node1Chunk, int node1Offset, ByteSequence<?> node2Chunk, int node2Offset, int subspaceCount) {
+        float res = 0;
+        FloatVector sum = FloatVector.zero(FloatVector.SPECIES_PREFERRED);
+
+        for (int m = 0; m < subspaceCount; m++) {
+            int centroidIndex1 = Byte.toUnsignedInt(node1Chunk.get(m + node1Offset));
+            int centroidIndex2 = Byte.toUnsignedInt(node2Chunk.get(m + node2Offset));
+            int centroidLength = subvectorSizesAndOffsets[m][0];
+            final int vectorizedLength = FloatVector.SPECIES_PREFERRED.loopBound(centroidLength);
+            int length1 = centroidIndex1 * centroidLength;
+            int length2 = centroidIndex2 * centroidLength;
+
+            if (centroidLength == FloatVector.SPECIES_PREFERRED.length()) {
+                FloatVector a = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length1);
+                FloatVector b = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length2);
+                var diff = a.sub(b);
+                sum = diff.fma(diff, sum);
+            }
+            else {
+                int i = 0;
+                for (; i < vectorizedLength; i += FloatVector.SPECIES_PREFERRED.length()) {
+                    FloatVector a = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length1 + i);
+                    FloatVector b = fromVectorFloat(FloatVector.SPECIES_PREFERRED, codebooks[m], length2 + i);
+                    var diff = a.sub(b);
+                    sum = diff.fma(diff, sum);
+                }
+                // Process the tail
+                for (; i < centroidLength ; ++i) {
+                    var diff = codebooks[m].get(length1 + i) - codebooks[m].get(length2 + i);
+                    res += MathUtil.square(diff);
+                }
+            }
+        }
+        res += sum.reduceLanes(VectorOperators.ADD);
+        return res;
+    }
+
+    float pqScoreEuclidean_256(VectorFloat<?>[] codebooks, int[][] subvectorSizesAndOffsets, ByteSequence<?> node1Chunk, int node1Offset, ByteSequence<?> node2Chunk, int node2Offset, int subspaceCount) {
+        float res = 0;
+        FloatVector sum = FloatVector.zero(FloatVector.SPECIES_256);
+
+        for (int m = 0; m < subspaceCount; m++) {
+            int centroidIndex1 = Byte.toUnsignedInt(node1Chunk.get(m + node1Offset));
+            int centroidIndex2 = Byte.toUnsignedInt(node2Chunk.get(m + node2Offset));
+            int centroidLength = subvectorSizesAndOffsets[m][0];
+            final int vectorizedLength = FloatVector.SPECIES_256.loopBound(centroidLength);
+            int length1 = centroidIndex1 * centroidLength;
+            int length2 = centroidIndex2 * centroidLength;
+
+            if (centroidLength == FloatVector.SPECIES_256.length()) {
+                FloatVector a = fromVectorFloat(FloatVector.SPECIES_256, codebooks[m], length1);
+                FloatVector b = fromVectorFloat(FloatVector.SPECIES_256, codebooks[m], length2);
+                var diff = a.sub(b);
+                sum = diff.fma(diff, sum);
+            }
+            else {
+                int i = 0;
+                for (; i < vectorizedLength; i += FloatVector.SPECIES_256.length()) {
+                    FloatVector a = fromVectorFloat(FloatVector.SPECIES_256, codebooks[m], length1 + i);
+                    FloatVector b = fromVectorFloat(FloatVector.SPECIES_256, codebooks[m], length2 + i);
+                    var diff = a.sub(b);
+                    sum = diff.fma(diff, sum);
+                }
+                // Process the tail
+                for (; i < centroidLength ; ++i) {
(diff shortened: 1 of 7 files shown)


jshook (approved):
This looks reasonable to me. It would be nice to see the results laid out a little better for before and after, but it still looks like a solid improvement.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -1543,4 +1543,1025 @@ public void calculatePartialSums(VectorFloat<?> codebook, int codebookIndex, int
     public float pqDecodedCosineSimilarity(ByteSequence<?> encoded, int clusterCount, VectorFloat<?> partialSums, VectorFloat<?> aMagnitude, float bMagnitude) {
         return pqDecodedCosineSimilarity(encoded, 0, encoded.length(),  clusterCount, partialSums, aMagnitude, bMagnitude);
     }
+
+    /*-------------------- Score functions--------------------*/
+    //adding SPECIES_64 & SPECIES_128 for completeness, will it get there?
+    float pqScoreEuclidean_512(VectorFloat<?>[] codebooks, int[][] subvectorSizesAndOffsets, ByteSequence<?> node1Chunk, int node1Offset, ByteSequence<?> node2Chunk, int node2Offset, int subspaceCount) {
```
r-devulap, line 1549: This name is mis-leading, probably better to rename this to `pqScoreEuclidean_preferred`?

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -1543,4 +1543,1025 @@ public void calculatePartialSums(VectorFloat<?> codebook, int codebookIndex, int
...
+
+        for (int m = 0; m < subspaceCount; m++) {
+            int centroidIndex1 = Byte.toUnsignedInt(node1Chunk.get(m + node1Offset));
+            int centroidIndex2 = Byte.toUnsignedInt(node2Chunk.get(m + node2Offset));
+            int centroidLength = subvectorSizesAndOffsets[m][0];
+            final int vectorizedLength = FloatVector.SPECIES_PREFERRED.loopBound(centroidLength);
+            int length1 = centroidIndex1 * centroidLength;
+            int length2 = centroidIndex2 * centroidLength;
+
+            if (centroidLength == FloatVector.SPECIES_PREFERRED.length()) {
```
r-devulap, line 1561: Why do we need this special case? It introduces a lot of duplicated code across all the functions. Wouldn’t the loop below already handle this scenario correctly?

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -1543,4 +1543,1025 @@ public void calculatePartialSums(VectorFloat<?> codebook, int codebookIndex, int
...
+        if(subvectorSizesAndOffsets[0][0] >= FloatVector.SPECIES_PREFERRED.length()) {
+            return pqScoreEuclidean_512(codebooks, subvectorSizesAndOffsets, node1Chunk, node1Offset, node2Chunk, node2Offset, subspaceCount);
+        }
+        else if(subvectorSizesAndOffsets[0][0] >= FloatVector.SPECIES_256.length()) {
+            return pqScoreEuclidean_256( codebooks,  subvectorSizesAndOffsets, node1Chunk, node1Offset,  node2Chunk, node2Offset, subspaceCount);
+        }
+        else if (subvectorSizesAndOffsets[0][0]  >=  FloatVector.SPECIES_128.length()) {
+            return pqScoreEuclidean_128( codebooks,  subvectorSizesAndOffsets, node1Chunk, node1Offset,  node2Chunk, node2Offset, subspaceCount);
+        }
+        return pqScoreEuclidean_64( codebooks,  subvectorSizesAndOffsets, node1Chunk, node1Offset,  node2Chunk, node2Offset, subspaceCount);
```
r-devulap, line 1709: We currently have 4 separate implementations (`pqScoreEuclidean_64/_128/_256/_512`) that are structurally identical aside from the `FloatVector.SPECIES_*` used. This is a lot of duplicated hot-path code and increases the risk of fixes/optimizations diverging across variants.

Could we fold these into a single generic-ish implementation that takes the species as a parameter, e.g.:

- `pqScoreEuclideanImpl(VectorSpecies<Float> species, ...)`
- `pqScoreEuclidean(...)` selects `species` (`SPECIES_PREFERRED`, `SPECIES_256`, `SPECIES_128`, `SPECIES_64`) once and dispatches to the shared impl.

This keeps the vectorized logic in one place and avoids the confusing `_512` naming given it actually uses `SPECIES_PREFERRED` on some paths. It should also make it much easier to keep Euclidean/Dot/Cosine implementations consistent.

r-devulap (commented):
First‑pass review: there’s quite a bit of duplicated code, and I think we can consolidate it significantly.

shyla226:
> First‑pass review: there’s quite a bit of duplicated code, and I think we can consolidate it significantly.

@r-devulap, Thank you very much for your feedback. I agree there is quite a bit of duplication of code, but I had tried creating a generic function, providing `VectorSpecies` as an argument.
For cosine function, something like this
 `float pqScoreCosine_generic(final VectorSpecies<Float> SPEC, VectorFloat<?>[] codebooks, int[][] subvectorSizesAndOffsets, ByteSequence<?> encodedChunk, int encodedOffset, VectorFloat<?> centeredQuery, int subspaceCount)`

What I found was that, if the `SPECIES` is not set at compile-time, we don't see the performance benefit.

I tried it again after I saw your feedback, I see the same behavior as I did before unfortunately! 
I am open to other thoughts you have.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
shyla226: I have renamed the functions, code available [here](https://github.com/datastax/jvector/commit/6bbf42f706c6e9d32f2af96e610de4d2dac14537)

I rearranged the code a little bit for better clarity.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
shyla226: I did it this way, so we don't process the tail when it is not needed, similar to `dotPreferred `& `dotProductPreferred `functions in the file. Not processing the tail seemed to yield some performance benefit in my previous experiments. 
Let me do a few more tests and get back to you on this one

r-devulap:
@shyla226 Quick question regarding this patch.

It updates `diversityFunctionFor` in the abstract `PQVectors` class, but I’m not convinced the benchmark `PQDistanceCalculationBenchmark.diversityCalculation` would show any improvement from this change 
as you claimed.

In the benchmark setup, `pqVectors` is populated via `pq.encodeAll(ravv)`, which produces an `ImmutablePQVectors`. As a result, the call to `diversityFunctionFor` is dispatched to `ImmutablePQVectors.diversityFunctionFor` — the overridden implementation that uses precomputed codebook partial sums — rather than the base implementation in `PQVectors` that this patch modifies. You can confirm this in the attached flame graph: the class method `MutablePQVectors.diversityFunctionFor` never appears in the benchmark. 

flame-graph-benchmark

That code path is only exercised when the instance is a `MutablePQVectors`, which does *not* override the method. The new microbenchmarks you added for `MutablePQVectors` may indeed show improvements, since they actually hit the modified code.

Would you be able to show whether this change yields improvements in a high‑level index‑construction benchmark rather than only the microbenchmarks?. My understanding is that `MutablePQVectors` is used during incremental (larger‑than‑memory) index construction, but I don't believe we currently have benchmarks covering that. @jshook or @MarkWolters can likely confirm.

For the standard in‑memory (on‑heap) index build, the code path` ImmutablePQVectors.diversityFunctionFor` isn’t on the hot path, and I'm not sure whether incremental index construction would behave differently.

r-devulap:
@shyla226 could you generate a flame graph for the workload and post it here? Curious to see where the hotspots are and if the reduce add shows up in there.

r-devulap:
@shyla226 thanks for sharing those. I am not familiar with these workloads yet, please allow sometime for me to verify these benchmarks.

r-devulap:
@shyla226 and I connected over email to discuss this. I attempted to reproduce the NoSQLBench results, but my concern is that the results are noisy and do not show an obvious benefit from this patch. I also suggested running BenchYAML to demonstrate the performance impact of the change and recommended testing on 1M-scale datasets. Closing this for now. Please feel free to reopen the issue if you can provide BenchYAML performance results that clearly demonstrate a measurable improvement.

PR #625, opened 2026-02-13
Introduce run.yml policy and run-scoped experiment logging for Grid benchmarks

tlwillke:
## Summary

This PR introduces YAML-driven benchmark compute + reporting selection, and adds run-scoped logging artifacts (`sys_info.json`, `dataset_info.csv`, `experiments.csv`) so Grid benchmark runs are reproducible and easier to analyze offline. Reporting/benchmark policy is now run-level (via `yaml-configs/run.yml`), while dataset YAML files are simplified to dataset-tuned construction/search settings only.

---

### Run-level policy file: `yaml-configs/run.yml`

- Added a new run-level config file that defines:
  - **benchmark computation policy** (`benchmarks`: what to compute)
  - **console projection** (`console`: subset + named metrics)
  - **experiments logging policy** (`logging`: selection + sink type + run metadata)
- Dataset YAMLs are now focused on **dataset-tuned parameters only** (construction/search grids). Reporting/benchmark controls are no longer expected in dataset YAMLs.

### Config versioning clarity

- Renamed the old vague YAML `version` field to **`onDiskIndexVersion`** (still validated against `OnDiskGraphIndex.CURRENT_VERSION`).
- Introduced **`yamlSchemaVersion`** (starting at **1**) to version YAML schema independently of on-disk index header format.
- Added deprecated back-compat for legacy files that still use `version: 6`.

### Compute vs display vs logging separation

- Benchmarks specify **what is computed**; console/logging are **projections** only.
- Added stable `Metric.key` identifiers for all benchmark outputs and telemetry so selection and CSV columns are robust (no header-string matching).
- Enforced:
  - **strict subset validation** for benchmark-stat selections (must be computed)
  - **best-effort warnings** for runtime-unavailable telemetry/metrics (omitted from output/CSV, with warnings)

### Run-scoped artifacts and logging

- When logging is enabled, each BenchYAML invocation creates a run directory under:
  - `logging/<run_id>/`
- The run directory contains (see examples below):
  - `sys_info.json` — host/OS/JVM/SIMD/threading/memory + stable `system_id`
  - `dataset_info.csv` — one row per dataset used in the run
  - `experiments.csv` — flat, append-only table of fixed params + selected output-key columns
- `experiments.csv` schema stability:
  - fixed columns are defined centrally in `ExperimentsSchemaV1`
  - output-key columns are an ordered union across all topK scenarios encountered in the run
  - non-applicable / missing values are written as **empty CSV fields** (easy for pandas/matplotlib)

### Plumbing / ergonomics

- Added `RunArtifacts` to centralize:
  - run setup + validation
  - dataset registration (`dataset_info.csv`)
  - row logging (`experiments.csv`)
- Updated BenchYAML and HelloVectorWorld to use `run.yml` + `RunArtifacts` (no manual wiring of compute/display/logging maps).
- Refactored Grid signatures to take a single `RunArtifacts` object instead of many reporting parameters.
- Legacy callers continue to work via the legacy Grid overload + `RunArtifacts.disabled()`.

### Legacy compatibility

- Legacy YAML files **without** `yamlSchemaVersion` are treated as **schema 0**:
  - parsed leniently (unknown fields ignored)
  - emit a **deprecation warning**
  - optional extraction of legacy `search.benchmarks` is supported for legacy behavior when `run.yml` is absent
- AutoBenchYAML is not migrated to run.yml yet; it continues to work with legacy configs.

---

## Behavior

- **Logging enabled** (`run.yml` has `logging.type`):
  - BenchYAML writes run artifacts under `logging/<run_id>/`
  - prints a startup message pointing to the run directory and how to delete it to reclaim space
  - appends experiment rows as configurations execute
- **Logging disabled** (no `logging` section or blank `type` in `run.yml`):
  - run artifacts are not created
  - `RunArtifacts` becomes a no-op via `RunArtifacts.disabled()`
- Console output continues to work via run-level selection; dataset configs tune only construction/search parameters.

---

## Notes / limitations

- `experiments.csv` uses `Metric.key` strings as column names for benchmark outputs/telemetry; missing values are empty CSV fields.
- `sys_info.json` SIMD reports the configured vector width via `io.github.jbellis.jvector.vector_bit_size` (with `simd_config_present`), avoiding brittle runtime probing.
- Threading in `sys_info.json` distinguishes build executor parallelism vs parallel streams (common FJP), explaining why build/query values can differ.
- This PR intentionally does not overhaul per-index build time attribution (e.g., cache hits have no build time by design); deeper build-timing correctness is left for a future pass.
- Dataset YAML breaking changes are intentional: `version` → `onDiskIndexVersion`, plus removal of dataset-level benchmarks/console/logging fields in favor of `run.yml`.

## Example of run.yml and yaml-config for dataset
ada002-100k yml
run yml

## Examples of artifacts (sys_info, dataset_info, experiments)

experiments csv
dataset_info csv
sys_info json

### DIFF
--- a/jvector-examples/pom.xml
+++ b/jvector-examples/pom.xml
@@ -120,6 +120,11 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.15.1</version>
+        </dependency>
+        <dependency>
             <groupId>com.kohlschutter.junixsocket</groupId>
             <artifactId>junixsocket-core</artifactId>
             <version>2.8.1</version>
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/RunContext.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/RunContext.java
@@ -24,7 +24,7 @@ import java.util.UUID;
 /**
  * Immutable run-scoped context for logging artifacts.
  *
- * - runId: human-friendly directory name (e.g., "2026-02-09_14-23-08")
+ * - runId: human-friendly directory name (rendered from logging.runId template; e.g., "20260209-222308Z" or "my-exp_20260209-222308Z")
  * - runUuid: collision-proof identity for the run
  * - runDir: logDir/runId
  * - schemaVersion: version of the logging artifact schema
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/RunIdUtil.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/RunIdUtil.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.jbellis.jvector.example.reporting;
+
+import org.apache.commons.io.FilenameUtils;
+import java.time.Clock;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+
+/**
+ * Utility for generating a per-run directory name ("runId") for logging artifacts.
+ *
+ * <p>The runId is rendered from a user-supplied template containing the placeholder {@code "{ts}"},
+ * which is replaced with a UTC timestamp formatted as {@code yyyyMMdd-HHmmssZ} (e.g. {@code 20260213-232254Z}).
+ * If the template is blank or missing, {@code "{ts}"} is used. If the template does not include {@code "{ts}"},
+ * {@code _{ts}} is appended to ensure uniqueness.</p>
+ *
+ * <p>The rendered value is then sanitized to be a single, cross-platform-safe path component:
+ * path traversal is removed, only {@code [A-Za-z0-9._-]} is permitted (others become {@code _}),
+ * trailing dots/spaces are stripped (Windows), and Windows reserved device names are prefixed.</p>
+ */
+public final class RunIdUtil {
+    private static final DateTimeFormatter TS_UTC =
+            DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss'Z'").withZone(ZoneOffset.UTC);
+
+    public static String renderRunId(String template, Clock clock) {
+        String t = (template == null || template.isBlank()) ? "{ts}" : template.trim();
+        if (!t.contains("{ts}")) t = t + "_" + "{ts}";
+
+        String ts = TS_UTC.format(clock.instant());
+        String rendered = t.replace("{ts}", ts);
+
+        // Prevent directory climbing
+        String nameOnly = FilenameUtils.getName(rendered.replace('\\', '/'));
+
+        // Allow only alphanumeric, dots, hyphens, underscores
+        // This covers Linux, Windows, and Mac
+        String sanitized = nameOnly.replaceAll("[^A-Za-z0-9._-]", "_")
+                .replaceAll("_{2,}", "_")
+                .replaceAll("[. ]+$", ""); // Windows: no trailing dot/space
+
+        if (sanitized.isEmpty()) sanitized = "run_" + ts;
+
+        // Windows reserved device names (case-insensitive)
+        String base = sanitized;
+        int dot = base.indexOf('.');
+        if (dot >= 0) base = base.substring(0, dot);
+        String upper = base.toUpperCase(java.util.Locale.ROOT);
+
+        if (upper.equals("CON") || upper.equals("PRN") || upper.equals("AUX") || upper.equals("NUL")
+                || upper.matches("COM[1-9]") || upper.matches("LPT[1-9]")) {
+            sanitized = "run_" + sanitized;
+        }
+
+        return sanitized;
+    }
+}
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/RunReporting.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/RunReporting.java
@@ -19,11 +19,11 @@ package io.github.jbellis.jvector.example.reporting;
 import io.github.jbellis.jvector.example.yaml.RunConfig;
 
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Clock;
 import java.time.Instant;
-import java.time.LocalDateTime;
-import java.time.format.DateTimeFormatter;
 import java.util.UUID;
 
 /**
@@ -34,7 +34,6 @@ import java.util.UUID;
  * {@link RunContext} for downstream writers (dataset_info.csv, experiments.csv).
  */
 public final class RunReporting {
-    private static final DateTimeFormatter RUN_ID_FMT = DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss");
 
     private final RunContext run;
     private final String jvectorRef;
@@ -53,26 +52,30 @@ public final class RunReporting {
     }
 
     public static RunReporting open(RunConfig runCfg) throws IOException {
+        // Resolve base logging directory (default "logging")
         String logDir = (runCfg != null && runCfg.logging != null && runCfg.logging.dir != null && !runCfg.logging.dir.isBlank())
-                ? runCfg.logging.dir
+                ? runCfg.logging.dir.trim()
                 : "logging";
 
-        String runId = (runCfg != null && runCfg.logging != null && runCfg.logging.runId != null && !runCfg.logging.runId.isBlank())
-                ? runCfg.logging.runId
-                : LocalDateTime.now().format(RUN_ID_FMT);
+        // Resolve runId from template using UTC clock (default "{ts}" inside RunIdUtil)
+        Clock clock = Clock.systemUTC();
+        String runIdTemplate = (runCfg != null && runCfg.logging != null) ? runCfg.logging.runId : null;
+        String runId = RunIdUtil.renderRunId(runIdTemplate, clock);
 
+        // Run metadata
         String jvectorRef = (runCfg != null && runCfg.logging != null && runCfg.logging.jvectorRef != null)
                 ? runCfg.logging.jvectorRef
                 : "";
 
         UUID runUuid = UUID.randomUUID();
-        Instant createdAt = Instant.now();
+        Instant createdAt = clock.instant();
+
         Path runDir = Paths.get(logDir).resolve(runId);
+        Files.createDirectories(runDir);
 
         Integer buildThreads = detectBuildThreads();
         Integer queryThreads = java.util.concurrent.ForkJoinPool.getCommonPoolParallelism();
 
-        // Threads: fill later if you want; safe to start null
         String systemId = SysInfoWriter.writeSysInfo(
                 runDir,
                 RunContext.SCHEMA_VERSION,
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/yaml/RunConfig.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/yaml/RunConfig.java
@@ -70,7 +70,7 @@ public class RunConfig {
     /** Run-level logging settings + selection for experiments.csv. */
     public static class RunLogging extends BenchmarkSelection {
         public String dir;       // e.g. "logging"
-        public String runId;     // optional override
+        public String runId;     // optional template; supports "{ts}" (UTC). Default: "{ts}"
         public String jvectorRef; // tag/commit to record
         public String type;      // e.g. "csv"
     }
--- a/jvector-examples/yaml-configs/run.yml
+++ b/jvector-examples/yaml-configs/run.yml
@@ -25,7 +25,7 @@ console:
 logging:
 #  dir: logging # set as desired
 #  jvectorRef: "no ref"   # tag or git sha (recommended)
-#  runId: # optional override; otherwise timestamp
+#  runId: "my-experiment-{ts}"   # optional; template with {ts} (UTC timestamp). Default: "{ts}"
 #  type: csv # only csv is presently supported
 #  benchmarks:
 #    throughput: [ AVG ] # additional options: [AVG, MEDIAN, MAX]


jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/CsvLogger.java
```diff
@@ -0,0 +1,4 @@
+package io.github.jbellis.jvector.example.util;
+
+public class CsvLogger {
```
ashkrisk: unused?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ExecutionTimeBenchmark.java
```diff
@@ -75,6 +75,7 @@ public List<Metric> runBenchmark(
         }
 
         double avgRuntimeSec = totalRuntime / queryRuns / 1e9;
-        return List.of(Metric.of("Avg Runtime (s)", format, avgRuntimeSec));
+        return List.of(Metric.of("search.execution_time.avg_runtime_s",
```
ashkrisk, line 78: It looks like this class is not wired up. Should we remove the file altogether? Execution time is (1 / QPS) anyway.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
```diff
@@ -253,16 +253,37 @@ public List<Metric> runBenchmark(
...
+                    "Avg QPS (of " + numTestRuns + ")",
+                    formatAvgQps,
+                    avgQps));
+
+            list.add(Metric.of("search.throughput.stddev_qps",
+                    "± Std Dev",
+                    formatAvgQps,
+                    stdDevQps));
+
+            list.add(Metric.of("search.throughput.cv_pct",
```
ashkrisk, line 266: Can we change this to `cv_pc_qps` to make it a little more obvious what this is?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/yaml/MultiConfig.java
```diff
@@ -17,20 +17,38 @@
...
     public ConstructionParameters construction;
     public SearchParameters search;
+    public String dataset;
+
+    private static final String defaultDirectory = "jvector-examples/yaml-configs/";
+    private static final java.util.regex.Pattern YAML_SCHEMA_VERSION_KEY =
+            java.util.regex.Pattern.compile("(?m)^\\s*yamlSchemaVersion\\s*:");
+
+    private int yamlSchemaVersion;
+    private int onDiskIndexVersion;
```
ashkrisk, line 45: Do we need the field `onDiskIndexVersion`? It has a getter and setter but these aren't referenced anywhere. It doesn't seem to affect any functionality either, including logging / reporting.

Similarly for the eponymous field in `RunConfig`.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ExecutionTimeBenchmark.java
tlwillke: Yes, it has been unused but is unrelated / orthogonal to this PR.  Please open an issue.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/CsvLogger.java
tlwillke: I believe so.  I was experimenting with logging early last year.  This looks to be leftover.  Not part of this PR.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/yaml/MultiConfig.java
tlwillke: Good catch.  I should have wired in the check.  The legacy version does indeed check the on-disk index version against the yaml config to ensure that the parameters are revisited when the index is revised.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/CsvLogger.java
tlwillke: Done.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
tlwillke: All of the keys are a bit cryptic (not sure if `cv_pc_qps` is clearer).  Luckily, they are not user-facing and are easily understood in context.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/Metric.java
```diff
@@ -21,32 +21,35 @@
 
 /**
  * A single column in the table:
+ *   - key: Stable identifier for selection/logging (e.g. "search.throughput.avg_qps")
  *   - header: The column title
  *   - fmtSpec: The format specifier, e.g. ".3f" for floats or "s" for strings
  *   - extractor: How to pull the value from the summary‐map
  */
 public class Metric {
+    private final String key;
```
jshook, line 30: We should try to move to dimensional labeling as a follow-on improvment. The hierarchic naming works ok, but breaks down for some key usage patterns which I think are useful to us. (Selecting by aspect, relying on orthogonality where we can)

I'll put in a card to make a proposal around this. Thankfully dimensional labeling can naturally desctructure into a linearized naming scheme for compatibility.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
```diff
@@ -253,16 +253,37 @@ public List<Metric> runBenchmark(
...
+                    "CV %",
+                    ".1f",
+                    coefficientOfVariation));
         }
         if (computeMedianQps) {
```
jshook, line 271: In general, if there is no harm in computing these things, they should just always be computed. The conditional presence of key metrics creates more downstream confusion than it solves in upstream logic. The presence of these in degenerate cases even allows for useful cross-checking of other outputs.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/DatasetInfoWriter.java
```diff
@@ -0,0 +1,189 @@
+/*
```
jshook, line 1: There are other "per access" issues with accessing datasets which this change won't cover. We need a proper DatasetInfo handle type which is realized before consumption of actual data. I have a working model for how to handle this, but it will be a follow-on card.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/ExperimentsCsvWriter.java
```diff
@@ -0,0 +1,109 @@
+/*
```
jshook, line 1: We may want to pivot this over to JSONL. It has much more structural (hierarchic even) support for data layout, has well-defined grammar, and is universally supported, perhaps even better than CSV.
Our other tools will allow us to read the results meaninfully, as in we shouldn't be scanning through CSV in general for key measurements or analysis.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/JsonUtil.java
```diff
@@ -0,0 +1,94 @@
...
+ * limitations under the License.
+ */
+
+package io.github.jbellis.jvector.example.reporting;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+public final class JsonUtil {
```
jshook, line 29: Shouldn't we be using Gson or similar here? It's quite a complete kit and handles all the SERDES we need.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/ReportingSelectionResolver.java
```diff
@@ -0,0 +1,248 @@
...
+ *  - selection: sink selection (benchmarks subset + named metrics)
+ *  - catalog: mapping from YAML names -> Metric.key templates/keys
+ *  - ctx: context for template substitution (e.g., topK)
+ *
+ * Rules:
+ *  - benchmark stat selection must be a subset of computed stats (fail fast)
+ *  - named metric selection must use recognized names (fail fast)
+ *  - runtime availability is best effort: caller can warn+omit missing produced keys
+ */
+public final class ReportingSelectionResolver {
```
jshook, line 51: There is a lot going on here in terms of behavior and wiring. We should probably have a bit more of an explanatory section in the javadoc or elsewhere showing the data flow and purpose of this _nexus_ logic.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
```diff
@@ -94,7 +94,7 @@ private static void execute(Pattern pattern, boolean enableIndexCache, List<Func
             DataSet ds = DataSets.loadDataSet(datasetName).orElseThrow(
                     () -> new RuntimeException("Dataset " + datasetName + " not found")
             );
-            Grid.runAll(ds, enableIndexCache, mGrid, efConstructionGrid, neighborOverflowGrid, addHierarchyGrid, refineFinalGraphGrid, featureSets, buildCompression, compressionGrid, topKGrid, usePruningGrid, null);
```
jshook, line 97: The fact that this method signature keeps growing is a design smell that we need to address soon. I've seen the same happen in other PRs. We probably need to add a formalism or pattern around defining test scopes and features which aligns better to how users see the system.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/yaml/MultiConfig.java
tlwillke: Everything is wired.  The `onDiskIndexVersion` is now checked in both the legacy and strict new yaml schema.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/ThroughputBenchmark.java
tlwillke: Unfortunately there is harm.  Some of these computations add significant time to the run.   Benchmark computations have been selectable in the past, so not changing this paradigm with this PR.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/Metric.java
tlwillke: TIA for the proposal / issue.  We'll scope it outside of this PR.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/DatasetInfoWriter.java
tlwillke: TIA for opening an issue.  Will resolve outside the scope of this PR.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/ExperimentsCsvWriter.java
tlwillke: The logging design is extensible and purposely architected to allow for multiple formats, including csv, json, parquet, etc.  Welcoming future extensions, but outside the scope of this PR.  Offering csv now to satisfy the large number of analysis and plotting tools that support this format.  A new issue is welcome.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/ReportingSelectionResolver.java
tlwillke: Good catch.  I will add some helpful explanations.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/JsonUtil.java
tlwillke: Thanks for the suggestion.  Done.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Bench.java
tlwillke: Good point.  Originally, the logging design passed a number of parameters.  I got that down to 1.  Now the list is dominated by other parameters that should be addressed in other work.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/AccuracyBenchmark.java
```diff
@@ -99,21 +99,18 @@ public List<Metric> runBenchmark(
 
         var list = new ArrayList<Metric>();
         if (computeRecall) {
-            // compute recall for this run
             double recall = AccuracyMetrics.recallFromSearchResults(
                     cs.getDataSet().getGroundTruth(), results, topK, topK
             );
-            list.add(Metric.of("Recall@" + topK, formatRecall, recall));
+            list.add(Metric.of("search.accuracy.recall_at_" + topK, "Recall@" + topK, formatRecall, recall));
```
MarkWolters, line 105: These changes will necessitate changes as well to BenchmarkSummarizer in the `extract...Metric` calls where it uses this key to recall the data or it will break the regression test suite

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/CountBenchmark.java
```diff
@@ -125,14 +125,14 @@ public List<Metric> runBenchmark(
 
         var list = new ArrayList<Metric>();
         if (computeAvgNodesVisited) {
-            list.add(Metric.of("Avg Visited", formatAvgNodesVisited, avgVisited));
+            list.add(Metric.of("search.count.avg_visited", "Avg Visited", formatAvgNodesVisited, avgVisited));
```
MarkWolters, line 128: As above this necessitates changes to BenchmarkSummarizer to avoid breaking regression testing

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/LatencyBenchmark.java
```diff
@@ -143,17 +143,15 @@ public List<Metric> runBenchmark(
 
         var list = new ArrayList<Metric>();
         if (computeAvgLatency) {
-            list.add(Metric.of("Mean Latency (ms)", formatAvgLatency, mean));
+            list.add(Metric.of("search.latency.mean_ms", "Mean Latency (ms)", formatAvgLatency, mean));
```
MarkWolters, line 146: necessitates changes to BenchmarkSummarizer

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/AccuracyBenchmark.java
MarkWolters: scratch that, the header isn't changing and key is a new value, ignore this comment

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/ExperimentsCsvWriter.java
```diff
@@ -0,0 +1,109 @@
...
+    private final RunContext run;
+    private final Path out;
+    private final List<String> fixedColumns;
+    private final List<String> outputKeyColumns;
+
+    public ExperimentsCsvWriter(RunContext run,
+                                List<String> fixedColumns,
+                                List<String> outputKeyColumns) throws IOException {
+        this.run = run;
+        this.out = run.runDir().resolve("experiments.csv");
```
MarkWolters, line 45: Nit: can we make the name configurable?

tlwillke: Absolutely.

tlwillke: Thanks for discussing offline.  The directory name is now templated.  The files within follow a standard naming convention.  Please review the latest commit.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/ReportingSelectionResolver.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    // -----------------------------
+    // Validation (pre-build)
+    // -----------------------------
+
+    /** If compute spec is null/empty, treat as defaultCompute. */
+    public static Map<String, List<String>> effectiveComputeSpec(Map<String, List<String>> compute,
+                                                                 Map<String, List<String>> defaultCompute) {
+        return (compute == null || compute.isEmpty()) ? defaultCompute : compute;
```
MarkWolters, line 129: As defaultCompute is an arg here it could potentially also be null or empty

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/SearchReportingCatalog.java
```diff
@@ -0,0 +1,98 @@
...
+        return new ReportingSelectionResolver.Catalog(
+                benchmarkKeyTemplates(),
+                namedMetricKeys(),
+                "benchmarks",  // sink-neutral selector prefix
+                "metrics"      // sink-neutral selector prefix
+        );
+    }
+
+    private static Map<String, Map<String, List<String>>> benchmarkKeyTemplates() {
+        return Map.of(
```
MarkWolters, line 56: There are a few different places in our code now that need to make use of the available benchmarks and their categories. For a future enhancement it make make sense to implement a registry of these to avoid having to hardcode these in multiple locations.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/BenchYAML.java
```diff
@@ -57,21 +58,16 @@ public static void main(String[] args) throws IOException {
...
-                        rawname.endsWith(hdf5) ? rawname.substring(0, rawname.length() - hdf5.length()) : rawname;
-                // pre-loading and early error phase
-                // TODO: There are some housekeeping stages which get invoked as part of "loading" which should be deferred to avoid double processing
-                DataSets.loadDataSet(datasetName).orElseThrow(
-                        () -> new RuntimeException("Could not load dataset:" + datasetName)
-                );
-
-                MultiConfig config = MultiConfig.getDefaultConfig(datasetName);
-                allConfigs.add(config);
+            System.out.println("Executing the following datasets: " +
```
MarkWolters, line 61: Do we still want to allow the use of `println` with the logging facilities available?

MarkWolters (approved):
Left a few comments with my thoughts but I don't see anything that would be a show stopper

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/ReportingSelectionResolver.java
tlwillke: Good point.  I put a guard in place.  By design, it should never be null, but you never know.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/SearchReportingCatalog.java
tlwillke: Couldn't agree more!  We do need a registry.  Good topic for an issue.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/BenchYAML.java
tlwillke: This logging is not system logging per say.  It will not capture such messages since it is focused on recording experimental results instead of system health.  So, I think this is ok.

Issue #629, opened 2026-02-16
Missing OS level XSAVE check in AVX-512 runtime checks


r-devulap:
Current checks do not validate XSAVE support at runtime. On VMs where 512‑bit register state saving is disabled, this can result in #UD faults when executing AVX‑512 instructions.

PR #630, opened 2026-02-16
Replace __cpuid_check for AVX512 with compiler builtin cpu features

r-devulap:
Fixes https://github.com/datastax/jvector/issues/629. 

This pull request simplifies AVX-512 feature detection logic in the `jvector_simd_check.c` file by replacing manual CPUID bit checks with built-in compiler functions (available only on gcc and llvm compilers). The earlier code was missing XSAVE support at runtime. On VMs where 512‑bit register state saving is disabled, this can result in #UD faults when executing AVX‑512 instructions.

CPU feature detection improvements:

* Replaced manual CPUID register checks for AVX-512 features with calls to `__builtin_cpu_supports` for each required feature, improving code readability and portability.
* Added a call to `__builtin_cpu_init()` to ensure proper initialization before feature checks.
* Rename `check_compatibility` to `check_avx512_compatibility`.

### DIFF
--- a/jvector-native/src/test/java/io/github/jbellis/jvector/vector/cnative/NativeSimdOpsTest.java
+++ b/jvector-native/src/test/java/io/github/jbellis/jvector/vector/cnative/NativeSimdOpsTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


r-devulap:
> Looks like only GCC is supported at the moment. Refer [jextract_vector_simd.h](https://github.com/datastax/jvector/blob/f0c52d4017be6c0c5a55dc6c6b9e36b52b55809a/jvector-native/src/main/c/jextract_vector_simd.sh).

We seem to be testing on windows CI where the native AVX512 module is not built. Do we want it to work on Windows or is it not important enough to worry about.

r-devulap:
> If it is determined by the library, it would be good to include the cpuId detection for ARM, as many of us use mac's for development and we can take advantage of advanced vector instructions. The latter may be out of scope for this PR.

The vector code written using Panama API should already generate the appropriate vector instructions on an ARM CPU. The JVM handles both the runtime checks and JIT generation of NEON vector code.  

AFAIK, JVector does not have any native ARM code and hence doesn't require any CPUID detection for ARM.

r-devulap:
> when is the Panama API vs Jvector native module used, and what is it based on?

AFAIK, the routines in the native AVX‑512 module can’t be written using the Panama Vector API (I still need to determine the exact limitations—work in progress). If that turns out to be true, then yes, we’ll need to implement the ARM equivalents

PR #636, opened 2026-02-19
Add quantization timing metrics to YAML benchmarking

tlwillke:
## Summary

This PR adds quantization timing telemetry to BenchYAML reporting (console + experiments.csv), and makes it selectable from `run.yml`. The new metrics capture quantizer **compute** and **encoding** time for both **index construction** and **search-time vector encoding**, with stable `Metric.key` identifiers.

### Quantization timing metrics (new)
- Emit quant timing metrics as `Metric` outputs:
  - **Index construction:** `construction.index_quant_time_s.<QT>.(compute_time_s|encoding_time_s)`
  - **Search-time:** `search.search_quant_time_s.<QT>.(compute_time_s|encoding_time_s)`
- Per-quantizer keys are emitted for PQ/BQ; NVQ reports **compute only** (NVQ encoding is intentionally not measured).
- Human-readable headers are phase-tagged to disambiguate:
  - `Idx PQ compute (s)`, `Idx PQ encoding (s)`, `Search PQ compute (s)`, etc.

### Run-level selection support (prefix expansion)
- Added prefix-style named metric selection in `run.yml`:
  - `metrics.construction.index_quant_time_s` → expands to all `construction.index_quant_time_s.*` metrics
  - `metrics.construction.search_quant_time_s` → expands to all `search.search_quant_time_s.*` metrics
- Selection + application now supports “prefix families” of metrics (similar to `recall_at_{topK}` expansion).

### Logging fixes (experiments.csv)
- Updated logging schema planning to include prefix-selected metrics so quant timing columns appear in `experiments.csv` when selected via run.yml.
- CSV column keys remain `Metric.key` strings; missing/non-applicable values remain empty fields.

### Console ergonomics
- Added 2-line header rendering with a max column width cap so wide metric names wrap cleanly.
- Manual header breaks are supported via `\n` in the `Metric.header` string (useful for specific columns like Avg QPS).

### Runtime warnings (availability)
- Preserved “warn once” behavior for missing exact-key metrics (e.g., index build time on cache hit).
- Quant timing warnings were adjusted so we do not warn about NVQ encoding (not measured by design).

## Notes / limitations
- PQ compute time is recorded when the PQ compressor is computed (cache miss). On cache hits, compute time is intentionally blank (no compute performed).
- NVQ encoding time is intentionally not measured; NVQ reports compute only.
- Prefix selection is used for quant timing because concrete per-quantizer keys are only known once the run produces metrics.
quantization_metrics_logging

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
@@ -69,6 +69,7 @@ import java.util.*;
 import java.util.function.Function;
 import java.util.function.IntFunction;
 import java.util.stream.IntStream;
+import java.util.function.Supplier;
 
 /**
  * Tests a grid of configurations against a dataset
@@ -962,44 +963,9 @@ public class Grid {
         return results;
     }
 
+    /** Overload for non-reporting use */
     private static VectorCompressor<?> getCompressor(Function<DataSet, CompressorParameters> cpSupplier, DataSet ds) {
-        var cp = cpSupplier.apply(ds);
-        if (!cp.supportsCaching()) {
-            return cp.computeCompressor(ds);
-        }
-
-        var fname = cp.idStringFor(ds);
-        return cachedCompressors.computeIfAbsent(fname, __ -> {
-            var path = Paths.get(pqCacheDir).resolve(fname);
-            if (path.toFile().exists()) {
-                try {
-                    try (var readerSupplier = ReaderSupplierFactory.open(path)) {
-                        try (var rar = readerSupplier.get()) {
-                            var pq = ProductQuantization.load(rar);
-                            System.out.format("%s: %s codebooks loaded from %s%n", ds.getName(), pq, fname);
-                            return pq;
-                        }
-                    }
-                } catch (IOException e) {
-                    throw new UncheckedIOException(e);
-                }
-            }
-
-            var start = System.nanoTime();
-            var compressor = cp.computeCompressor(ds);
-            System.out.format("%s: %s codebooks computed in %.2fs,%n", ds.getName(), compressor, (System.nanoTime() - start) / 1_000_000_000.0);
-            if (cp.supportsCaching()) {
-                try {
-                    Files.createDirectories(path.getParent());
-                    try (var writer = new BufferedRandomAccessWriter(path)) {
-                        compressor.write(writer, OnDiskGraphIndex.CURRENT_VERSION);
-                    }
-                } catch (IOException e) {
-                    throw new UncheckedIOException(e);
-                }
-            }
-            return compressor;
-        });
+        return getCompressor(cpSupplier, ds, null, null, null);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/BenchmarkTablePrinter.java
```diff
@@ -72,33 +92,55 @@ private void initializeHeader(List<Metric> cols) {
...
+            System.out.printf("  %-20s %s%n", e.getKey(), String.valueOf(e.getValue()));
+        }
     }
 
     private void printHeader(List<Metric> cols) {
-        // Prepare array: First "Overquery", then each Metric header
-        Object[] hdrs = new Object[cols.size() + 1];
-        hdrs[0] = "Overquery";
+        // Two header lines: split long headers onto a second line when possible
+        Object[] hdrs1 = new Object[cols.size() + 1];
```
jshook, line 117: Why not make this a String?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/BenchmarkTablePrinter.java
```diff
@@ -109,7 +151,7 @@ private void printHeader(List<Metric> cols) {
      */
     public void printRow(double overquery,
                          List<Metric> cols) {
-        initializeHeader(cols);
+        initializeHeader(cols); // lazy: prints header on the first row after resetTable()
 
         // Build argument array: First overquery, then each Metric.extract(...)
         Object[] vals = new Object[cols.size() + 1];
```
jshook, line 157: Same as above. What's the reason for not making this a String?

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -953,6 +1002,79 @@ private static VectorCompressor<?> getCompressor(Function<DataSet, CompressorPar
...
+     * <p>We do <b>not</b> record load times. If a cached compressor is loaded, no compute metric is emitted
+     * (value remains {@code null} and will be blank in CSV/console).</p>
+     *
+     * @param cpSupplier supplies dataset-specific compressor parameters (YAML-derived)
+     * @param ds dataset being benchmarked
+     * @param metrics per-run construction metrics sink (may be null)
+     * @param phase which phase to attribute compute time to (index construction vs search-time setup)
+     * @param quantType YAML quantization type label (e.g., PQ/BQ); if null, no quant metric is recorded
+     */
+     private static VectorCompressor<?> getCompressor(Function<DataSet, CompressorParameters> cpSupplier,
```
MarkWolters: There is significant duplication in this getCompressor method with the previously existing getCompressor method. While not a showstopper for this PR it should be addressed.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -1000,4 +1122,122 @@ public void close() throws Exception {
             searchers.close();
         }
     }
+
+    // Helper that maps YAML-derived CompressorParameters to a stable label
+    private static String quantTypeOf(CompressorParameters cp) {
+        if (cp == null || cp == CompressorParameters.NONE) return null;
+        if (cp instanceof CompressorParameters.PQParameters) return "PQ";
+        if (cp instanceof CompressorParameters.BQParameters) return "BQ";
+        // Unknown/unsupported type for quant timing purposes
```
MarkWolters, line 1110: Nit: Print error message informing user that an unsupported type was passed

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/BenchmarkTablePrinter.java
tlwillke: Good question.  It could be done that way.  I kept them raw types to allow the template to handle rounding and padding in a single pass.  It's really ok either way.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/BenchmarkTablePrinter.java
tlwillke: See other comment.  Same answer.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
tlwillke: Thanks for pointing that out.  Will add.

tlwillke: Turns out this is already caught by Compression.  No changes needed.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
tlwillke: Good point.  I refactored to eliminate the overlap and created a couple of helpers to bring clarity.

MarkWolters (approved):

PR #637, opened 2026-02-20
add DataSetInfo and some guards around loading errors

jshook:
This PR makes dataset loading more robust in the face of file corruption errors.
It also indirects dataset access through DataSetInfo, caching the attached DataSet locally, and enabling access via getDataSet(). This will vastly speed up testing flows which are uneccesssarily processing datasets into memory right when the actual data is not yet accessed.

[UPDATE]

Due to a request for fixing some longer-standing issues, the scope of this PR has increased moderately:

* DataSet loader formats which can not provide metadata (VSF, ...) on their own now pull details from dataset_metadata.yml for VSF. If none such is available, an error should be thrown.

* Addtionally, the contract type (DataSetProperties) is now a layer of requirements added onto the DataSetInfo type, including an indicator for dataset preprocessing aspects (zero vectors, dupes) and normalization.

* surefire tests were disabled in examples. enabled, and fixed up a unit test

* DataSetProperties is a carrier type in the new DataSetInfo contract

After this change, any dataset which is loaded where the VSF can't be resolved from some definitive source will cause an error to be thrown, as it should.

### DIFF
--- a/jvector-examples/yaml-configs/datasets.yml
+++ b/jvector-examples/yaml-configs/datasets.yml
@@ -1,3 +1,4 @@
+neighborhood-watch-100k:
   - cohere-english-v3-100k
   - ada002-100k
   - openai-v3-small-100k
@@ -7,8 +8,10 @@
   - e5-small-v2-100k
   - e5-base-v2-100k
   - e5-large-v2-100k
+neighborhood-watch-1M:
   - ada002-1M
   - colbert-1M
+ann-benchmarks:
   - glove-25-angular.hdf5
   - glove-50-angular.hdf5
   - lastfm-64-dot.hdf5


jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetInfo.java
```diff
@@ -0,0 +1,109 @@
...
+    ///
+    /// This is always available without triggering a data load.
+    public String getName() {
+        return name;
+    }
+
+    /// Returns the similarity function for this dataset.
+    ///
+    /// This is always available without triggering a data load.
+    /// For MFD datasets this is always {@link VectorSimilarityFunction#COSINE};
```
tlwillke: By default, the VSF used should be the VSF that was used to generate the ground truth data stored with the dataset.  It should not be hardwired to COSINE.  So, we do need to store this information at the dataset level and retrieve it.  If the dataset is normalized, l2, cosine, and the dot product should result in the same recall.  However, the l2 and dot product are typically computationally faster for obvious reasons.  How do you want to address this?

By the way, the similarity function is dataset-specific for the HDF5 loader (albeit limited to COSINE and EUCLIDEAN (L2)) and hardwired to COSINE for the MFD loader.  This should be corrected.  The user should be able to easily override the defaults for experimental purposes.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderHDF5.java
tlwillke:
```suggestion
```

tlwillke: Would move the download print so that it isn't output unless the file is found and ready for download.  Right now, it displays for HDF5 even if the file is not found and the MFD loader is used.

tlwillke, line 179:
```suggestion
Files.copy(in, dsFilePath, StandardCopyOption.REPLACE_EXISTING);
            System.out.println("Downloading: " + url);
```

Here's a good place for the download notification.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetInfo.java
jshook: I'll add a metadata companion file for these simpler formats which don't have direct metadata support, with a loader error should a specific VSF not be found for the dataset by name.

jshook: A central properties file and a side-path property loader has been added for formats which can't provide suitable metadata on their own. This has been pre-seeded with the correct VSF

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderHDF5.java
jshook: done

PR #638, opened 2026-02-25
add memory and disk stats to regression tests

MarkWolters:
This PR adds memory and disk stats to the regression tests run by autoBenchYAML as well as to the analysis and charts generated afterwards. Specifically the 3 new metrics collected for comparison are:

- max heap during index build phase
- max heap during query phase
- total disk size of monitored directories

This PR also adds the exact (to 2 decimal places) value of each point on the Y axis to make it easier to read and interpret the results.


tlwillke (approved):
Looks good.

jshook (approved):
No obvious issues, looks good

PR #647, opened 2026-03-19
Improve performance and readability of recall and precision measurements

tlwillke:
## Summary

This PR optimizes `AccuracyMetrics` to improve the performance and readability of recall and precision measurements. The primary focus is shifting from $O(N^2)$ list-scanning to $O(N)$ Set-based lookups and reducing object allocation overhead by interacting directly with primitive `NodeScore` arrays. 

### Logic cleanup and "dead code" removal
- **Unused Signature Removal:** Removed the redundant `topKCorrect(List, List, ...)` signature, consolidating logic into a single private method that operates directly on `SearchResult`.
- **Branch Elimination:** Removed the unreachable `if (gtView.size() > retrieved.size())` branch. Due to existing guard clauses (`kGT <= kRetrieved` and `kRetrieved <= retrieved.size()`), this condition was mathematically impossible.
- **Parity Preservation:** Preserved specific, argument-focused exception messaging to ensure clear feedback for invalid `k` parameters.

### Performance optimizations
- **Algorithmic Shift:** Replaced $O(N)$ `List.contains()` calls with $O(1)$ `HashSet.contains()`. This moves the core intersection logic from quadratic to linear time complexity.
- **AP Nested Scan Removal:** In `averagePrecisionAtK`, replaced the $O(i)$ `subList(0, i).contains(p)` duplicate check with a `HashSet` lookup. This transforms the AP calculation from an $O(K^2)$ operation to $O(K)$.
- **Allocation Reduction:** Eliminated intermediate boxing and `ArrayList` creation in the `SearchResult` path. The logic now iterates directly over the `NodeScore[]` array, significantly reducing GC pressure during large benchmark runs.
- **Loop Efficiency:** Replaced Stream API calls with manual `for` loops in high-frequency paths to avoid the object overhead and "setup tax" of the Stream API.
- **HashSet Pre-Sizing:** Sets are now initialized with explicit capacities ($K / 0.75$) to prevent expensive internal rehashing.

### Test coverage
- **Functional Validation:** Verified Recall, AP, and MAP calculations against hand-calculated results to ensure mathematical parity with standard IR definitions.
- **Duplicate Handling:** Confirmed that duplicate IDs in search results are handled via the "seen" Set, preventing artificial score inflation while maintaining $O(1)$ lookup speed.
- **Exception Parity:** Explicitly verified that `IllegalArgumentException` strings remain 1:1 identical to the original implementation to prevent breaking downstream error-parsing.

## Performance results

Benchmarks were conducted using 10,000 queries on the `gecko-100k` dataset (768 dimensions). 

| Metric | Original Time | Optimized Time | Delta |
| :--- | :--- | :--- | :--- |
| **recall@10** | **~8ms**| ~same | - |
| **recall@100** | **~58ms** | **~25ms** | **~57% reduction** |
| **AP@100 / MAP@100** | **~62ms** | **~26ms** | **~58% reduction** |

While the benefit is negligible for $K=10$, the optimization becomes critical as $K$ increases. The reduction in AP measurement time is particularly significant as it eliminates the $O(K^2)$ complexity previously caused by nested sublist scans.

## Notes / limitations
- **Autoboxing:** While this PR eliminates intermediate `List` allocations, it still utilizes `HashSet<Integer>`. Further gains could be achieved using primitive-specific collections (like `IntHashSet`) if necessary in the future.

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/AccuracyMetrics.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/AccuracyMetrics.java
@@ -65,13 +65,20 @@ public class AccuracyMetrics {
         // Load factor is 0.75, so sized to kGT / 0.75.
         Set<Integer> gtSet = new HashSet<>((int) (kGT / 0.75f) + 1);
         for (int i = 0; i < kGT; i++) {
-            gtSet.add(gt.get(i));
+            int ord = gt.get(i);
+            if (!gtSet.add(ord)) {
+                throw new IllegalArgumentException("Duplicate ground truth ordinal in top-" + kGT + ": " + ord);
+            }
         }
 
-        // Manual primitive loop for speed (no Stream setup).
+        Set<Integer> seenRetrieved = new HashSet<>((int) (kRetrieved / 0.75f) + 1);
         int hits = 0;
         for (int i = 0; i < kRetrieved; i++) {
-            if (gtSet.contains(nodes[i].node)) {
+            int p = nodes[i].node;
+            if (!seenRetrieved.add(p)) {
+                throw new IllegalArgumentException("Duplicate retrieved ordinal in top-" + kRetrieved + ": " + p);
+            }
+            if (gtSet.contains(p)) {
                 hits++;
             }
         }
@@ -99,17 +106,21 @@ public class AccuracyMetrics {
         // Sized hashset used for performance.
         Set<Integer> gtSet = new HashSet<>((int) (k / 0.75f) + 1);
         for (int i = 0; i < k; i++) {
-            gtSet.add(gt.get(i));
+            int ord = gt.get(i);
+            if (!gtSet.add(ord)) {
+                throw new IllegalArgumentException("Duplicate ground truth ordinal in top-" + k + ": " + ord);
+            }
         }
 
-        // Handles potential duplicates in O(1).
-        Set<Integer> seen = new HashSet<>((int) (k / 0.75f) + 1);
-
+        Set<Integer> seenRetrieved = new HashSet<>((int) (k / 0.75f) + 1);
         double score = 0.;
         int hits = 0;
         for (int i = 0; i < k; i++) {
             int p = nodes[i].node;
-            if (gtSet.contains(p) && seen.add(p)) {
+            if (!seenRetrieved.add(p)) {
+                throw new IllegalArgumentException("Duplicate retrieved ordinal in top-" + k + ": " + p);
+            }
+            if (gtSet.contains(p)) {
                 hits++;
                 score += (double) hits / (i + 1);
             }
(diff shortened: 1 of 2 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/AccuracyMetrics.java
```diff
@@ -41,43 +39,44 @@ public static double recallFromSearchResults(List<? extends List<Integer>> gt, L
...
             throw new IllegalArgumentException("Insufficient ground truth for the number of retrieved elements");
         }
-        Long correctCount = IntStream.range(0, gt.size())
-                .mapToObj(i -> topKCorrect(gt.get(i), retrieved.get(i), kGT, kRetrieved))
-                .reduce(0L, Long::sum);
+
+        long correctCount = 0;
+        for (int i = 0; i < gt.size(); i++) {
+            correctCount += topKCorrect(gt.get(i), retrieved.get(i), kGT, kRetrieved);
+        }
```
ashkrisk: If your goal is to avoid boxing, `mapToLong` would also work
```java
long correctCount = IntStream.range(0, gt.size())
        .mapToLong(i -> topKCorrect(gt.get(i), retrieved.get(i), kGT, kRetrieved))
        .sum();
```

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/AccuracyMetrics.java
```diff
@@ -41,43 +39,44 @@ public static double recallFromSearchResults(List<? extends List<Integer>> gt, L
...
-        var temp = Arrays.stream(retrieved.getNodes()).mapToInt(nodeScore -> nodeScore.node)
-                .boxed()
-                .collect(Collectors.toList());
-        return topKCorrect(gt, temp, kGT, kRetrieved);
-    }
+        // Manual primitive loop for speed (no Stream setup).
+        int hits = 0;
+        for (int i = 0; i < kRetrieved; i++) {
+            if (gtSet.contains(nodes[i].node)) {
+                hits++;
```
ashkrisk: Wondering if we should handle dupes in `retrieved` like we do in case of `averagePrecisionAtK`. That said, the earlier version of this function didn't handle duplicates either.

ashkrisk (approved):

jvector-examples/src/main/java/io/github/jbellis/jvector/example/util/AccuracyMetrics.java
tlwillke: Duplicates are now considered an Exception.  They are now treated the same way in both methods.

jshook (approved):
looks ok, glad to see the numerical testing

MarkWolters (approved):

PR #650, opened 2026-04-01
Add new JMH benchmarks for FusedPQ and graph index queries

r-devulap:
This PR adds two new JMH benchmark classes to evaluate vector‑search performance across both fused PQ based indices and in‑memory graph indices. These benchmarks were useful when combined with async‑profiler, helping narrow down performance hotspots and guide further optimizations. 

Included Benchmarks:

`FusedPQQueryBenchmark `– Builds a fused PQ disk index and measures query performance across configurable dimensions, PQ parameters, and search settings.

`QueryTimeBenchmark `– Benchmarks per‑query latency on an in‑memory graph index (optionally PQ‑compressed), supporting multiple dimensions and search configurations.

### DIFF
--- a/benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/FusedPQQueryBenchmark.java
+++ b/benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/FusedPQQueryBenchmark.java
@@ -1,0 +1,204 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/FusedPQQueryBenchmark.java
```diff
@@ -0,0 +1,189 @@
+
```
MarkWolters, line 16: Needs copyright statement to avoid test failures

jshook (approved):
Seems fine to me

r-devulap:
Don't need these anymore. Will close them.

PR #651, opened 2026-04-02
Wire calculatePartialSums to native SIMD via Panama FFI downcall

r-devulap:
This change uses a native implementation of `calculatePartialSums` to accelerate PQ query scoring.  
On `ada002-100k` with FUSED_PQ (numPQsubspaces/M =96, JDK build 23.0.1+11-39), it delivers **2–3× higher QPS** and **40–65% lower mean latency** across common overquery settings.  Index build time, disk usage, and heap usage show no meaningful regression.  The optimization is isolated to the PQ path; non‑PQ queries are unaffected.

### topK = 10

| Overquery | QPS (main) | QPS (native) | Speedup | Latency ms (main) | Latency ms (native) | Latency ↓ |
|----------|------------|--------------|---------|-------------------|---------------------|-----------|
| 1× | 8,987 | **26,101** | **2.9×** | 0.462 | **0.171** | **−63%** |
| 2× | 8,361 | **21,706** | **2.6×** | 0.505 | **0.202** | **−60%** |
| 5× | 7,199 | **14,364** | **2.0×** | 0.590 | **0.292** | **−51%** |
| 10× | 5,796 | **9,640** | **1.7×** | 0.731 | **0.431** | **−41%** |

### topK = 100

| Overquery | QPS (main) | QPS (native) | Speedup | Latency ms (main) | Latency ms (native) | Latency ↓ |
|----------|------------|--------------|---------|-------------------|---------------------|-----------|
| 1× | 5,743 | **9,568** | **1.7×** | 0.736 | **0.439** | **−40%** |
| 2× | 4,174 | **5,574** | **1.3×** | 1.022 | **0.733** | **−28%** |
``

Summary of changes in this PR: 

- Wire `calculatePartialSums` in NativeVectorUtilSupport to a new Panama FFI downcall for the native `calculate_partial_sums_f32_512` SIMD implementation.
- Replace icelake-server gcc target with skylake-avx512 in build script (icelake-server isnt required to build our native code)
- Remove global mutable state: eliminate initialIndexRegister, indexIncrement, maskSeventhBit, maskEighthBit globals and their constructor initializer; move mask constants (maskSeventhBit, maskEighthBit) to local scope inside lookup_partial_sums
- Add shared reduce_add_128_ps and reduce_add_256_ps helper functions using proper horizontal-add sequences instead of store-to-array loops
- Remove redundant if (length >= N) guards in all SIMD kernels — the loop body already handles the zero-iteration case correctly
- Replace store-to-aligned-array horizontal reduction pattern with the new helpers across all 128- and 256-bit dot product and euclidean distance functions
- Remove preferred_size parameter from dot_product_f32 and euclidean_f32; always dispatch to AVX-512 when length >= 16
- Standardize inline annotations: replace __attribute__((always_inline)) inline with JV_FINLINE / JV_INLINE macros throughout

### DIFF
--- a/jvector-native/src/main/c/jextract_vector_simd.sh
+++ b/jvector-native/src/main/c/jextract_vector_simd.sh
@@ -49,7 +49,7 @@ CURRENT_GCC_VERSION=$(gcc -dumpversion)
 # Check if the current GCC version is greater than or equal to the minimum required version
 if [ "$(printf '%s\n' "$MIN_GCC_VERSION" "$CURRENT_GCC_VERSION" | sort -V | head -n1)" = "$MIN_GCC_VERSION" ]; then
     rm -rf ../resources/libjvector.so
-    gcc -fPIC -O3 -march=icelake-server -c jvector_simd.c -o jvector_simd.o
+    gcc -fPIC -O3 -march=skylake-avx512 -c jvector_simd.c -o jvector_simd.o
     gcc -fPIC -O3 -march=x86-64 -c jvector_simd_check.c -o jvector_simd_check.o
     gcc -shared -o ../resources/libjvector.so jvector_simd_check.o jvector_simd.o
 
--- a/jvector-native/src/main/c/jvector_simd.c
+++ b/jvector-native/src/main/c/jvector_simd.c
@@ -259,7 +259,7 @@ JV_INLINE float euclidean_f32(const float* a, int aoffset, const float* b, int b
            : euclidean_f32_256(a, aoffset, b, boffset, length);
 }
 
-JV_INLINE void calculate_partial_sums_dot_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, float* partialSums) {
+void calculate_partial_sums_dot_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, float* partialSums) {
     int codebookBase = codebookIndex * clusterCount;
     float tempdat[16];
     if (size == 2) {
@@ -365,7 +365,7 @@ JV_INLINE void calculate_partial_sums_dot_f32_512(const float* codebook, int cod
     }
 }
 
-JV_INLINE void calculate_partial_sums_euclidean_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, float* partialSums) {
+void calculate_partial_sums_euclidean_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, float* partialSums) {
     int codebookBase = codebookIndex * clusterCount;
     float tempdat[16];
     if (size == 2) {
@@ -728,15 +728,4 @@ float pq_decoded_cosine_similarity_f32_512(const unsigned char* baseOffsets, int
     return sumResult / sqrtf(aMagnitudeResult * bMagnitude);
 }
 
-void calculate_partial_sums_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, int similarityFunction, float* partialSums) {
-    switch (similarityFunction) {
-        case 0:
-            calculate_partial_sums_euclidean_f32_512(codebook, codebookIndex, size, clusterCount, query, queryOffset, partialSums);
-            break;
-        case 1:
-            calculate_partial_sums_dot_f32_512(codebook, codebookIndex, size, clusterCount, query, queryOffset, partialSums);
-            break;
-        default:
-            break;
-    }
-}
+
--- a/jvector-native/src/main/c/jvector_simd.c
+++ b/jvector-native/src/main/c/jvector_simd.c
@@ -19,23 +19,27 @@
 #include <math.h>
 #include "jvector_simd.h"
 
-__m512i initialIndexRegister;
-__m512i indexIncrement;
-__m512i maskSeventhBit;
-__m512i maskEighthBit;
 
-__attribute__((constructor))
-void initialize_constants() {
-    if (check_avx512_compatibility()) {
-        initialIndexRegister = _mm512_setr_epi32(-16, -15, -14, -13, -12, -11, -10, -9,
-                                             -8, -7, -6, -5, -4, -3, -2, -1);
-        indexIncrement = _mm512_set1_epi32(16);
-        maskSeventhBit = _mm512_set1_epi16(0x0040);
-        maskEighthBit = _mm512_set1_epi16(0x0080);
-    }
+JV_FINLINE float reduce_add_256_ps(__m256 v) {
+    __m128 lo = _mm256_castps256_ps128(v);
+    __m128 hi = _mm256_extractf128_ps(v, 1);
+    __m128 sum128 = _mm_add_ps(lo, hi);
+    __m128 shuf = _mm_movehdup_ps(sum128);
+    __m128 sums = _mm_add_ps(sum128, shuf);
+    shuf = _mm_movehl_ps(shuf, sums);
+    sums = _mm_add_ss(sums, shuf);
+    return _mm_cvtss_f32(sums);
 }
 
-float dot_product_f32_64(const float* a, int aoffset, const float* b, int boffset) {
+JV_FINLINE float reduce_add_128_ps(__m128 v) {
+    __m128 shuf = _mm_movehdup_ps(v);
+    __m128 sums = _mm_add_ps(v, shuf);
+    shuf = _mm_movehl_ps(shuf, sums);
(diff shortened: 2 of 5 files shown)


jshook (commented):
I would like to see much more coverage of these with numerical tests. Are there some already which aren't seen here?

jvector-native/src/main/c/jextract_vector_simd.sh
```diff
@@ -49,7 +49,7 @@ CURRENT_GCC_VERSION=$(gcc -dumpversion)
 # Check if the current GCC version is greater than or equal to the minimum required version
 if [ "$(printf '%s\n' "$MIN_GCC_VERSION" "$CURRENT_GCC_VERSION" | sort -V | head -n1)" = "$MIN_GCC_VERSION" ]; then
     rm -rf ../resources/libjvector.so
-    gcc -fPIC -O3 -march=icelake-server -c jvector_simd.c -o jvector_simd.o
+    gcc -fPIC -O3 -march=skylake-avx512 -c jvector_simd.c -o jvector_simd.o
```
ashkrisk, line 52: Is there a strong reason to lower the target micro-architecture version?

jvector-native/src/main/c/jvector_simd.c
```diff
@@ -533,15 +631,112 @@ void calculate_partial_sums_best_dot_f32_512(const float* codebook, int codebook
...
+
+    return sumResult / sqrtf(aMagnitudeResult * bMagnitude);
+}
+
+void calculate_partial_sums_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, int similarityFunction, float* partialSums) {
+    switch (similarityFunction) {
+        case 0:
+            calculate_partial_sums_euclidean_f32_512(codebook, codebookIndex, size, clusterCount, query, queryOffset, partialSums);
+            break;
+        case 1:
```
ashkrisk: Can we use public enums here? Jextract should automatically make the enums available to the Java code as constants. Alternatively we could skip the parameter-based dispatch altogether and simply expose both versions of the function to Java code.

jvector-native/src/main/c/jvector_simd.c
```diff
@@ -533,15 +631,112 @@ void calculate_partial_sums_best_dot_f32_512(const float* codebook, int codebook
...
+
+    return res;
+}
+
+float pq_decoded_cosine_similarity_f32_512(const unsigned char* baseOffsets, int baseOffsetsOffset, int baseOffsetsLength, int clusterCount, const float* partialSums, const float* aMagnitude, float bMagnitude) {
+    __m512 sum = _mm512_setzero_ps();
+    __m512 vaMagnitude = _mm512_setzero_ps();
+    int i = 0;
+    int limit = baseOffsetsLength - (baseOffsetsLength % 16);
+    const __m512i initialIndexRegister = _mm512_setr_epi32(-16, -15, -14, -13, -12, -11, -10, -9, -8, -7, -6, -5, -4, -3, -2, -1);
```
ashkrisk, line 690: It's good that this isn't a global variable anymore, but given that it's used in multiple places does it make sense to have it as a global constant?

jvector-native/src/main/java/io/github/jbellis/jvector/vector/cnative/NativeSimdOps.java
ashkrisk: Looks like a lot of functions that are no longer in the public header are still declared here. Should fix itself on re-running jextract.

ashkrisk (commented):
Looks like an excellent set of optimizations. Left a few comments.

+1 to @jshook's comment about numerical tests. This PR touches almost every single function in the native supporting library, and it would be good to have a set of tests accompanying it, perhaps also in C.

r-devulap:
> I would like to see much more coverage of these with numerical tests. Are there some already which aren't seen here?

@jshook @ashkrisk 
Apologies for the delayed response. For the primary changes introduced in this patch, there is already solid test coverage exercising all the modified scenarios. These are covered in the existing tests here:
https://github.com/datastax/jvector/blob/46bd1158277e090600ad830e8e899e3e047d1f3d/jvector-tests/src/test/java/io/github/jbellis/jvector/quantization/TestCompressedVectors.java#L131-L169

Beyond that, there are only two additional native functions exposed to Java via FFI—`assemble_and_sum_f32_512` and `pq_decoded_cosine_similarity_f32_512`—and neither of these is affected by this change. The remaining modifications are mechanical in nature (e.g., adding static and inline qualifiers) and do not alter behavior.

jvector-native/src/main/c/jextract_vector_simd.sh
r-devulap: This change addresses an actual bug in our build configuration. We currently compile targeting icelake-server, but at runtime we only check for skylake-avx512. This mismatch allows the compiler to emit Ice Lake–specific instructions that may be executed on a Skylake CPU, which can result in a SIGILL.

Targeting skylake-avx512 resolves this issue and is sufficient for the kernels we currently have; there’s no requirement for Ice Lake–specific features here.

jvector-native/src/main/c/jvector_simd.c
r-devulap: I generally try to avoid global variables and prefer function‑local const values to keep dependencies explicit and contained, unless avoiding globals would cause significant duplication. In this case, it’s only used in two places, so the duplication is minimal.

jvector-native/src/main/java/io/github/jbellis/jvector/vector/cnative/NativeSimdOps.java
r-devulap: Good catch — these should indeed be removed. It looks like we check in NativeSimdOps.java with signatures that are supposed to mirror the public header. That also explains why I keep seeing a locally modified file every time I build this branch.

r-devulap:
> +1 to @jshook's comment about numerical tests. This PR touches almost every single function in the native supporting library, and it would be good to have a set of tests accompanying it, perhaps also in C.

I think this is covered by my earlier comment, but to clarify: only three native functions are exposed to Java, and this patch modifies just one of them. That function already has strong numerical test coverage, which was actually helpful in catching bugs in an earlier version of the code.

jvector-native/src/main/c/jvector_simd.c
r-devulap: > Alternatively we could skip the parameter-based dispatch altogether and simply expose both versions of the function to Java code.

Agree. I have updated to use this approach.

jvector-native/src/main/java/io/github/jbellis/jvector/vector/cnative/NativeSimdOps.java
r-devulap: Fixed it.

jvector-native/src/main/c/jvector_simd.c
```diff
@@ -520,11 +618,11 @@ void bulk_quantized_shuffle_cosine_f32_512(const unsigned char* shuffles, int co
 }
 
 // Partial sum calculations that also record best distances, as this is necessary for Fused ADC quantization
-void calculate_partial_sums_best_dot_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, float* partialSums, float* partialBestDistances) {
+JV_INLINE void calculate_partial_sums_best_dot_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, float* partialSums, float* partialBestDistances) {
```
ashkrisk, line 621: What's the difference between these `_best_` functions and their not-best counterparts? The `_best_` functions no longer look to be wired up, should they be removed altogether in favour of your new implementation?

jvector-native/src/main/c/jvector_simd.c
```diff
@@ -278,103 +248,229 @@ float euclidean_f32_512(const float* a, int aoffset, const float* b, int boffset
...
-        // first, we want to initialize a vector with the lane number added as an index
-        indexRegister = _mm512_add_epi32(indexRegister, indexIncrement);
-        // then we want to multiply by dataBase
-        __m512i scale = _mm512_mullo_epi32(indexRegister, dataBaseVec);
-        // then we want to add the base offsets
-        __m512i convOffsets = _mm512_add_epi32(scale, baseOffsetsInt);
-
-        __m512 partials = _mm512_i32gather_ps(convOffsets, data, 4);
-        sum = _mm512_add_ps(sum, partials);
+void calculate_partial_sums_dot_f32_512(const float* codebook, int codebookIndex, int size, int clusterCount, const float* query, int queryOffset, float* partialSums) {
```
ashkrisk, line 262: Nit: You've created a section in this file for APIs exposed to Java, but this function is not in that section

r-devulap:
Closing this since it’s addressed by #668.

PR #652, opened 2026-04-02
adding regression test datasets

MarkWolters:
regression tests are not producing information. Even though the files are correctly downloaded, without specifying similarity metadata the "download" fails. This PR adds the datasets used by the regression tests to the metadata file.


jshook (approved):
LGTM

tlwillke:
Hi @MarkWolters, this was addressed in https://github.com/datastax/jvector/pull/654, along with a bunch of other dataset configuration settings.  You should be all set now.

PR #655, opened 2026-04-06
Replace __cpuid_check for AVX512 with compiler builtin cpu features

### LINKED ISSUE #629: Missing OS level XSAVE check in AVX-512 runtime checks

r-devulap:
Fixes https://github.com/datastax/jvector/issues/629.

This pull request simplifies AVX-512 feature detection logic in the jvector_simd_check.c file by replacing manual CPUID bit checks with built-in compiler functions (available only on gcc and llvm compilers). The earlier code was missing XSAVE support at runtime. On VMs where 512‑bit register state saving is disabled, this can result in #UD faults when executing AVX‑512 instructions.

CPU feature detection improvements:

1. Replaced manual CPUID register checks for AVX-512 features with calls to __builtin_cpu_supports for each required feature, improving code readability and portability.
  Added a call to __builtin_cpu_init() to ensure proper initialization before feature checks.
2. . Rename `check_compatibility `to `check_avx512_compatibility`.
3. Add a test to verify output of the runtime check by validating it with `/proc/cpuinfo`.

### DIFF
--- a/jvector-native/src/main/c/jvector_simd.c
+++ b/jvector-native/src/main/c/jvector_simd.c
@@ -26,7 +26,7 @@ __m512i maskEighthBit;
 
 __attribute__((constructor))
 void initialize_constants() {
-    if (check_compatibility()) {
+    if (check_avx512_compatibility()) {
         initialIndexRegister = _mm512_setr_epi32(-16, -15, -14, -13, -12, -11, -10, -9,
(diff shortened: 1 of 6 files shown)


jshook (approved):
LGTM

ashkrisk (approved):

PR #656, opened 2026-04-07
Allow bench workflow to run on PRs from forks

ashkrisk:


.github/workflows/run-bench.yml
```diff
@@ -41,8 +42,8 @@ jobs:
 
           # Default branches based on event type
           if [[ "${{ github.event_name }}" == "pull_request" ]]; then
-            echo "Pull request detected. Using main and PR branch: ${{ github.head_ref }}"
-            BRANCHES='["main", "${{ github.head_ref }}"]'
```
jshook, line 45: This seems to be a change from "branch" to "sha", rather than also adding support for the sha. Don't we need it to work for both?

ashkrisk: Yes; but this works for two reasons:
- Git refs can be either branch names or SHAs.
- The `elif` condition just after this handles cases where the workflow is manually dispatched by providing a branch name (or SHA). This conditional branch only deals with PRs, and that gives us the freedom to pick how we want to refer to the head commit, either by branch or SHA

ashkrisk: In the latest PR (#688) I've changed this to use the Github ref instead of the SHA

ashkrisk:
Closing in favour of https://github.com/datastax/jvector/pull/688

PR #658, opened 2026-04-09
Update regression datasets

MarkWolters:
Updating regression tests (AutoBenchYAML) to use the datasets.yml configuration rather than hard-coding its datasets.

- AutoBenchYAML.java updated to dynamically load datasets for testing from yaml
- DataSets.java updated to return a specific subsection of datasets.yml based on section key passed
- new subsection ("regression-tests") added to datasets.yml
- new yml files added for the datasets used by regression (primarily the only difference is that we don't want to run as many permutations of overquery or high k value in regression due to time concerns)
- new datasets needed by regression added to dataset-metadata.yml


tlwillke (approved):
LGTM

jshook (approved):
LGTM

Issue #580, opened 2025-11-26
Add On-Disk Graph Index with Delete Tracking and Streaming N:1 Compaction


dian-lun-lin:
### Problem

```OnDiskGraphIndex``` is immutable. Today, to delete nodes or perform compaction, clients must:
1. Load each on-disk graph into an in-memory ```OnHeapGraphIndex```.
2. Apply deletions.
3. Rewrite a new on-disk graph from the fully materialized heap index.

For large graphs (millions of nodes), this approach requires loading the full graph structure, which frequently leads to OOM and does not scale.

### Proposal

Introduce an ```OnDiskGraphIndexCompactor``` that performs streaming N:1 compaction over multiple on-disk indexes without loading them into memory.

The compactor provides:

1. **Delete Tracking**
Track live/dead nodes using an in-memory ```FixedBitSet```, allowing deletion without rewriting the index or constructing an ```OnHeapGraphIndex```.

2. **Streaming N:1 Compaction that performs streaming compaction**
Write a new ```OnDiskGraphIndex``` via a streaming writer (no full materialization of the graph). Because the existing writer does not support streaming individual nodes, this proposal includes implementing a new ```CompactorWriter```.

3. **Custom Ordinal Remapping**
Allow clients to pass a custom ```OridnalMapper``` that remaps ordinals from each source index to ordinals in the compacted output index.

### Compactor Internal State
```java
List<ImmutableGraphIndex> sources;                    // Source graph indexes to compact
Map<ImmutableGraphIndex, FixedBitSet> livenodes;      // Tracks which ordinals are live 
Map<ImmutableGraphIndex, OrdinalMapper> remappers;    // Maps old ordinals -> new ordinals
```

### API Usage
```java
// Create a compactor from multiple immutable graph indexes
OnDiskGraphIndexCompactor compactor =
    new OnDiskGraphIndexCompactor(List.of(source1, source2, source3));

// Mark which nodes are still live in each source
compactor.setLiveNodes(source1, liveNodes1); // FixedBitSet
compactor.setLiveNodes(source2, liveNodes2);

// Supply remapping functions for each source
compactor.setRemapper(source1, remapper1);
compactor.setRemapper(source2, remapper2);

// Execute streaming compaction
OnDiskGraphIndex result = compactor.compact(outputPath, numNeighbors);
```

### Implementation of Compaction

1. Traverse each node in each index
Iterate through all ordinals across all sources and skip deleted ones.

2. Perform search across all sources
[...]

PR #659, opened 2026-04-14
Streaming N:1 compaction

dian-lun-lin:
This PR addresses #580 that adds `OnDiskGraphIndexCompactor`, a streaming N:1 compaction algorithm for merging multiple on-disk HNSW graph indexes into a single compacted index.                                                             
                  
  ```                                                                                                                                                        
  source[0].index  ─┐
  source[1].index  ─┤──► OnDiskGraphIndexCompactor ──► compacted.index                                                                                       
  source[N].index  ─┘                                                                                                                                        
  ```
                                                                                                                                                             
  For a full description of the algorithm and benchmarking instructions, see docs/compaction.md and benchmarks-jmh/src/main/java/io/github/jbellis/jvector/bench/CompactorBenchmark.md.
                                                                                                                                                             
  ### Support                                                                                                                                                
  - **Streaming, low-memory**: no full in-memory graph construction; runs under `-Xmx5g` even for 10M-vector, 2560-dim datasets
  - **Deletion support**: live-node `FixedBitSet` per source excludes deleted nodes from output                                                              
  - **Ordinal remapping**: user-provided `OrdinalMapper` maps each source's local ordinals to a contiguous global ordinal space; implemented `OffsetMapper` that handles the  common sequential case                                                                                                                                     
                                                                                                                                                             
  ### Usage                                                                                                                                                  
                  
  ```java
  List<OnDiskGraphIndex> sources = List.of(index0, index1, index2);
                                                                                                                                                             
  // Mark all nodes live (no deletions)                                                                                                                      
  List<FixedBitSet> liveNodes = sources.stream()                                                                                                             
      .map(s -> { var bs = new FixedBitSet(s.size()); bs.set(0, s.size()); return bs; })                                                                     
      .collect(toList());                                                                                                                                    
                                                                                                                                                             
  // Sequential ordinal remapping: source[s] node i → global offset[s] + i                                                                                   
  int offset = 0; 
  List<OrdinalMapper> remappers = new ArrayList<>();                                                                                                         
  for (var src : sources) {
      remappers.add(new OrdinalMapper.OffsetMapper(offset, src.size()));                                                                                     
      offset += src.size();
  }                                                                                                                                                          
                  
  var compactor = new OnDiskGraphIndexCompactor(                                                                                                             
      sources, liveNodes, remappers,
      VectorSimilarityFunction.COSINE,                                                                                                                       
      /* executor= */ null                  // null = use by default shared threadpool in compactor
  );                                                                                                                                                         
                  
  compactor.compact(Path.of("compacted.index"));                                                                                                             
  ```             

### Benchmarking example
```
java -Xmx220g --add-modules jdk.incubator.vector \
  -cp benchmarks-jmh/target/compactor-benchmark.jar \
  io.github.jbellis.jvector.bench.CompactorBenchmark \
  -p workloadMode=PARTITION_AND_COMPACT \
  -p datasetNames=ada002-100k \
  -p numPartitions=4 \
  -p splitDistribution=FIBONACCI \
  -p indexPrecision=FUSEDPQ \
  -wi 0 -i 1 -f 1
  ```

  ### Key Changes                                                                                                                                            
   
  - **`OnDiskGraphIndexCompactor`** — core compaction algorithm with parallel ForkJoinPool execution and backpressure windowing                              
  - **`PQRetrainer`** — balanced proportional sampling + sequential sorted reads for efficient codebook retraining
  - Minor API visibility changes to `GraphSearcher`, `GraphIndexBuilder`, and `PQVectors` required by the compactor                                          
  - **`CompactorBenchmark`** — JMH benchmark with `PARTITION_AND_COMPACT`, `PARTITION`, `COMPACT`, and `BUILD` modes                  
  - Unit tests covering basic compaction, deletions, ordinal remapping, and FusedPQ scenarios                                                                
                                                                                                                                                             
  ### Recall                                                                                                                                                 
                                                                                                                                                             
  Comparison against build-from-scratch (results averaged over three runs). 

  - Build from scratch: build with PQ, search using FusedPQ with FP reranking.
  - Compaction: build source partitions with PQ, compact using FusedPQ with FP rescoring, search using FusedPQ with FP reranking. Source partitions are based on a Fibonacci distribution with 4 partitions.
                                                                                                                                                             
  | Dataset               | Dim  | Build from Scratch | Compaction | Delta  |                                                                                
  |-----------------------|-----:|-------------------:|-----------:|-------:|
  | cap-6M                |  768 |              0.626 |      0.619 | -0.008 |                                                                                
  | cap-1M                |  768 |              0.656 |      0.656 |  0.000 |                                                                                
  | gecko-100k            |  768 |              0.690 |      0.701 | +0.011 |                                                                                
  | e5-small-v2-100k      |  384 |              0.572 |      0.586 | +0.014 |                                                                                
  | ada002-1M             | 1536 |              0.687 |      0.703 | +0.016 |                                                                                
  | e5-base-v2-100k       |  768 |              0.676 |      0.692 | +0.016 |                                                                                
  | cohere-english-v3-10M | 1024 |              0.544 |      0.561 | +0.017 |                                                                                
  | e5-large-v2-100k      | 1024 |              0.686 |      0.703 | +0.017 |                                                                                
  | ada002-100k           | 1536 |              0.751 |      0.769 | +0.018 |                                                                                
  | cohere-english-v3-1M  | 1024 |              0.593 |      0.612 | +0.019 |                                                                                
                                                                                                                                                             
  Recall is generally comparable to build-from-scratch and often better, though some datasets show small drops. All datasets compact successfully under `-Xmx5g`; compaction has also been validated on a 2560-dim 10M-vector dataset under the same constraint.

### CODE REVIEW

File: docs/compaction.md [hunk 1]
```diff
@@ -0,0 +1,199 @@
...
+## Recall 
+
+
+The table compares building from scratch with compaction under the following configurations (results averaged over three runs):
+
+- Build from scratch: build with PQ; search using FusedPQ with FP reranking.
+- Compaction: build source partitions with PQ; compact using FusedPQ with FP rescoring; search using FusedPQ with FP reranking.
+
+| Dataset              | Dim  | Build from Scratch | Compaction |  Delta |
+|----------------------|-----:|-------------------:|-----------:|-------:|
```
jshook: Do we need to modify this section for brevity?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
```diff
@@ -0,0 +1,1459 @@
...
+            return true;
+        }
+
+    }
+
+    /**
+     * Handles writing the compacted graph index to disk, managing header, node records,
+     * upper layers, and footer in the on-disk format.
+     */
+    private static final class CompactWriter implements AutoCloseable {
```
jshook: Maybe break this out. The parent file is already very large.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/reporting/SystemStatsCollector.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Background collector of {@code /proc} system metrics (CPU topology, load, memory, disk I/O).
+ * Spawns a bash process that appends JSONL lines to a file every 30 seconds.
+ */
+public final class SystemStatsCollector {
+    private static final Logger log = LoggerFactory.getLogger(SystemStatsCollector.class);
+
+    private static final String SCRIPT = String.join("\n",
```
jshook: Perhaps this logic shouldn't be broken out as it is. Instead of invoking a shell wrapper, it should probably be direct reads and pattern matching in Java.

dian-lun-lin: Thanks for the comment. The bash ProcessBuilder is replaced with a `ScheduledExecutorService` that reads `/proc/cpuinfo`, `/proc/meminfo`, `/proc/loadavg`, and `/proc/diskstats` directly via java.nio.file.Files. Same JSONL output format.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
dian-lun-lin: Extracted CompactWriter into its own top-level file.

docs/compaction.md
dian-lun-lin: I revised to the following:

Recall comparison (results averaged over three runs):                                                                                    
 - Build from scratch: build one index over the full dataset with PQ scoring; search using FusedPQ with FP reranking.                     
  - Compaction: partition the dataset into 4 source indexes (Fibonacci distribution), build each with PQ scoring, then compact into one index; search using FusedPQ with FP reranking.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -51,7 +51,7 @@ public class GraphSearcher implements Closeable {
     // Scratch data structures that are used in each {@link #searchInternal} call. These can be expensive
     // to allocate, so they're cleared and reused across calls.
```
MarkWolters, line 52: I understand the necessity to expose these methods as public, but I'd prefer not to expose more of our internal state than is necessary. If it's not possible to retain them as package level protected can we at least mark them with @Internal or something similar?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -51,7 +51,7 @@ public class GraphSearcher implements Closeable {
...
     private final NodeQueue candidates;
-    final NodeQueue approximateResults;
+    public final NodeQueue approximateResults;
```
MarkWolters: Does this need to be public? Can we add getter/setter instead?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
```diff
@@ -0,0 +1,1227 @@
...
+            throw new IllegalArgumentException("sources and remappers must have the same size");
+        }
+    }
+
+    /**
+     * Validates that liveNodes bitsets match the size of their corresponding sources.
+     */
+    private void validateLiveNodesBounds(List<OnDiskGraphIndex> sources, List<FixedBitSet> liveNodes) {
+        for (int s = 0; s < sources.size(); ++s) {
+            if (liveNodes.get(s).length() != sources.get(s).size(0)) {
```
MarkWolters: I had to modify this to `if (liveNodes.get(s).length() != sources.get(s).getIdUpperBound())` to get my integration to not blow up. Either I am calculating liveNodes incorrectly or this needs to be modified. Here's what I am doing (from my [forked repo](https://github.com/MarkWolters/cassandra/blob/compaction-integration/src/java/org/apache/cassandra/index/sai/disk/vector/CompactionGraphMerger.java#L187)
```for (var src : sources)
        {
            int liveCount = src.graph().size(0);
            int idBound = src.graph().getIdUpperBound();
            var bs = new FixedBitSet(idBound);
            if (idBound == liveCount)
            {
                bs.set(0, liveCount);
            }
            else
            {
                // Segment has tombstoned slots (idBound > liveCount); only mark actual live nodes.
                var nodeIt = src.graph().getNodes(0);
                while (nodeIt.hasNext()) bs.set(nodeIt.nextInt());
            }
            liveNodes.add(bs);
            remappers.add(new OrdinalMapper.OffsetMapper(offset, liveCount));
            offset += liveCount;
        }```

File: jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/CompactWriter.java [hunk 7]
```diff
@@ -0,0 +1,328 @@
...
+        }
+        if (fusedPQEnabled && version == 6 && level == 1 && level1PqCode != null) {
+            level1FeatureRecords.add(new UpperLayerFeatureRecord(ordinal, level1PqCode.copy()));
+        }
+    }
+
+    public void close() throws IOException {
+        final var endOfGraphPosition = writer.position();
+        writer.seek(endOfGraphPosition);
+        writer.flush();
```
MarkWolters, line 252: I think there should be a call to `writer.close()` here as well

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
```diff
@@ -0,0 +1,1771 @@
...
+        int REF = RamUsageEstimator.NUM_BYTES_OBJECT_REF;
+
+        // Shallow size of this object (header + fields)
+        // Current fields: sources, liveNodes, numLiveNodesPerSource, remappers, maxDegrees,
+        //                dimension(int), maxOrdinal(int), numTotalNodes(int),
+        //                ownsExecutor(boolean), executor, taskWindowSize(int), similarityFunction
+        long size = OH + 8L * REF + Integer.BYTES * 4 + 1;
+
+        // liveNodes: FixedBitSet per source
+        for (var entry : liveNodes) {
```
MarkWolters: check for liveNodes == null

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
```diff
@@ -0,0 +1,1771 @@
...
+            size += entry.ramBytesUsed();
+        }
+
+        // numLiveNodesPerSource: ArrayList of Integers
+        size += OH + REF + (long) numLiveNodesPerSource.size() * (OH + Integer.BYTES);
+
+        // remappers: each MapMapper holds an oldToNew HashMap and newToOld Int2IntHashMap
+        // Estimate based on the number of mappings
+        for (var mapper : remappers) {
```
MarkWolters: check for remappers == null

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FusedPQ.java
```diff
@@ -67,6 +69,23 @@ public ProductQuantization getPQ() {
         return pq;
     }
 
+    /** For compaction use. See {@link FusedFeature#getCompressor}. */
+    @Override
+    @SuppressWarnings("unchecked")
+    public io.github.jbellis.jvector.quantization.VectorCompressor<io.github.jbellis.jvector.vector.types.ByteSequence<?>> getCompressor() {
```
MarkWolters: Is there a reason we are using the fully qualified name and not importing to make it easier to read?

MarkWolters (changes requested):
The main comments I left that I am concerned about are in OnDiskGraphIndexCompactor. Without checking for null liveNodes and remappers we could get an NPE if CC calls this after those are cleaned up before refinement.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
```diff
@@ -0,0 +1,1771 @@
...
+
+            completed++;
+            inFlight--;
+
+            if (nextToSubmit < total) {
+                submitOne.accept(batches.get(nextToSubmit++));
+                inFlight++;
+            }
+            if (completed % 10 == 0) {
+                log.info("Compaction I/O progress: {}/{} batches written to disk", completed, total);
```
MarkWolters: I think this should be a debug level statement, otherwise the entire cassandra system log gets filled up with these messages

dian-lun-lin: Changed to debug level

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/feature/FusedPQ.java
dian-lun-lin: Changed to import. Thanks!

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
dian-lun-lin: Added the check. Thanks.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndexCompactor.java
dian-lun-lin: Added the check. Thanks.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/CompactWriter.java
dian-lun-lin: Added the close.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
dian-lun-lin: Added the getter for approvimateResults

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
dian-lun-lin: Now marked them with @Experimental.

MarkWolters (approved):
Thanks for addressing all my comments, approved

tlwillke (approved):
Thanks for all of the discussion and changes, including the benchmarks and docs/notes.  Looks good as an experimental feature.

PR #664, opened 2026-05-07
fix for bug when fused pq is used with no hierarchy

MarkWolters:
The Bug                                                                                                                                                                                
                                                                                                                                                                                         
  CompactWriter.writeFooter() is the counterpart of AbstractGraphIndexWriter.writeSparseLevels(). Both methods write the "source feature" block that sits between the L0 node data and   
  the actual footer bytes. OnDiskGraphIndex.loadInMemoryFeatures() reads this block to populate hierarchyCachedFeatures, which FusedPQDecoder.similarityTo() uses.                       
                                                                                                                                                                                         
  writeSparseLevels() has two branches:                                                                                                                                                  
  - Hierarchy enabled (getMaxLevel >= 1): writes a PQ code for every level-1 node                                                                                                      
  - Hierarchy disabled (getMaxLevel == 0): writes one entry — the entry node's own PQ code                                                                                               
                                                                                          
  CompactWriter.writeFooter() had only the first branch (via level1FeatureRecords). With enableHierarchy: false, level1FeatureRecords is always empty, so nothing gets written. When     
  loadInMemoryFeatures then tries to read one entry from what it thinks is that block, it consumes bytes from the footer header instead — populating hierarchyCachedFeatures with one    
  garbage entry keyed on a non-zero integer. When GraphSearcher.initializeInternal calls scoreFunction.similarityTo(entry.node) (line 331), entry.node is 0 (first remapped ordinal),    
  which is not in the garbage-keyed map → Node 0 is not in the hierarchy.                                                                                                                
                                                                                                                                                                                       
  The Fix (2 files)

  - CompactWriter.java: Added entryNode and entryNodePqCode fields, a setEntryNodePqCode() setter, and an else if (entryNodePqCode != null) branch in writeFooter() that writes exactly  
  what AbstractGraphIndexWriter writes in its no-hierarchy branch.
  - OnDiskGraphIndexCompactor.java: Refactored resolveEntryNode() into resolveEntryNodeSource() (returns {sourceIdx, originalOrdinal}), and in compact(), after compactLevels()          
  completes, reads the entry node's vector from its source graph, encodes it with the retrained PQ, and passes it to the writer before calling writeFooter().

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/CompactWriter.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/CompactWriter.java
@@ -61,6 +61,7 @@ final class CompactWriter implements AutoCloseable {
     private final ProductQuantization pq;
     private final int baseDegree;
     private final int maxOrdinal;
+    private final int entryNode;
     private final ThreadLocal<ByteBuffer> bufferPerThread;
     private final ThreadLocal<ByteSequence<?>> zeroPQ;
     private final boolean fusedPQEnabled;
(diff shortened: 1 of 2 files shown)


dian-lun-lin:
LGTM

PR #666, opened 2026-05-14
make downloader more robust

jshook:
This does a couple things to make downloads more defensive:
1. Recognizes when multiple (dataset,facet) configs point to the same file and
    - for query vectors, allows it
    - for base vectors, throws an error (for now, this is not supported without windowed reads)
    - for ground truth vectors, throws an error
2. Recognizes when different instances of the same logical (dataset,facet) config points at potentially different paths.
    - This is a hard error. Before, this was a first-finder wins scenario, which is non-deterministic.
3. - For scenarios where a single file was shared,  make the downloader concurrently safe.
    - Before it was possible for two different threads to use the same physical output buffer, but only with S3.

These are defensive changes which simply cause these types of errors to be surfaced. The bias here is to avoid surprises or non-deterministic results.

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderSimpleMFD.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderSimpleMFD.java
@@ -42,11 +42,14 @@ import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
 
 /// A dataset loader that works with fvec/ivec datasets described by YAML catalog files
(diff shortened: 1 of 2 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderSimpleMFD.java
```diff
@@ -690,6 +724,142 @@ private static String expandEnvVars(String value) {
...
+            }
+        }
+
+        List<String> hardErrors = new ArrayList<>();
+        for (var e : ownersByPath.entrySet()) {
+            List<OwnerRef> owners = e.getValue();
+            if (owners.size() <= 1) continue;
+
+            boolean involvesBaseOrGt = owners.stream()
+                    .anyMatch(o -> "base".equals(o.facet) || "gt".equals(o.facet));
```
MarkWolters, line 826: Would it be possible to make these constants ("base", "gt") configurable? This works for our current naming standards but hardcoding it here enforces a sort of invisible rule that's disconnected from the dataset creation process. I don't think it needs to be changed immediately to merge, but its something we might want to change for future flexibility.

MarkWolters (approved):
approved, but I think we should try to refrain from using hard-coded conventions in the future if possible (see comment).

jvector-examples/src/main/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderSimpleMFD.java
jshook: I've had this come up in a couple places before, and the main issue is that if there is no standard, then we have to wire the facets up everywhere and it lowers the utility of having an access layer altogether.
But what I have done also is allow all common "spellings" of a facet type to be recognized to support cases which are unambiguous.

The same names are used in upstream dataset creation (in some tools) canonically, with one favored as the generally prescribed name. We do need _some_ type of naming consistency for facets that all the tools can agree on. Happy to make a change if we can address these other concerns too.

jvector-examples/src/test/java/io/github/jbellis/jvector/example/benchmarks/datasets/DataSetLoaderSimpleMFDTest.java
```diff
@@ -17,18 +17,28 @@
 
 import com.sun.net.httpserver.HttpExchange;
 import com.sun.net.httpserver.HttpServer;
+import io.github.jbellis.jvector.example.benchmarks.datasets.DataSetLoaderSimpleMFD;
+import io.github.jbellis.jvector.example.benchmarks.datasets.DataSetMetadataReader;
```
ashkrisk, line 21: Unnecessary imports, these classes are in the same package

ashkrisk (approved):

ashkrisk (commented):
I'm running into exceptions when trying to run Grid. I deleted the catalog cache beforehand and removed everything except the default public catalog.
```
Local path dataset_cache/public/openai_v3_small_1536_100k_query_10000.fvecs is referenced by multiple dataset facets:
  openai-v3-small-1536-100k:query (from jvector-examples/yaml-configs/dataset-catalogs/.catalog-cache/include-8204d4975290e21e40a549237fc1aa7b5ed49ab3329939fe7479e28e0e3e9090.yaml.cache)
  openai-v3-small-100k:query (from jvector-examples/yaml-configs/dataset-catalogs/.catalog-cache/include-8204d4975290e21e40a549237fc1aa7b5ed49ab3329939fe7479e28e0e3e9090.yaml.cache)
```

Looks like the default public catalog may need updates before merging this.

PR #669, opened 2026-05-26
add openai_datasets

jshook:
This adds the name entries for the openai datasets in our shared config.
They still need to have some reasonable per-dataset defaults added, but they are testable now with default settings.


tlwillke (approved):
LTGM.  Thanks

MarkWolters (approved):

PR #670, opened 2026-05-29
adding memory stats to regression benchmarks

MarkWolters:
This is a minor change to the AutoBenchYAML regression testing infrastructure. We already tracked on heap and off heap memory usage but only reported the statistics to the json summary file and not to the csv file used in post processing to create the visualizations. This PR simply adds memory stats to the final rollup so that they can be represented in visualizations along with the other stats.


tlwillke (approved):
LGTM.

jshook (approved):
LGTM

PR #673, opened 2026-06-09
Release perf artifact

MarkWolters:
Updates the visualize_benchmarks.py script to add footnotes to each generated chart detailing 

- Instance type the test was run on
- JDK used to run the test
- Flags passed to the JVM
- yaml config files used for test configuration for each dataset used

Recall_at_10


jshook (approved):
A very good addition!

tlwillke:
Question: In the example, the Xmx flag is shown but the size value is not included (See the plot and -Xmxg).  Is that what you intended?

jvector-examples/src/main/resources/visualize_benchmarks.py
```diff
@@ -34,20 +36,253 @@
...
+        for line in java_lines:
+            # Evaluate conditional flag blocks
+            for m in re.finditer(
+                r'\$\{\{[^}]*matrix\.jdk\s*>=\s*(\d+)\s*&&\s*\'([^\']+)\'', line
+            ):
+                if jdk_major >= int(m.group(1)):
+                    flags.append(m.group(2).strip())
+
+            # Remove all GitHub Actions expressions, then shell variable refs
+            clean = re.sub(r'\$\{\{[^}]*\}\}', '', line)
```
tlwillke, line 211: This is a correctness fix to avoid '<dynamic>' from being stripped:

# Remove GitHub Actions expressions, but preserve shell variable placeholders
clean = re.sub(r'\$\{\{[^}]*\}\}', '', line)
clean = re.sub(r'\$\{[^}]+\}', '<dynamic>', clean)

However, it would be better to display the resolved value for HALF_MEM_GB.

tlwillke (changes requested):
One correctness fix proposed.

jvector-examples/src/main/resources/visualize_benchmarks.py
MarkWolters: fixed to show actual value for half of the memory available

tlwillke (approved):
Helpful!

PR #679, opened 2026-06-15
make cores vs threads detection more robust

jshook:
This makes detection of physical cores more robust across three key system types, on best-effort basis.
The previous logic presumed HT cores and could easily get the answer wrong on any other physical architecture.

Another related fix will delegate to PhysicalCoreExecutor DRYly, keeping thread pool sizing consistent as much as possible.


jshook:
This has already been picked into the compaction pr and will hit main when that is merged.

PR #681, opened 2026-06-17
Use bulk writes instead of per-element writes in vector serialization

r-devulap:
Replace inefficient element-by-element write loops with bulk write operations in MemorySegmentVectorProvider:

- writeFloatVector: Extract underlying float array and use writeFloats() instead of looping with writeFloat()
- writeByteSequence: Extract underlying byte array and use write() instead of looping with writeByte()

With highway as the SIMD backend #668 , the scalar writes show up as a bottleneck when constructing index on a AWS instance x8i.24xlarge (2-socket 96 core Intel GNR) . 

| Dataset           | Index Build Time Before (s) | Index Build Time After (s) |
|------------------|------------------------------|-----------------------------|
| openai-1536-1m   | 105.59                       | 45.22                        |
| openai-3072-1m   | 164.63                       | 102.75                       |


jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentVectorProvider.java
```diff
@@ -98,7 +98,7 @@ public void readByteSequence(RandomAccessReader r, ByteSequence<?> sequence) thr
     @Override
     public void writeByteSequence(IndexWriter out, ByteSequence<?> sequence) throws IOException
     {
-        for (int i = 0; i < sequence.length(); i++)
-            out.writeByte(sequence.get(i));
+        byte[] data = (byte[]) ((MemorySegmentByteSequence) sequence).get().heapBase().get();
+        out.write(data, 0, sequence.length());
     }
```
tlwillke, line 103:
```suggestion
{
        java.nio.ByteBuffer bb = ((MemorySegmentByteSequence) sequence).get().asByteBuffer();
        out.write(bb.array(), bb.arrayOffset(), bb.remaining());
    }
```

tlwillke (changes requested):
writeFloatVector is fine.  writeByteSequence has a bug.  You are ignoring slice offsets.  MemorySegmentByteSequence supports slicing, but .heapBase().get() returns the original unsliced array.  I have suggested a change that fixes this and added minimal testing.

Once this is addressed and the tests pass, it is good to go.

On performance, while I appreciate the end-to-end benchmarking and consider it necessary, I would also like to see a unit-level microbenchmark isolating writeFloatVector and writeByteSequence to quantify the exact bulk-write speedup.

jvector-native/src/main/java/io/github/jbellis/jvector/vector/MemorySegmentVectorProvider.java
r-devulap: Thanks for pointing that out. I mistakenly assumed CI would catch this bug. I’ll fix it.

r-devulap:
Had to move `MemorySegmentVectorProviderTest` to jvector-native module, it fails to run on JAVA 20.

r-devulap:
Added microbenchmarks. Let me know if that is what you what you were looking for. Here is the comparison with main branch on a x8i.24xlarge (96 core Intel GNR)
```
| Benchmark        | Length | Mode  | Cnt | Branch Score | Main Score  | Speedup (x) | Units |
|------------------|--------|-------|-----|--------------|-------------|-------------|-------|
| writeByteVector  | 512    | thrpt | 10  | 6029596.287  | 3904220.037 | 1.54x       | ops/s |
| writeByteVector  | 1024   | thrpt | 10  | 3230877.500  | 1924332.336 | 1.68x       | ops/s |
| writeByteVector  | 1536   | thrpt | 10  | 2076349.428  | 1273820.153 | 1.63x       | ops/s |
| writeFloatVector | 512    | thrpt | 10  | 1996524.347  | 1486490.747 | 1.34x       | ops/s |
| writeFloatVector | 1024   | thrpt | 10  | 978881.406   | 520671.373  | 1.88x       | ops/s |
| writeFloatVector | 1536   | thrpt | 10  | 672933.883   | 438508.050  | 1.53x       | ops/s |
```

tlwillke (approved):
Thanks for the fix and adding the microbenchmarks.  Looks good.

MarkWolters (approved):
Looks fine to me

jshook (approved):
Looks good

PR #682, opened 2026-06-19
Disable TopK/filtered pruning, preserve legacy threshold behavior

### LINKED ISSUE #678: Search pruning via usePruning can be too aggressive for non-hierarchical graph indices
The usePruning option is a graph-search optimization. When enabled, the search may skip expanding some graph nodes if the current scoring function indicates that further expansion is unlikely to improve the result set.

However, when the graph is searched, pruning can be too aggressive: the score may incorrectly make a node look like a local maximum, causing the search to skip neighbors that would have led to better results.

The result is recall loss. The effect is strongest on non-hierarchical indexes because all routing happens in the base layer. Hierarchical indexes are less affected because upper-layers perform global routing before base-layer pruning begins.

The library should disable pruning for upper layers for non-hierarchical graph indices.  It should also possibly disable pruning for certain approximate scorers that are more sensitive to it.  A better pruning heuristic should also be considered.

The problematic behavior appears to come from the pruning path that skips expanding a popped candidate’s neighbors:

```
if (scoreTracker.shouldStop()

        && candidates.size() >= rerankK - approximateResults.size()) {

    continue;

}
```

This is not just early termination. It skips edge loading / neighbor expansion for a candidate that was selected from the candidate queue.

### Recall impact of query-time pruning

Dataset: `openai-v3-large-1536-100k`  
Metric: `10-recall@10`

| Primary scoring mode | Hierarchy | usePruning | 10-recall@10 | Latency |
|---|---:|---:|---:|---:|
| PQ | Yes | Yes | 0.81 | 0.209s |
| PQ | Yes | No | 0.83 | 0.215s |
| PQ | No | Yes | 0.74 | — |
| PQ | No | No | 0.83 | — |
| Full precision | Yes | Yes | 0.96 | 0.229s |
| Full precision | Yes | No | 0.97 | 0.256s |
| Full precision | No | Yes | 0.81 | — |
| Full precision | No | No | 0.96 | — |

Query-time pruning improves latency in the full-precision hierarchical case, but it can materially reduce recall when hierarchy is disabled. The database option should therefore be treated as permission to prune only when graph configuration supports it safely.

<sub>
Configuration: `outDegree=[64]`, `efConstruction=[200]`, `neighborOverflow=[1.2f]`, `addHierarchy=[Yes/No as shown]`, `refineFinalGraph=[Yes]`, `fusedGraph=[No]`.
</sub>

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/markwolters/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/markwolters/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">

Update to add additional dataset testing:

Dataset | Primary scoring   mode | Hierarchy | usePruning | 10-recall@10 | Latency
-- | -- | -- | -- | -- | --
ada002-100k | PQ | Yes | Yes | 0.72 | 0.297
ada002-100k | PQ | Yes | No | 0.75 | 0.32
ada002-100k | PQ | No | Yes | 0.47 | 0.271
ada002-100k | PQ | No | No | 0.73 | 0.347
cohere-english-v3-100k | PQ | Yes | Yes | 0.77 | 1.008
cohere-english-v3-100k | PQ | Yes | No | 0.82 | 1.044
cohere-english-v3-100k | PQ | No | Yes | 0.49 | 0.975
cohere-english-v3-100k | PQ | No | No | 0.81 | 1.1
openai-v3-small-100k | PQ | Yes | Yes | 0.77 | 0.686
openai-v3-small-100k | PQ | Yes | No | 0.81 | 0.702
openai-v3-small-100k | PQ | No | Yes | 0.61 | 0.66
openai-v3-small-100k | PQ | No | No | 0.81 | 0.706
openai-v3-large-3072-100k | PQ | Yes | Yes | 0.73 | 0.834
openai-v3-large-3072-100k | PQ | Yes | No | 0.75 | 0.841
openai-v3-large-3072-100k | PQ | No | Yes | 0.58 | 0.855
openai-v3-large-3072-100k | PQ | No | No | 0.74 | 0.903
openai-v3-large-1536-100k | PQ | Yes | Yes | 0.8 | 0.676
openai-v3-large-1536-100k | PQ | Yes | No | 0.82 | 0.692
openai-v3-large-1536-100k | PQ | No | Yes | 0.6 | 0.663
openai-v3-large-1536-100k | PQ | No | No | 0.82 | 0.734
cohere-english-v3-1M | PQ | Yes | Yes | 0.63 | 1.093
cohere-english-v3-1M | PQ | Yes | No | 0.72 | 1.158
cohere-english-v3-1M | PQ | No | Yes | 0.31 | 1.037
cohere-english-v3-1M | PQ | No | No | 0.71 | 1.209
ada002-100k | Full precision | Yes | Yes | 0.89 | 0.266
ada002-100k | Full precision | Yes | No | 0.94 | 0.298
ada002-100k | Full precision | No | Yes | 0.8 | 0.228
ada002-100k | Full precision | No | No | 0.94 | 0.312
cohere-english-v3-100k | Full precision | Yes | Yes | 0.84 | 0.227
cohere-english-v3-100k | Full precision | Yes | No | 0.9 | 0.267
cohere-english-v3-100k | Full precision | No | Yes | 0.7 | 0.176
cohere-english-v3-100k | Full precision | No | No | 0.9 | 0.27
openai-v3-small-100k | Full precision | Yes | Yes | 0.91 | 0.268
openai-v3-small-100k | Full precision | Yes | No | 0.95 | 0.301
openai-v3-small-100k | Full precision | No | Yes | 0.69 | 0.237
openai-v3-small-100k | Full precision | No | No | 0.95 | 0.349
openai-v3-large-3072-100k | Full precision | Yes | Yes | 0.93 | 0.455
openai-v3-large-3072-100k | Full precision | Yes | No | 0.96 | 0.505
openai-v3-large-3072-100k | Full precision | No | Yes | 0.73 | 0.406
openai-v3-large-3072-100k | Full precision | No | No | 0.96 | 0.588
openai-v3-large-1536-100k | Full precision | Yes | Yes | 0.92 | 0.249
openai-v3-large-1536-100k | Full precision | Yes | No | 0.96 | 0.291
openai-v3-large-1536-100k | Full precision | No | Yes | 0.78 | 0.238
openai-v3-large-1536-100k | Full precision | No | No | 0.96 | 0.336
cohere-english-v3-1M | Full precision | Yes | Yes | 0.7 | 0.371
cohere-english-v3-1M | Full precision | Yes | No | 0.81 | 0.48
cohere-english-v3-1M | Full precision | No | Yes | 0.38 | 0.317
cohere-english-v3-1M | Full precision | No | No | 0.81 | 0.598

</body>

</html>

tlwillke:
This is a hotfix for JV 4.0.0-rc.8.  Resolves #678. 

## Summary

This hotfix disables the unsafe `threshold == 0` graph-search pruning path while preserving legacy threshold-search behavior for compatibility.

`GraphSearcher.usePruning(boolean)` is now deprecated and retained only for source and binary compatibility. Calls to this method no longer enable TopK or filtered-search pruning.

Threshold searches, where `threshold > 0`, continue to use the existing legacy threshold early-termination path. This is intentional because threshold search may already be used by customers, and changing that behavior in this hotfix could be a compatibility break.

In short:

```text
threshold == 0:
  RelaxedMonotonicityTracker is not used
  usePruning(true) is ignored
  plain and filtered TopK use standard graph-search termination

threshold > 0:
  legacy TwoPhaseTracker behavior is preserved
  threshold search behavior remains unchanged for compatibility
```

## Why This Change Is Needed

The existing TopK/filtered pruning implementation did not demonstrate reliable value and carried recall risk.

The original pruning design had several problems:

1. **Pruning was used to skip graph expansion.**  
   The search could pop a candidate and skip expanding its neighbors, then continue searching. In graph search, expansion is how better candidates are discovered, so this changes the traversal and can reduce recall.

2. **TopK pruning was redundant with standard graph-search termination.**  
   Plain TopK already has the standard frontier-vs-result-queue termination condition. Adding another relaxed-monotonicity signal can only terminate earlier than the normal search and therefore risks recall loss.

3. **Filtered TopK did not show useful benefit.**  
   This implementation already applies `acceptOrds` during traversal. In low-cardinality filtering tests, additional pruning saved almost no visits while sometimes reducing recall.

4. **Threshold pruning is imperfect but may be in use.**  
   Diagnostic threshold testing showed that the legacy threshold path can save substantial work for very selective thresholds, but it can lose recall for less-selective thresholds. Because this behavior existed previously, this hotfix preserves it rather than changing threshold semantics.

## Evidence From Filtering Tests

Filtered TopK was the primary case where pruning was expected to help. The observed benefit was negligible.

| Filter selectivity | Hierarchy | Visit reduction from pruning | Recall delta from pruning |
|---:|---:|---:|---:|
| 50% | off | 0.81% | -0.0082 |
| 10% | off | 0.05% | -0.0006 |
| 2% | off | 0.01% | -0.0003 |
| 50% | on | 0.06% | -0.0006 |
| 10% | on | 0.02% | -0.0001 |
| 2% | on | 0.00% | 0.0000 |

This indicates that the existing filtered-search implementation already gets the useful behavior from applying `acceptOrds` during traversal, and the additional relaxed pruning tracker does not provide meaningful value.

## Evidence From Threshold Diagnostics

Threshold pruning remains preserved for compatibility, but diagnostic testing showed why it should not be expanded or treated as a validated general pruning mechanism.

| Expected threshold matches | Visit reduction from legacy threshold pruning | Recall with pruning |
|---:|---:|---:|
| 25 | 79.05% | 1.000 |
| 100 | 79.05% | 0.932 |
| 500 | 68.50% | 0.618 |

### `GraphSearcher.usePruning(boolean)`

The API is deprecated and retained for compatibility.

```java
/**
 * @deprecated TopK and filtered graph-search pruning is disabled because the
 * existing heuristic can reduce recall and has not shown reliable production
 * value. This method is retained for API compatibility and has no effect.
 *
 * Threshold searches, where {@code threshold > 0}, continue to use their
 * legacy threshold early-termination behavior.
 */
@Deprecated
public void usePruning(boolean usage) {
    pruneSearch = false;
}
```

## Compatibility Notes

- The public `usePruning(boolean)` method still exists.
- Existing callers will not get source or binary breakage from this change.
- Calls to `usePruning(true)` no longer enable TopK or filtered pruning.
- Threshold search behavior is preserved for compatibility.
- Callers that previously relied on TopK/filtered pruning may see slightly higher visited counts, but recall behavior should be safer and more predictable.

### `TestPruningCompatibility`

Adds explicit coverage for the new compatibility contract.

#### `testScoreTrackerFactoryPolicy`

Verifies tracker selection:

```text
threshold == 0:
  pruneSearch=false -> NoOpTracker
  pruneSearch=true  -> NoOpTracker

threshold > 0:
  pruneSearch=false -> TwoPhaseTracker
  pruneSearch=true  -> TwoPhaseTracker
```

This confirms that `RelaxedMonotonicityTracker` is disabled for `threshold == 0`, while legacy threshold behavior is preserved.

#### `testUsePruningIgnoredForTopKAndFilteredTopK`

Verifies that `usePruning(true)` has no effect for:

```text
plain TopK
filtered TopK
hierarchy off
hierarchy on
```

The test compares pruning-on and pruning-off searches over the same graph, queries, and filters, and asserts:

```text
same returned nodes
same visited count
```

#### `testUsePruningIgnoredForThresholdSearch`

Verifies that `usePruning(true)` and `usePruning(false)` produce identical threshold-search behavior.

This confirms that the public pruning API is ignored, while the legacy threshold path remains unchanged.

### `TestLowCardinalityFiltering`

Keeps coverage for filtered TopK search behavior and filter correctness.

The test evaluates filtered search across multiple selectivities:

```text
50%
10%
2%
```

For each selectivity, the test compares pruning-off and pruning-on behavior over the same graph, queries, and filter set. It validates that returned nodes satisfy `acceptOrds` and records the visit/recall effect of pruning.

This test documents why filtered TopK pruning is being disabled: the measured visit reduction was negligible while recall could decrease slightly.

### `Test2DThreshold`

Continues to validate legacy threshold-search behavior.

This test is retained because threshold searches are intentionally preserved for compatibility. It checks that threshold queries continue to return sufficient matching nodes and meet the existing visited/recall expectations.

## Release Note Text

```markdown
Deprecated and disabled `GraphSearcher.usePruning(boolean)` for TopK and filtered searches.

The existing TopK/filter pruning heuristic could reduce recall and did not show reliable production value. The method remains available for source and binary compatibility, but enabling it no longer activates pruning for `threshold == 0` searches.

Legacy threshold-search behavior is preserved: searches with `threshold > 0` continue to use the existing threshold early-termination path.
```

## Future Work

Any future pruning implementation should be reintroduced only after a separate redesign and validation effort. A replacement should:

- never skip graph expansion for popped candidates;
- avoid upper-layer pruning;
- avoid redundant pruning for plain TopK;
- clearly separate TopK, filtered TopK, and threshold/range semantics;
- define the traversal signal it tracks;
- include representative recall and latency benchmarks across production-like datasets, dimensions, filter selectivities, and threshold selectivities.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
@@ -86,7 +86,7 @@ public class GraphSearcher implements Closeable {
         this.rerankedResults = new NodeQueue(new BoundedLongHeap(100), NodeQueue.Order.MIN_HEAP);
         this.visited = new IntHashSet();
 
-        this.pruneSearch = true;
+        this.pruneSearch = false;
         this.scoreTrackerFactory = new ScoreTracker.ScoreTrackerFactory();
     }
 
@@ -117,12 +117,16 @@ public class GraphSearcher implements Closeable {
     }
 
     /**
-     * When using pruning, we are using a heuristic to terminate the search earlier.
-     * In certain cases, it can lead to speedups. This is set to false by default.
-     * @param usage a boolean that determines whether we do early termination or not.
+     * @deprecated TopK and filtered graph-search pruning is disabled because the
+     * existing heuristic can reduce recall and has not shown reliable production
+     * value. This method is retained for API compatibility and has no effect.
+     *
+     * Threshold searches, where {@code threshold > 0}, continue to use their
+     * legacy threshold early-termination behavior.
      */
+    @Deprecated
     public void usePruning(boolean usage) {
-        pruneSearch = usage;
+        pruneSearch = false;
     }
 
     /**
@@ -339,14 +343,17 @@ public class GraphSearcher implements Closeable {
 
     private boolean stopSearch(NodeQueue localCandidates, ScoreTracker scoreTracker, int rerankK, float threshold) {
         float topCandidateScore = localCandidates.topScore();
+
         // we're done when we have K results and the best candidate is worse than the worst result so far
         if (approximateResults.size() >= rerankK && topCandidateScore < approximateResults.topScore()) {
             return true;
         }
-        // when querying by threshold, also stop when we are probabilistically unlikely to find more qualifying results
+
+        // preserve legacy threshold early termination
         if (threshold > 0 && scoreTracker.shouldStop()) {
             return true;
         }
+
         return false;
     }
 
@@ -394,8 +401,9 @@ public class GraphSearcher implements Closeable {
             assert approximateResults.size() == 0; // should be cleared by setEntryPointsFromPreviousLayer
             approximateResults.setMaxSize(rerankK);
 
-            // track scores to predict when we are done with threshold queries
-            var scoreTracker = scoreTrackerFactory.getScoreTracker(pruneSearch, rerankK, threshold);
+            // TopK and filtered pruning are disabled. Threshold searches retain their
+            // legacy threshold early-termination path inside ScoreTrackerFactory.
+            var scoreTracker = scoreTrackerFactory.getScoreTracker(false, rerankK, threshold);
 
             // the main search loop
             while (candidates.size() > 0) {
--- a/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestLowCardinalityFiltering.java
+++ b/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestLowCardinalityFiltering.java
@@ -51,8 +51,8 @@ public class TestLowCardinalityFiltering extends LuceneTestCase {
      */
     @Test
     public void testLowCardinalityFiltering() throws IOException {
-            testLowCardinalityFiltering(32, 0.044f, 0.91f, false);
-            testLowCardinalityFiltering(32, 0.048f, 0.93f, true);
+            testLowCardinalityFiltering(32, 0.055f, 0.95f, false);
+            testLowCardinalityFiltering(32, 0.055f, 0.95f, true);
     }
     public void testLowCardinalityFiltering(int maxDegree, float visitedRatioThreshold, float recallThreshold, boolean addHierarchy) throws IOException {
         var R = getRandom();
--- a/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestPruningCompatibility.java
+++ b/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestPruningCompatibility.java
@@ -1,0 +1,212 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.jbellis.jvector.graph;
+
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+import io.github.jbellis.jvector.LuceneTestCase;
+import io.github.jbellis.jvector.TestUtil;
+import io.github.jbellis.jvector.graph.similarity.DefaultSearchScoreProvider;
+import io.github.jbellis.jvector.util.Bits;
+import io.github.jbellis.jvector.util.FixedBitSet;
+import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
+import io.github.jbellis.jvector.vector.types.VectorFloat;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
+public class TestPruningCompatibility extends LuceneTestCase {
+    private static final int N_VECTORS = 10_000;
+    private static final int N_QUERIES = 10;
+    private static final int DIMENSIONS = 16;
+    private static final int TOP_K = 10;
+    private static final int RERANK_K = 100;
+    private static final int THRESHOLD_TOP_K_CAP = 1_000;
+    private static final int MAX_DEGREE = 32;
+    private static final VectorSimilarityFunction SIMILARITY = VectorSimilarityFunction.COSINE;
+
+    @Test
+    public void testScoreTrackerFactoryPolicy() {
+        var factory = new ScoreTracker.ScoreTrackerFactory();
+
+        assertTrue(factory.getScoreTracker(false, RERANK_K, 0.0f) instanceof ScoreTracker.NoOpTracker);
+        assertTrue(factory.getScoreTracker(true, RERANK_K, 0.0f) instanceof ScoreTracker.NoOpTracker);
+
+        // Preserve legacy threshold behavior: threshold searches still use TwoPhaseTracker.
+        assertTrue(factory.getScoreTracker(false, RERANK_K, 0.5f) instanceof ScoreTracker.TwoPhaseTracker);
+        assertTrue(factory.getScoreTracker(true, RERANK_K, 0.5f) instanceof ScoreTracker.TwoPhaseTracker);
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
+    public void testUsePruningIgnoredForTopKAndFilteredTopK() throws IOException {
+        for (boolean addHierarchy : List.of(false, true)) {
+            Fixture fixture = buildFixture(addHierarchy);
+
+            for (VectorFloat<?> query : fixture.queries) {
+                assertSameWithPruningOffAndOn(fixture, query, Bits.ALL, 0.0f, TOP_K, RERANK_K);
+                assertSameWithPruningOffAndOn(fixture, query, fixture.evenOrds, 0.0f, TOP_K, RERANK_K);
+            }
+        }
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
+    public void testUsePruningIgnoredForThresholdSearch() throws IOException {
+        for (boolean addHierarchy : List.of(false, true)) {
+            Fixture fixture = buildFixture(addHierarchy);
+
+            for (VectorFloat<?> query : fixture.queries) {
+                float threshold = exactThreshold(fixture.ravv, query, 100);
+
+                assertSameWithPruningOffAndOn(
+                        fixture,
+                        query,
+                        Bits.ALL,
+                        threshold,
+                        THRESHOLD_TOP_K_CAP,
+                        THRESHOLD_TOP_K_CAP);
+            }
+        }
+    }
+
+    private void assertSameWithPruningOffAndOn(Fixture fixture,
+                                               VectorFloat<?> query,
+                                               Bits acceptOrds,
+                                               float threshold,
+                                               int topK,
+                                               int rerankK) {
+        SearchResult pruningOff = search(fixture, query, acceptOrds, threshold, topK, rerankK, false);
+        SearchResult pruningOn = search(fixture, query, acceptOrds, threshold, topK, rerankK, true);
+
+        assertEquals(pruningOff.getVisitedCount(), pruningOn.getVisitedCount());
+        assertEquals(pruningOff.getNodes().length, pruningOn.getNodes().length);
+        assertArrayEquals(sortedNodes(pruningOff), sortedNodes(pruningOn));
+
+        if (threshold > 0.0f) {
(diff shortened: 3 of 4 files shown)


jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestLowCardinalityFiltering.java
```diff
@@ -51,8 +51,8 @@ It splits the vectors in two classes (with probability 0.5) and tests that the f
      */
     @Test
     public void testLowCardinalityFiltering() throws IOException {
-            testLowCardinalityFiltering(32, 0.044f, 0.91f, false);
-            testLowCardinalityFiltering(32, 0.048f, 0.93f, true);
```
MarkWolters, line 55: Nit: why are we changing test conditions here?

MarkWolters (changes requested):
Code looks good, as do the tests. There is a minor conflict in GraphSearcher that needs to be resolved before this can be merged.

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestLowCardinalityFiltering.java
tlwillke: Removing the destructive pruning will slight increase the nodes visited count.  But the recall vs search latency tradeoff is worth it.

tlwillke:
This branch is a hotfix for 4.0.0-rc.8.  We need to cherry pick the fix for the latest main.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
```diff
@@ -117,21 +117,16 @@ public ImmutableGraphIndex.View getView() {
...
-    /**
-     * When using pruning, we are using a heuristic to terminate the search earlier.
-     * In certain cases, it can lead to speedups. This is set to false by default.
-     * @param usage a boolean that determines whether we do early termination or not.
+     * @deprecated TopK and filtered graph-search pruning is disabled because the
+     * existing heuristic can reduce recall and has not shown reliable production
+     * value. This method is retained for API compatibility and has no effect.
+     *
+     * Threshold searches, where {@code threshold > 0}, continue to use their
+     * legacy threshold early-termination behavior.
```
ashkrisk, line 125: A clarification might be helpful here since this tripped me up a bit.

```suggestion
     * Threshold searches, where {@code threshold > 0}, continue to use their
     * legacy threshold early-termination behavior which disregards this setting.
```

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestPruningCompatibility.java
```diff
@@ -0,0 +1,212 @@
...
+                                               Bits acceptOrds,
+                                               float threshold,
+                                               int topK,
+                                               int rerankK) {
+        SearchResult pruningOff = search(fixture, query, acceptOrds, threshold, topK, rerankK, false);
+        SearchResult pruningOn = search(fixture, query, acceptOrds, threshold, topK, rerankK, true);
+
+        assertEquals(pruningOff.getVisitedCount(), pruningOn.getVisitedCount());
+        assertEquals(pruningOff.getNodes().length, pruningOn.getNodes().length);
+        assertArrayEquals(sortedNodes(pruningOff), sortedNodes(pruningOn));
```
ashkrisk, line 104: nit: should we be sorting the nodes here? If we're expecting the exact same results that would include the order of the results as well.

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestPruningCompatibility.java
```diff
@@ -0,0 +1,212 @@
...
+    private static final int THRESHOLD_TOP_K_CAP = 1_000;
+    private static final int MAX_DEGREE = 32;
+    private static final VectorSimilarityFunction SIMILARITY = VectorSimilarityFunction.COSINE;
+
+    @Test
+    public void testScoreTrackerFactoryPolicy() {
+        var factory = new ScoreTracker.ScoreTrackerFactory();
+
+        assertTrue(factory.getScoreTracker(false, RERANK_K, 0.0f) instanceof ScoreTracker.NoOpTracker);
+        assertTrue(factory.getScoreTracker(true, RERANK_K, 0.0f) instanceof ScoreTracker.NoOpTracker);
```
ashkrisk, line 53: nit: A comment here would improve readability
```java
// `usePruning` parameter is deprecated and should have no effect.
```

```suggestion
        // `usePruning` parameter is deprecated and should have no effect.
        assertTrue(factory.getScoreTracker(false, RERANK_K, 0.0f) instanceof ScoreTracker.NoOpTracker);
        assertTrue(factory.getScoreTracker(true, RERANK_K, 0.0f) instanceof ScoreTracker.NoOpTracker);
```

ashkrisk (approved):
I ran a test with cohere-1M, everything looks as expected. There's a big recall improvement with hierarchy off.

| topK  |  overquery  |  usePruning  |  addHierarchy  |  recall_main  |  recall_pr  |  recall_change |
| --- | --- | --- | --- | --- | --- | --- |
|  10  |  1.0  |  True  |  True  |  0.57925  |  0.59661  |  0.01736 |
|  10  |  2.0  |  True  |  True  |  0.75996  |  0.77704  |  0.01708 |
|  10  |  1.0  |  False  |  True  |  0.59821  |  0.59661  |  -0.00160 |
|  10  |  2.0  |  False  |  True  |  0.77833  |  0.77704  |  -0.00129 |
|  10  |  1.0  |  True  |  False  |  0.45247  |  0.58810  |  0.13563 | 
|  10  |  2.0  |  True  |  False  |  0.69523  |  0.77637  |  0.08114 |
|  10  |  1.0  |  False  |  False  |  0.57372  |  0.58810  |  0.01438 |
|  10  |  2.0  |  False  |  False  |  0.77301  |  0.77637  |  0.00336 |

jshook (approved):
I found nothing more to comment on, and ran locally to sanity check.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
tlwillke: Good point.  I'll add this to the cherry pick.

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestPruningCompatibility.java
tlwillke: That's a good point.  This is rigorous enough for now.  We rarely measure mAP and metrics that depend on strict rank.

tlwillke:
Closing this PR but retaining the branch.  It has been cherry picked into #686, although this branch was used for a hotfix.

PR #683, opened 2026-06-19
Add release notes for PR #681

r-devulap:
Adding release notes for #681

### DIFF
--- a/docs/release notes/4.1.0/681.performance.md
+++ b/docs/release notes/4.1.0/681.performance.md
@@ -1,28 +1,23 @@
-## Performance Improvements: Vector Serialization
+### Faster Index Construction via Bulk Vector Serialization
 
-### Summary
+**Description**
 Optimized vector serialization in `MemorySegmentVectorProvider` by replacing
-per-element scalar writes with bulk operations, significantly reducing
-serialization overhead.
+per-element scalar writes with bulk operations. At high core counts,
+per-element scalar writes become a serialization bottleneck: multiple threads
+contend on a single lock, stalling index construction and negating the benefit
+of additional cores. Bulk writes eliminate that contention, restoring scaling
+at large core counts.
 
-### Key Changes
-- **`writeFloatVector`**
-  - Uses backing `float[]` with `writeFloats()` instead of per-element
-    `writeFloat()`
+- **`writeFloatVector`** — uses backing `float[]` with `writeFloats()` instead
+  of per-element `writeFloat()`
+- **`writeByteSequence`** — uses backing `byte[]` with bulk `write()` instead
+  of per-element `writeByte()`
 
-- **`writeByteSequence`**
-  - Uses backing `byte[]` with bulk `write()` instead of per-element
-    `writeByte()`
+**How to enable**
+No configuration required. The improvement is applied automatically when using
+the native SIMD backend (`jvector-native`).
 
-### Impact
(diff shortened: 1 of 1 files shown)


docs/release notes/4.1.0/681.performance.md
```diff
@@ -0,0 +1,28 @@
+## Performance Improvements: Vector Serialization
+
+### Summary
+Optimized vector serialization in `MemorySegmentVectorProvider` by replacing
+per-element scalar writes with bulk operations, significantly reducing
+serialization overhead.
```
ashkrisk: Have we observed index build time improvements on systems with a single NUMA node? If not we might want to edit the release notes to reflect that.

docs/release notes/4.1.0/681.performance.md
```diff
@@ -0,0 +1,28 @@
+## Performance Improvements: Vector Serialization
```
ashkrisk: Release notes should not contain any `##`-level tags, refer the [Release Notes Readme](https://github.com/datastax/jvector/blob/main/docs/release%20notes/README.md#file-content).

docs/release notes/4.1.0/681.performance.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Summary
+Optimized vector serialization in `MemorySegmentVectorProvider` by replacing
+per-element scalar writes with bulk operations, significantly reducing
+serialization overhead.
+
+### Key Changes
```
ashkrisk: I think this section can be omitted given that it discusses implementation details. The summary already mentions that per-element scalar writes are replaced by bulk operations, and anyone interested in exactly how it's done can refer the PR.

docs/release notes/4.1.0/681.performance.md
```diff
@@ -0,0 +1,28 @@
...
+    `writeByte()`
+
+### Impact
+- Removes a major scalar-write bottleneck in the native SIMD backend
+- Improves index construction performance on large Intel hosts (e.g., AWS
+  `x8i.24xlarge`)
+
+### Index Build Improvements
+- **openai-1536-1m:** ~57% faster (105.59s → 45.22s)
+- **openai-3072-1m:** ~38% faster (164.63s → 102.75s)
```
ashkrisk: If we mention hard numbers we should probably mention the hardware we're using.

docs/release notes/4.1.0/681.performance.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Microbenchmark Summary
```
ashkrisk: IMO microbenchmarks are unnecessary for the release notes, but I don't have strong feelings about it.

docs/release notes/4.1.0/681.performance.md
r-devulap: Re-did the notes based on guidelines.

docs/release notes/4.1.0/681.performance.md
r-devulap: I will prefer to keep it.

docs/release notes/4.1.0/681.performance.md
r-devulap: Its not really a NUMA problem, its a scaling problem. I think the updated version should make it clear now.

ashkrisk (approved):

MarkWolters (approved):

PR #685, opened 2026-06-19
[FEATURE] Provide a possibility to store and load an empty graph

### LINKED ISSUE #684: [FEATURE] Provide a possibility to store and load an empty graph
### Introduction

At the moment, an empty graph (with no nodes and layers) could not be persisted on (and consequently, restored from) disk. In most cases, this is not an issue but causes major headaches to `opensearch-jvector` plugin in combination with Apache Lucene indexing data structures.

### Problem

Due to the fact that Apache Lucene manages immutable segments and does merges, it may happen that some segments end up with vector fields that are backed by empty graphs (fe, updates or deletes could easily lead to such state). Attempts to merge such segments with empty graphs caused https://github.com/opensearch-project/opensearch-jvector/pull/417, where we attempted to workaround the issue and not store graph at all.

The fix from above works for most index shapes but we run into another issue https://github.com/opensearch-project/opensearch-jvector/pull/552 that turned out to be more profound. The gist of it is  two fold:
 - indices with many vector fields that frequently override documents or/and write into different vector fields 
 - Apache Lucene does not prune unused field info from the segments (see please https://issues.apache.org/jira/browse/LUCENE-8551)

The combination of issues from above leads `opensearch-jvector` to read segments which supposed to have vector fields (unused field info) but not backed by any graph (empty graph cannot be persisted). Such split brain leads to issue(s) like https://github.com/opensearch-project/opensearch-jvector/pull/552 where  `opensearch-jvector` has no graph metadata for field.

### How storing empty graph helps

Storing an empty graph helps `opensearch-jvector` to track unused fields by having associated empty graph structure, which is the only source of truth for vector fields (it holds dimensions, empty graph, similarity function). Although empty, it closes the de-sync gap between Apache Lucene field infos and `opensearch-jvector`  vector fields.

### Versions Affected
- 4.x (we are on 4.0.0-rc8 but all previous versions are impacted)

reta:
Provide a possibility to store and load an empty graph, see please https://github.com/datastax/jvector/issues/684 for motivation and background.

Closes https://github.com/datastax/jvector/issues/684

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
@@ -48,7 +48,7 @@ import java.io.IOException;
  * in a View that should be created per accessing thread.
  */
 public interface ImmutableGraphIndex extends AutoCloseable, Accountable {
-    int OMITTED = OrdinalMapper.OMITTED; // same as OrdinalMapper, since OrdinalMapper::oldToNew may return it
+    int ENTRY_NODE_ABSENT = -1;
 
     /** Returns the number of nodes in the graph */
     @Deprecated
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
@@ -48,6 +48,7 @@ import java.io.IOException;
  * in a View that should be created per accessing thread.
  */
 public interface ImmutableGraphIndex extends AutoCloseable, Accountable {
+    /** Marks entry node as absent (fe, empty graph) */
     int ENTRY_NODE_ABSENT = -1;
 
     /** Returns the number of nodes in the graph */
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
@@ -24,6 +24,7 @@
 
 package io.github.jbellis.jvector.graph;
 
+import io.github.jbellis.jvector.graph.disk.OrdinalMapper;
 import io.github.jbellis.jvector.graph.similarity.ScoreFunction;
 import io.github.jbellis.jvector.util.Accountable;
 import io.github.jbellis.jvector.util.Bits;
@@ -35,7 +36,6 @@ import java.util.Objects;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.function.Function;
 
 /**
  * Represents a graph-based vector index.  Nodes are represented as ints, and edges are
@@ -48,6 +48,8 @@ import java.util.function.Function;
  * in a View that should be created per accessing thread.
  */
 public interface ImmutableGraphIndex extends AutoCloseable, Accountable {
+    int OMITTED = OrdinalMapper.OMITTED; // same as OrdinalMapper, since OrdinalMapper::oldToNew may return it
+
     /** Returns the number of nodes in the graph */
     @Deprecated
     default int size() {
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
@@ -131,9 +131,6 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
 
     private List<Int2ObjectHashMap<int[]>> loadInMemoryLayers(RandomAccessReader in) throws IOException {
         var imn = new ArrayList<Int2ObjectHashMap<int[]>>(layerInfo.size());
-        if (layerInfo.isEmpty()) {
-            return imn;
-        }
         // For levels > 0, we load adjacency into memory
         imn.add(null); // L0 placeholder so we don't have to mangle indexing
         long L0size = idUpperBound * (inlineBlockSize + Integer.BYTES * (1L + 1L + layerInfo.get(0).degree));
@@ -339,12 +336,12 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
 
     @Override
     public int size(int level) {
-        return  layerInfo.isEmpty() ? 0 : layerInfo.get(level).size;
+        return layerInfo.get(level).size;
     }
 
     @Override
     public int getDegree(int level) {
-        return layerInfo.isEmpty() ? 0 : layerInfo.get(level).degree;
+        return layerInfo.get(level).degree;
     }
 
     @Override
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
@@ -94,7 +94,7 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
         this.version = header.common.version;
         this.layerInfo = header.common.layerInfo;
         this.dimension = header.common.dimension;
-        if (header.common.entryNode == OMITTED) {
+        if (header.common.entryNode == ENTRY_NODE_ABSENT) {
             this.entryNode = null;
         } else {
             this.entryNode = new NodeAtLevel(header.common.layerInfo.size() - 1, header.common.entryNode);
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
@@ -50,7 +50,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -95,7 +94,11 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
         this.version = header.common.version;
         this.layerInfo = header.common.layerInfo;
         this.dimension = header.common.dimension;
-        this.entryNode = new NodeAtLevel(header.common.layerInfo.size() - 1, header.common.entryNode);
+        if (header.common.entryNode == OMITTED) {
+            this.entryNode = null;
+        } else {
+            this.entryNode = new NodeAtLevel(header.common.layerInfo.size() - 1, header.common.entryNode);
+        }
         this.idUpperBound = header.common.idUpperBound;
         this.features = header.features;
         this.neighborsOffset = neighborsOffset;
@@ -128,6 +131,9 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
 
     private List<Int2ObjectHashMap<int[]>> loadInMemoryLayers(RandomAccessReader in) throws IOException {
         var imn = new ArrayList<Int2ObjectHashMap<int[]>>(layerInfo.size());
+        if (layerInfo.isEmpty()) {
+            return imn;
+        }
         // For levels > 0, we load adjacency into memory
         imn.add(null); // L0 placeholder so we don't have to mangle indexing
         long L0size = idUpperBound * (inlineBlockSize + Integer.BYTES * (1L + 1L + layerInfo.get(0).degree));
@@ -333,12 +339,12 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
 
     @Override
     public int size(int level) {
-        return layerInfo.get(level).size;
+        return  layerInfo.isEmpty() ? 0 : layerInfo.get(level).size;
     }
 
     @Override
     public int getDegree(int level) {
-        return layerInfo.get(level).degree;
+        return layerInfo.isEmpty() ? 0 : layerInfo.get(level).degree;
     }
 
     @Override
@@ -435,7 +441,7 @@ public class OnDiskGraphIndex implements ImmutableGraphIndex, AutoCloseable, Acc
 
     @Override
     public int getMaxLevel() {
-        return entryNode.level;
+        return entryNode == null ? 0 : entryNode.level;
     }
 
     @Override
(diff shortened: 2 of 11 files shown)


jvector-tests/src/test/java/io/github/jbellis/jvector/graph/GraphIndexBuilderTest.java
```diff
@@ -156,6 +157,34 @@ public void testSaveAndLoad() throws IOException {
...
+        Supplier<GraphIndexBuilder> newBuilder = () ->
+            new GraphIndexBuilder(ravv, VectorSimilarityFunction.COSINE, 2, 10, 1.0f, 1.0f, true);
+
+        var indexDataPath = testDirectory.resolve("index_builder_empty.data");
+        var builder = newBuilder.get();
+
+        var graph = TestUtil.buildSequentially(builder, ravv);
+
+        try (var out = TestUtil.openDataOutputStream(indexDataPath)) {
+            ((OnHeapGraphIndex) graph).setAllMutationsCompleted();
```
ashkrisk, line 175: nit: `TestUtil.buildSequentially` already calls `setAllMutationsCompleted` (transitively)

jvector-base/src/main/java/io/github/jbellis/jvector/graph/ImmutableGraphIndex.java
```diff
@@ -48,6 +48,8 @@
  * in a View that should be created per accessing thread.
  */
 public interface ImmutableGraphIndex extends AutoCloseable, Accountable {
+    int OMITTED = OrdinalMapper.OMITTED; // same as OrdinalMapper, since OrdinalMapper::oldToNew may return it
```
ashkrisk: I'd suggest defining a new constant, say `ENTRY_NODE_ABSENT`, rather than reusing `OrdinalMapper.OMITTED`. I don't think there is any advantage in reusing the constant (expand the widget for more details).

<details><summary>More details</summary>

The description for `OrdinalMapper.OMITTED` mentions that this value may be returned `newToOld`, but no mention is made of `oldToNew`. In fact, the default OrdinalMapper implementation (`AbstractGraphIndexWriter::sequentialRenumbering`) appears to handle holes in the "original" ordinals by mapping them to `-1` (`OrdinalMapper.OMITTED` is currently defined as `Integer.MIN_VALUE`).

https://github.com/datastax/jvector/blob/c059e8695256d23dcc51e87fe397528c27dbf532/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OrdinalMapper.java#L31-L34

https://github.com/datastax/jvector/blob/c059e8695256d23dcc51e87fe397528c27dbf532/jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/AbstractGraphIndexWriter.java#L146-L148

</details>

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
```diff
@@ -333,12 +339,12 @@ public int getDimension() {
...
     @Override
     public int size(int level) {
-        return layerInfo.get(level).size;
+        return  layerInfo.isEmpty() ? 0 : layerInfo.get(level).size;
     }
 
     @Override
     public int getDegree(int level) {
-        return layerInfo.get(level).degree;
+        return layerInfo.isEmpty() ? 0 : layerInfo.get(level).degree;
```
ashkrisk: I don't think it's valid to have an empty `layerInfo`, even for an empty graph. Maybe the default should be removed, I'd say it's desirable to keep the existing behaviour which throws an error in case the `layerInfo` is somehow empty rather potentially causing a logic error somewhere down the line by returning zero. Similarly for all the other locations.

ashkrisk (commented):
@reta thanks for this PR. I've left a few comments. Additionally, I tried running a test by saving and loading an empty heap graph to an `OnDiskGraphIndex`, but this failed an assertion while loading the graph. The test in this PR only covers `OnHeapGraphIndex::save` and `OnHeapGraphIndex::load`.

I'll try and take a look at this later. In the meantime, if you manage to figure out the issue, please do update the PR.

<details><summary>Failure details</summary>

The error:
```
java.lang.AssertionError: Node ID -1419795886 out of bounds for layer 1
```

The sample program is below, please run with assertions enabled.

```java
import java.io.IOException;
import java.nio.file.Path;
import java.util.List;
import java.util.Map;

import io.github.jbellis.jvector.disk.ReaderSupplierFactory;

import io.github.jbellis.jvector.graph.GraphIndexBuilder;
import io.github.jbellis.jvector.graph.ListRandomAccessVectorValues;
import io.github.jbellis.jvector.graph.disk.GraphIndexWriter;
import io.github.jbellis.jvector.graph.disk.GraphIndexWriterTypes;
import io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex;
import io.github.jbellis.jvector.graph.disk.feature.FeatureId;
import io.github.jbellis.jvector.graph.disk.feature.InlineVectors;
import io.github.jbellis.jvector.graph.similarity.BuildScoreProvider;
import io.github.jbellis.jvector.vector.VectorSimilarityFunction;

public class Test {
    public static void main(String[] forwardArgs) throws IOException {
        
        var dim = 1024;
        var vsf = VectorSimilarityFunction.DOT_PRODUCT;

        var M = 32;
        var ef = 100;
        var nOv = 1.2f;
        var alpha = 1.2f;
        var hier = true;

        var ravv = new ListRandomAccessVectorValues(List.of(), dim);
        var bsp = BuildScoreProvider.randomAccessScoreProvider(ravv, vsf);
        var graphPath = Path.of("./local/tmp.jvgraph");

        try (
            var builder = new GraphIndexBuilder(bsp, dim, M, ef, nOv, alpha, hier);
            var graph = builder.build(ravv);
            var writer = GraphIndexWriter.getBuilderFor(GraphIndexWriterTypes.RANDOM_ACCESS_PARALLEL, graph, graphPath)
                .with(new InlineVectors(dim))
                .build();
        ) {
            writer.write(Map.of(FeatureId.INLINE_VECTORS, i -> new InlineVectors.State(ravv.getVector(i))));
            System.out.println("OHG max level = " + graph.getMaxLevel());
        }

        try (
            var readerSupplier = ReaderSupplierFactory.open(graphPath);
            var graph = OnDiskGraphIndex.load(readerSupplier);
        ) {
            var count = 0;
            var niter = graph.getNodes(0);
            while (niter.hasNext()) {
                count++;
                System.out.println(">>> " + niter.next());
            }
            System.out.println(String.format("Counted %d nodes", count));
            graph.getDegree(0);
        }
    }
}
```
</details>

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/GraphIndexBuilderTest.java
reta: I believe TestUtil.buildSequentially calls cleanup but which does not clear all mutations, removing explicit call for setAllMutationsCompleted will lead to:

```
java.lang.IllegalStateException: Cannot save a graph with pending mutations. Call cleanup() first
```

The explicit call is needed since cleanup() does short-circuit on empty graph (no nodes) and exits early.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/OnDiskGraphIndex.java
reta: Fixed that, the problem was caused by OnHeapGraph::getMaxLevel (https://github.com/datastax/jvector/pull/685/changes#diff-6fdbd8088adf1995ab097e86d1ca6f9f9f4cc1384a06bb438180284659812f2b) which returned -1 instead of 0, thanks!

reta:
> I'll try and take a look at this later. In the meantime, if you manage to figure out the issue, please do update the PR.

Thanks @ashkrisk , sorry forgot to update on this one,

```
Jun 23, 2026 8:53:10 A.M. io.github.jbellis.jvector.vector.VectorizationProvider lookup
WARNING: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
Jun 23, 2026 8:53:16 A.M. io.github.jbellis.jvector.util.PhysicalCoreExecutor resolvePhysicalCoreCount
INFO: PhysicalCoreExecutor: using 10 worker threads, determined by OS topology probe on "Mac OS X" (detected 10 physical cores, clamped to availableProcessors=10).
Jun 23, 2026 8:53:16 A.M. io.github.jbellis.jvector.util.PhysicalCoreExecutor resolvePhysicalCoreCount
WARNING: PhysicalCoreExecutor: previous auto-sized worker threads on this node would have been 5, but aligning to cores now sets them to 10. Set the jvector.physical_core_count system property to override.
08:53:16.350 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Writing common header at position 0
08:53:16.351 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Writing 1 layers
08:53:16.351 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Common header finished writing at position 288
08:53:16.361 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Writing common header at position 296
08:53:16.361 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Writing 1 layers
08:53:16.361 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Common header finished writing at position 584
08:53:16.361 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Writing common header at position 0
08:53:16.361 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Writing 1 layers
08:53:16.362 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Common header finished writing at position 288
OHG max level = 0
....

08:53:16.386 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Loading OnDiskGraphIndex from offset=0
08:53:16.386 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Loading common header at position 0
08:53:16.388 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - 1 layers
08:53:16.388 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Common header finished reading at position 288
08:53:16.388 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Header loaded: version=6, dimension=1024, entryNode=-1, layerInfoCount=1
08:53:16.388 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Position after reading header=296
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Version 5+ onwards uses a footer instead of header for metadata. Loading from footer
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Loading OnDiskGraphIndex footer from offset=600
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Loading header offset=592
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Loading OnDiskGraphIndex header from offset=296
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Loading common header at position 296
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - 1 layers
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.CommonHeader - Common header finished reading at position 584
08:53:16.389 [main] DEBUG io.github.jbellis.jvector.graph.disk.OnDiskGraphIndex - Header loaded: version=6, dimension=1024, entryNode=-1, layerInfoCount=1, Position after reading header=592
Counted 0 nodes
```

I run it against latest changes with assertions enabled - no issues, please let me know if I am missing something, thanks!

jvector-tests/src/test/java/io/github/jbellis/jvector/TestUtil.java
```diff
@@ -543,4 +543,107 @@ public long ramBytesUsed() {
...
+                    
+                    @Override
+                    public int size() {
+                        return 0;
+                    }
+                    
+                    @Override
+                    public void processNeighbors(int level, int node, ScoreFunction scoreFunction, IntMarker visited,
+                            NeighborProcessor neighborProcessor) {
+                    }
```
ashkrisk, line 582: nit: better to throw an exception here since all `node` values are invalid for the empty graph.

jvector-tests/src/test/java/io/github/jbellis/jvector/TestUtil.java
```diff
@@ -543,4 +543,107 @@ public long ramBytesUsed() {
...
+                    
+                    @Override
+                    public Bits liveNodes() {
+                        return Bits.NONE;
```
ashkrisk: nit: I'd prefer using Bits.ALL (although it theoretically shouldn't matter)

jvector-tests/src/test/java/io/github/jbellis/jvector/TestUtil.java
```diff
@@ -543,4 +543,107 @@ public long ramBytesUsed() {
...
+                    @Override
+                    public boolean contains(int level, int node) {
+                        return false;
+                    }
+                };
+            }
+
+            @Override
+            public int maxDegree() {
+                return 0;
```
ashkrisk: Let's return a valid maxDegree instead of defaulting to zero. Similarly for maxDegrees, etc.

ashkrisk (commented):
@reta the new test for for the empty OnDiskGraphIndex looks good. I think the earlier failure I was seeing had to do with improper file cleanup on my system, I sorted that out and it runs fine now.

My remaining comments are mostly nitpicks. There are a lot of lines consisting of multiple spaces followed by a newline (Bob has a tendency to preserve indentation for empty lines), it would be better if you can remove those as well.

reta:
> My remaining comments are mostly nitpicks. There are a lot of lines consisting of multiple spaces followed by a newline (Bob has a tendency to preserve indentation for empty lines), it would be better if you can remove those as well.

Thanks @ashkrisk , really appreciate it, I think we should be all set now

ashkrisk:
@reta we just merged #688 which fixes the CI issues for PRs from forks. Can you please rebase on the latest main?

reta:
> @reta we just merged #688 which fixes the CI issues for PRs from forks. Can you please rebase on the latest main?

Done, thanks a lot @ashkrisk !

reta:
@ashkrisk thanks again for the review, anything holding this change off? we really need it to fix a bug on opensearch-jvector and further on OpenRAG part, thank you

jshook (approved):
I saw no obvious issues with this, and tests look sound

MarkWolters (approved):
This PR updates the graph writing process to write a `-1` as entry node ordinal as a flag that the graph is empty. If empty graphs are written and then re-loaded by a previous version of jvector, eg for a rollback, the graph will be loaded as a valid graph, but the header correctly indicates size of 0. Searches against the graph using the OnDiskGraphIndex View do not cause failures. This seems like a reasonable approach.

ashkrisk (approved):

PR #692, opened 2026-07-06
fix for disk usage measurements

MarkWolters:
## Problem 1: The index cache directory accumulates across runs

  In Grid.runOneGraph (and runAllAndCollectResults), BenchmarkDiagnostics monitors two directories:
  diagnostics.startMonitoring("testDirectory", workDirectory);
  diagnostics.startMonitoring("indexCache", Paths.get(indexCacheDir));  // "index_cache/"
  getTotalBytes() sums both. The index_cache/ directory is persistent — it is never wiped between runs. So the reported disk figure grows monotonically across the run, regardless of which configuration is under test. The delta
  between a fused-PQ run and a non-fused-PQ run is invisible.

## Problem 2: Even within a single run, the snapshot captures all feature sets, not just one

  When multiple feature sets are built in the same runOneGraph call, the directory total covers all of them. You can't attribute disk usage to a specific configuration from that number.

  There is even a TODO in the code acknowledging this at line 233:
  // TODO this does not capture disk usage for cached indexes. Need to update

  ---
 ## The fix

  Rather than directory-level monitoring, measure the specific graph file for each feature set directly after it is written. In buildOnDisk, the path for each feature set's graph file is already computed:

  if (handles.containsKey(features)) {
      graphPath = handles.get(features).writePath();
  } else {
      graphPath = outputDir.resolve("graph" + n++);
  }

  After the build completes, Files.size(graphPath) gives the exact on-disk size for that specific configuration. This value should be returned from buildOnDisk (as part of a result map keyed by feature set) and then included
  in the BenchResult params or metrics map alongside the other per-configuration numbers.

### DIFF
--- a/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
+++ b/jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
@@ -230,7 +230,6 @@ public class Grid {
         // Prepare to collect index construction metrics for reporting....
         var constructionMetrics = new ConstructionMetrics();
 
-        // TODO this does not capture disk usage for cached indexes.  Need to update
         // Capture initial memory and disk state
         try (var diagnostics = new BenchmarkDiagnostics(getDiagnosticLevel())) {
             diagnostics.startMonitoring("testDirectory", workDirectory);
@@ -253,6 +252,7 @@ public class Grid {
                     (buildCompressorObj == null) ? "None" : String.valueOf(buildCompressorObj);
 
             Map<Set<FeatureId>, ImmutableGraphIndex> indexes = new HashMap<>();
+            Map<Set<FeatureId>, Long> indexFileSizes = new HashMap<>();
             if (buildCompressorObj == null) {
                 indexes = buildInMemory(featureSets, M, efConstruction, neighborOverflow, addHierarchy, refineFinalGraph, ds, workDirectory);
             } else {
@@ -270,6 +270,7 @@ public class Grid {
                     if (cached.isPresent()) {
                         System.out.printf("%s: Using cached graph index for %s%n", key.datasetName, fs);
                         indexes.put(fs, cached.get());
+                        try { indexFileSizes.put(fs, Files.size(cache.resolve(key).finalPath)); } catch (IOException ignored) {}
                     } else {
                         missing.add(fs);
                         if (cache.isEnabled()) {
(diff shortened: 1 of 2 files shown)


jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -1161,9 +1168,19 @@ private static String quantTypeOf(CompressorParameters cp) {
      * <p>Call {@link #appendTo(List)} to emit these as {@link Metric} entries so they can be selected for
      * console/CSV reporting.</p>
      */
+    private static final class BuildOnDiskResult {
```
ashkrisk: The Javadoc above refers to the wrong class now that BuildOnDiskResult has been inserted here. I'd also suggest moving this definition up to before the buildOnDisk methods are defined.

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
```diff
@@ -914,7 +922,6 @@ public static List<BenchResult> runAllAndCollectResults(
                                             diagnostics.capturePostPhaseSnapshot("Build");
                                             diagnostics.printDiskStatistics("Graph Index Build");
                                             var buildSnapshot = diagnostics.getLatestSystemSnapshot();
-                                            DiskUsageMonitor.MultiDirectorySnapshot buildDiskSnapshot = diagnostics.getLatestDiskSnapshot();
```
ashkrisk, line 917: May also want to remove the `DiskUsageMonitor` import

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
MarkWolters: Done, thanks for catching that

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
MarkWolters: Done

jvector-examples/src/main/java/io/github/jbellis/jvector/example/Grid.java
ashkrisk: The positioning issue is now fixed, but looks like the original Javadoc for ConstructionMetrics looks to have been accidentally removed along with it

MarkWolters: restored ConstructionMetrics javadoc

ashkrisk (approved):

dian-lun-lin (approved):

PR #693, opened 2026-07-07
Expose the NVQ getter

akash-shankaran:
In OpenSearch, we plan to make use of the NVQ Feature presence, if NVQ vectors need to be dequantized back to float. The current technique is using reflection, but it would be better to expose the getter which will make the implementation cleaner. 

https://github.com/opensearch-project/opensearch-jvector/pull/539/changes/BASE..1122f0d283d441efa697e72d2d0a9311b09e3820#diff-76d766440abecf3c20679bf1b8eb51eead85a9363a60478703aedcadf4d4f73dR417


MarkWolters (approved):
LGTM

reta (approved):

r-devulap (approved):
Seems simple enough. LGTM.

PR #694, opened 2026-07-08
Add FloatArray interface to allow easy float[] vector extraction from MemorySegmentVectorFloat and ArrayVectorFloat instances

### LINKED ISSUE #696: Zero-copy access to VectorFloat<?> vectors backed by float[] arrays
## Problem

At the moment,  the general patterns to extract `float[]` arrays from `VectorFloat<?>` instances are as such:
 - call `length()` + `get(index)` to have a full copy, using per-element access
 - call `VectorFloat<?>::get()` and check if it instance of `float[]`  (covered by `ArrayVectorFloat`)
 - call `VectorFloat<?>::get()` and check if it instance of `MemorySegment` , use reflection (since API is only available on JDK-22) to access `heapBase()` or `ValueLayout` (see please https://github.com/opensearch-project/opensearch-jvector/pull/507/changes#diff-af8a9e9960c5820cdc11161a01241bc490fb4aa17184c0e6b5397ae10af85307R18 as an example)

The reality is that all **current** implementations of `VectorFloat<?>` (but not necessarily future ones) are backed by  `float[]` and could provide fast, zero-copy access to it.

### New API, opt-in only

Add `FloatArray` interface to allow easy `float[]` vector extraction from `MemorySegmentVectorFloat` and `ArrayVectorFloat` instances. It solves the problem by directly exposing the backed array through `float[] array()` method, no copying. This is opt-in only, the change is 100% non-breaking. 

See please: https://github.com/datastax/jvector/pull/694

### Change existing API

Add `toArray` method to existing `VectorFloat<?>` API. This is simpler change (since we could provide the default implementation using `length()` + `get(index)`) but we cannot guarantee that every `toArray` implementation is going to be zero copy, although we could document that.

See please: https://github.com/datastax/jvector/pull/697

CC @ashkrisk @akash-shankaran

reta:
Add `FloatArray` interface to allow easy `float[]` vector extraction from `MemorySegmentVectorFloat` and `ArrayVectorFloat` instances. It solves the problem that we have to: 
 - use explicit typecast to known types (float[], MemorySegment) to figure out what kind of the holder object `VectorFloat` has
 - deal with reflection (MemorySegment available in JDK-22 but OpenSearch baseline is JDK-21) or per-element access semantics which is inefficient

The suggestion is to introduce new `FloatArray` interface to have a fast track `float[]` array conversion for `VectorFloat`s that support it. The change is 100% non-breaking.

Closes https://github.com/datastax/jvector/issues/696

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/FloatArray.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/FloatArray.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.github.jbellis.jvector.vector.types;
+
+/**
+ * Return the {@code float[]} representation of the {@link VectorFloat} if its type supports
+ * such conversions.
+ */
+public interface FloatArray {
+    float[] array();
+}
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/FloatArray.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/FloatArray.java
@@ -19,6 +19,8 @@ package io.github.jbellis.jvector.vector.types;
 /**
  * Return the {@code float[]} representation of the {@link VectorFloat} if its type supports
  * such conversions.
+ *
+ * @apiNote this is an experimental API and may change in the future releases.
  */
 public interface FloatArray {
     float[] array();
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorFloat.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorFloat.java
@@ -18,6 +18,7 @@ package io.github.jbellis.jvector.vector;
 
 import io.github.jbellis.jvector.disk.IndexWriter;
 import io.github.jbellis.jvector.util.RamUsageEstimator;
+import io.github.jbellis.jvector.vector.types.FloatArray;
 import io.github.jbellis.jvector.vector.types.VectorFloat;
 
 import java.io.IOException;
@@ -26,7 +27,7 @@ import java.util.Arrays;
 /**
  * VectorFloat implementation backed by an on-heap float array.
  */
-final public class ArrayVectorFloat implements VectorFloat<float[]>
+final public class ArrayVectorFloat implements VectorFloat<float[]>, FloatArray
 {
     private final float[] data;
 
@@ -122,6 +123,11 @@ final public class ArrayVectorFloat implements VectorFloat<float[]>
     public int hashCode()
     {
         return this.getHashCode();
+    }
+    
+    @Override
+    public float[] array() {
+        return get();
     }
 }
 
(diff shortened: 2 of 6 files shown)


r-devulap (commented):
Curious about the need for introducing a separate `FloatArray` interface here. Could we instead add an abstract method on `VectorFloat<?>` itself similar to the `array()` method introduced in this PR and require implementations like `ArrayVectorFloat` and `MemorySegmentVectorFloat` to provide it? That seems like it might keep the API simpler by avoiding an extra type just for array extraction.

reta:
> Curious about the need for introducing a separate `FloatArray` interface here. Could we instead add an abstract method on `VectorFloat<?>` itself similar to the `array()` method introduced in this PR and require implementations like `ArrayVectorFloat` and `MemorySegmentVectorFloat` to provide it?

If we don't care about compatibility much (since this classes are public and if by chance there are implementations outside the library), than yes - we could certainly do that. It will definitely make the pull request simpler. 

What we could do is to get rid of `FloatArray` is something like that (that would not break compatibility):

```
default float[] array() {
    throw new UnsupportedOperationException();
}
```

But again, if compatibility is not a concern - we could get rid of `FloatArray`, thanks @r-devulap !

r-devulap:
I will let @ashkrisk weigh in on it, he is more qualified to answer that :)

ashkrisk (changes requested):
While there are legitimate, performance related reasons to want direct access to the underlying `float[]` store, the API as it stands is not designed with that in mind. IMO either of the proposed approaches breaks the encapsulation of the `VectorFloat` interface in potentially undesirable ways.

At the moment I can only see [one place](https://github.com/opensearch-project/opensearch-jvector/blob/8dc00bf3892b325df78d1eed44fb6246c697e5e6/src/main/java/org/opensearch/knn/index/codec/jvector/JVectorFloatVectorValues.java#L152) where Opensearch accesses the underlying array. Another solution that might work here would be to add a `toArray()` method that works on all `VectorFloat` types and returns a copy of its underlying data.

@reta can you open an issue explaining your use-case in more detail? With more information we can figure out what the best solution is.

reta:
> While there are legitimate, performance related reasons to want direct access to the underlying `float[]` store, the API as it stands is not designed with that in mind. IMO either of the proposed approaches breaks the encapsulation of the `VectorFloat` interface in potentially undesirable ways.

@ashkrisk this is exactly the reason why VectorFloat **is not changed**, there is a new interface that adds this functionality to implementing classes that could be implemented.

> @reta can you open an issue explaining your use-case in more detail? With more information we can figure out what the best solution is.

See please this pull request https://github.com/opensearch-project/opensearch-jvector/pull/507/changes#diff-ddf73468c0184ab1f6ef7973480de2cbdf98d8eb63d0aec938ec4dd6b5d6e73cR153

reta:
> At the moment I can only see [one place](https://github.com/opensearch-project/opensearch-jvector/blob/8dc00bf3892b325df78d1eed44fb6246c697e5e6/src/main/java/org/opensearch/knn/index/codec/jvector/JVectorFloatVectorValues.java#L152) where Opensearch accesses the underlying array. Another solution that might work here would be to add a `toArray()` method that works on all `VectorFloat` types and returns a copy of its underlying data.

@ashkrisk exploring your idea here https://github.com/datastax/jvector/pull/697, thanks!

reta:
@ashkrisk anything pending (post our offline discussion) to move this one forward? we would really benefit from it on `opensearch-jvector` plugin side, thank you

zznate:
Hey folks (@ashkrisk and @tlwillke) - anything we can do to move this forward? We have some opensearch patches in flight depending on this. Happy to chip in if needed.

jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/FloatArray.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package io.github.jbellis.jvector.vector.types;
+
+/**
+ * Return the {@code float[]} representation of the {@link VectorFloat} if its type supports
+ * such conversions.
+ *
+ * @apiNote this is an experimental API and may change in the future releases.
+ */
```
MarkWolters, line 24: please add unit testing around this call and I will approve

reta: thanks @MarkWolters , tests added (apologies for delay, was out on vacation)

MarkWolters (approved):

r-devulap (approved):
LGTM.

ashkrisk (approved):

PR #695, opened 2026-07-09
Parallel Write performance improvements

MarkWolters:
This PR updates the graph parallel write process to take full advantage of the async write process. Whereas before although the write tasks were batched and handed off to achieve better performance they were not truly asynchronous. This update changes that to allow fully async graph index writes and increases the benefit of using parallel writes significantly. As an example prior to this update graph write time showed speedups of 4x - 8x using parallel writes
```
=== Testing Write Performance ===
Writing with NVQ + FUSED_ADC features...
Sequential write: 6074.18 ms
Parallel write:   1373.73 ms
Speedup:          4.42x
```
After this update the speedup is more along the lines of 24x - 32x
```
=== Testing Write Performance ===
Writing with NVQ + FUSED_ADC features...
Sequential write: 20147.29 ms
Parallel write:   627.52 ms
Speedup:          32.11x
```

This PR also introduces the ability to specify parallel writes as part of the test configuration through yaml files which is standard for BenchYAML and AutoBenchYAML test. The new parameter `parallelGraphConstruction` has been added to `ConstructionParameters` as a Boolean value. If set to `Yes` Grid will use parallel writes in graph construction. If set to `No` Grid will use the legacy serial writing pattern. If the parameter is omitted it will default to the legacy serial writes during graph construction.

Regression test results below. Note that the difference in index construction is relatively minor due to the pattern of using writeInline to write the features prior to the main call to write the vectors necessitating the avoidance of overwriting the existing feature data and limiting the capabilities of constructing the record in entirety before persisting to disk. In the future based on the plan to eventually deprecate this pattern the gains are in line with the statistics above.

Index_Construction_Time


docs/release notes/4.1.0/695.performance.md
```diff
@@ -0,0 +1,86 @@
+### Parallel Graph Index Writing via AsynchronousFileChannel
+
+**Description**
+Introduces `OnDiskParallelGraphIndexWriter`, a new writer that parallelizes the serialization
```
r-devulap: `OnDiskParallelGraphIndexWriter` is already in the main branch. Shouldn't this be an enhancement to the class rather introducing a new class?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -75,128 +97,240 @@ class NodeRecordTask implements Callable<Void> {
...
+        // allocated that is a no-op, but it sets initialPosition = 0 as required.
+        var writer = new ByteBufferIndexWriter(rangeBuffer);
+
+        for (int newOrdinal = startOrdinal; newOrdinal < endOrdinal; newOrdinal++) {
+            buildFullRecord(writer, newOrdinal);
+        }
+
+        // One channel.write() for the entire task range — one syscall, one OS I/O request.
+        rangeBuffer.flip();
+        channel.write(rangeBuffer, baseOffset + (long) startOrdinal * recordSize).get();
```
r-devulap: There is a comment in the main branch that says: `This is critical for correctness as AsynchronousFileChannel.write() may not write all bytes in one call.`. Doesn't this apply here as well? This `.write()` need not complete writing all the bytes? The previous code had a loop to account for this: 

```
while (buffer.hasRemaining()) {
            int written = channel.write(buffer, currentPosition).get();
            .... 
        }
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
```diff
@@ -75,128 +97,240 @@ class NodeRecordTask implements Callable<Void> {
...
+            // called for them, so the entire record is ours — write it in one shot.
+            ByteBuffer buf = allocRegion(recordSize);
+            buf.putInt(newOrdinal);
+            for (var feature : inlineFeatures) {
+                for (int i = 0; i < feature.featureSize(); i++) buf.put((byte) 0);
+            }
+            buf.putInt(0); // neighbor count
+            for (int n = 0; n < graph.getDegree(0); n++) buf.putInt(-1);
+            buf.flip();
+            pending.add(channel.write(buf, recordBase));
```
r-devulap: Same thing here: do we need to account for partial writes?

docs/release notes/4.1.0/695.performance.md
MarkWolters: Updated in latest checkin

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: You are correct that the comments originally stated that the blocking `Future.get()` calls were necessary this proved to be a misinterpretation of an error condition that occurred during testing and the comments were not correctly updated to reflect this. In the original implementation the asynchronicity of this class was essentially broken by blocking on the `Future.get()` unnecessarily because of an interpretation that the `AsynchronousFileChannel.write()` call was not writing all bytes causing data corruption. However the root cause of the issue was later determined to be a bug in the `ByteBufferIndexWriter` implementation which was resolved by PR 607 (https://github.com/datastax/jvector/pull/607). The original comments never got cleaned up, but should be removed with this PR.

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
MarkWolters: see previous comment above

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/NodeRecordTask.java
r-devulap: Looking up the JAVA doc of  `AsynchronousByteChannel` (which is supposed to work the same way as `AsynchronousFileChannel`):  https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/nio/channels/AsynchronousByteChannel.html: 


_This method initiates an asynchronous write operation to write a sequence of bytes to this channel from the given buffer. The handler parameter is a completion handler that is invoked when the write operation completes (or fails). The result passed to the completion handler is the number of bytes written.
The write operation may write up to r bytes to the channel, where r is the number of bytes remaining in the buffer, that is, src.remaining() at the time that the write is attempted._


 `Future.get()` returns the number of bytes written which may or may not match `src.remaining()` because **_it may write up to `r` bytes_**. Don't we need to check that all the bytes were written before proceeding?

MarkWolters: The situation where `<r` bytes are written would be extremely rare. AsynchronousFileChannel issues blocking pwrite() calls under the hood and the only time that should fail to write the full amount would be if the disk fills up partway through the write, if the filesystem quota limits are hit partway through or if the write exceeds the per-syscall write cap which is around 2G. Under normal circumstances that should never occur.

That said "extremely rare" is still a non-zero possibility so you have a point. If one of those circumstances were to occur it would basically be a silent failure which is no good. I'll push an update today to address this.

docs/release notes/4.1.0/695.performance.md
```diff
@@ -0,0 +1,98 @@
...
+  each task now allocates its own range-sized (fast path) or per-region (legacy path)
+  buffer instead.
+
+**Also in this PR:** a `parallelGraphConstruction` boolean was added to
+`ConstructionParameters`, letting BenchYAML / AutoBenchYAML test configs opt into
+`OnDiskParallelGraphIndexWriter` for index construction instead of the default serial
+`OnDiskGraphIndexWriter`, without any code changes.
+
+**Performance**
+Example run writing with NVQ + FUSED_ADC features. Before this change, parallel writes
```
jshook, line 30: If we include benchmark data in docs, we need to qualify the test scenario with system specs:
CPU/RAM/Storage type and layout

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
```diff
@@ -163,23 +166,17 @@ public ParallelGraphWriter(RandomAccessWriter writer,
             + Integer.BYTES // neighbor count
             + graph.getDegree(0) * Integer.BYTES; // neighbors + padding
 
-        // Thread-local views for safe neighbor iteration
-        // CopyOnWriteArrayList handles concurrent additions safely
+        // Thread-local views for safe neighbor iteration.
+        // CopyOnWriteArrayList handles concurrent additions safely.
         this.viewPerThread = ThreadLocal.withInitial(() -> {
```
jshook, line 171: Although it can be a twist, we should start trying to remove usage of ThreadLocal where possible. Of course, this pushes us closer to explicit task structuring, but there are caveats for virtual threads with ThreadLocal which re not ideal.

jshook: (ah noticed the change was for comments only, sorry if superfluous)

jshook (commented):
The filesystem used for testing here is not obvious. I would like to see the tests run on a small handful of fileystems at least, and this might require us to improve our GHA runner configs. It would be good to have some stabilizing performance data to compare with later, though.

skrcode:
@MarkWolters — I ran JAIPilot Cloud against this exact head (1f24c985). It produced an AI-generated draft companion that adds omitted-ordinal coverage and consolidates duplicated NodeRecordTask/Grid paths: https://github.com/skrcode/jvector/pull/2

Cloud evidence: the same 6 focused tests passed before and after; the full 5-module reactor passed 204/204 tests. This is a review aid only; nothing is auto-merged.

docs/release notes/4.1.0/695.performance.md
MarkWolters: Added above

jvector-base/src/main/java/io/github/jbellis/jvector/graph/disk/ParallelGraphWriter.java
MarkWolters: I'm going to close this as it refers to a comment rather than any code change. Please feel free to reopen if you feel it should be addressed in this pr

PR #697, opened 2026-07-10
Add toArray method to existing VectorFloat<?> to access underlying float[] vectors representation

### LINKED ISSUE #696: Zero-copy access to VectorFloat<?> vectors backed by float[] arrays

reta:
Add `toArray `method to existing `VectorFloat<?>`API. This is simpler change (since we could provide the default implementation using `length()` + `get(index)` but we cannot guarantee that every `toArray` implementation is going to be zero copy, although we could document that.

Closes https://github.com/datastax/jvector/issues/696

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorFloat.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ArrayVectorFloat.java
@@ -123,5 +123,10 @@ final public class ArrayVectorFloat implements VectorFloat<float[]>
     {
         return this.getHashCode();
     }
+
+    @Override
+    public float[] toArray() {
+        return get();
+    }
 }
 
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/VectorFloat.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/types/VectorFloat.java
@@ -55,4 +55,12 @@ public interface VectorFloat<T> extends Accountable
         }
         return result;
     }
+    
+    default float[] toArray() {
+        final float[] arr = new float[length()];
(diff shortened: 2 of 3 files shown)


reta:
Closing in favour of https://github.com/datastax/jvector/pull/694 as discussed with @ashkrisk offline

PR #703, opened 2026-07-24
Admin interface

MarkWolters:
## JMX Runtime Configuration for Graph Index Builder

Introduces `GraphIndexBuilderConfig`, a JMX-managed singleton that exposes `GraphIndexBuilder` construction parameters as runtime-tunable attributes. Before this change, options such as `addHierarchy`, `refineFinalGraph`, `parallelBuild`, and build-time compression type could only be set at the call site and required a code change or restart to modify.

**What's new**

- **`management` package** — `GraphIndexBuilderConfigMBean` declares 8 managed attributes across four groups (graph topology, write path, build compression type, PQ parameters); `GraphIndexBuilderConfig` is the singleton implementation registered under `io.github.jbellis.jvector:type=GraphIndexBuilderConfig`. `CompressionType` enum (`NONE`, `PQ`, `BQ`) drives build-time scoring compression selection.

- **`GraphIndexBuilder`** — New constructor `(RandomAccessVectorValues, VectorSimilarityFunction, …)` reads `addHierarchy`, `refineFinalGraph`, and compression type from `GraphIndexBuilderConfig` at construction time. Existing constructors that accept those booleans explicitly are deprecated but unchanged in behavior. The new `getBuildScoreProvider` helper wires PQ and BQ build-time compression end-to-end using the JMX-configured parameters.

- **`RandomAccessOnDiskGraphIndexWriter.Builder`** — Unified builder replaces the separate `OnDiskGraphIndexWriter.Builder`. Reads `GraphIndexBuilderConfig.isParallelBuild()` at build time to select `OnDiskParallelGraphIndexWriter` or `OnDiskGraphIndexWriter`. Both produce an identical on-disk format, so the flag can be toggled without re-indexing.

- **No breaking changes** — All existing API call sites continue to compile and behave identically; deprecated constructors are retained and honor caller-supplied values.

**Attributes managed via JMX**

| Attribute | Type | Default |
|---|---|---|
| `AddHierarchy` | boolean | true |
| `RefineFinalGraph` | boolean | true |
| `ParallelBuild` | boolean | false |
| `BuildCompressionType` | String | `"NONE"` |
| `PqMFactor` | int | 8 |
| `PqK` | int | 256 |
| `PqCenterData` | boolean | false |
| `PqAnisotropicThreshold` | float | -1.0 |

All attributes are `@Experimental`. MBean registration is best-effort — failure logs a warning but does not affect normal operation.


jshook (commented):
I don't mind JMX being added for these things but I do want to see there be a better way. As in if JMX is supported, it should probably be based on a better internal interface.
If you are open to this, we can discuss before pushing forward on this change.

MarkWolters:
> I don't mind JMX being added for these things but I do want to see there be a better way. As in if JMX is supported, it should probably be based on a better internal interface. If you are open to this, we can discuss before pushing forward on this change.

Updated PR to include flexible management registry allowing for other implementations but keeping jmx as the default

jshook (approved):
Looks good

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -296,6 +458,55 @@ public GraphIndexBuilder(BuildScoreProvider scoreProvider,
...
+                             int dimension,
+                             List<Integer> maxDegrees,
+                             int beamWidth,
+                             float neighborOverflow,
+                             float alpha,
+                             ForkJoinPool simdExecutor,
+                             ForkJoinPool parallelExecutor) {
+        this(scoreProvider, dimension, maxDegrees, beamWidth, neighborOverflow, alpha,
+             resolveJmxAddHierarchy(maxDegrees),
+             resolveJmxRefineFinalGraph(),
```
ashkrisk: I'm not a fan of having constructors that depend on global state. It might be better if the code paths with managed parameters are cleanly separated out.

The simplest example of this would be to have static factory methods which clearly indicate that managed state is being used, for example:
```java
public GraphIndexBuilder ofManaged(... /* arguments except addHierarchy and refineFinalGraph */) { ... }
```

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
```diff
@@ -266,6 +400,31 @@ public GraphIndexBuilder(BuildScoreProvider scoreProvider,
...
+     * @param maxDegrees       the maximum number of connections a node can have in each layer; if fewer entries
+     *      *                  are specified than the number of layers, the last entry is used for all remaining layers.
+     * @param beamWidth        the size of the beam search to use when finding nearest neighbors.
+     * @param neighborOverflow the ratio of extra neighbors to allow temporarily when inserting a
+     *                         node. larger values will build more efficiently, but use more memory.
+     * @param alpha            how aggressive pruning diverse neighbors should be.  Set alpha &gt; 1.0 to
+     *                         allow longer edges.  If alpha = 1.0 then the equivalent of the lowest level of
+     *                         an HNSW graph will be created, which is usually not what you want.
+     */
+    public GraphIndexBuilder(BuildScoreProvider scoreProvider,
```
ashkrisk: There are a lot of new constructors, should we use the builder pattern or similar to cut down on constructor spam?

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
MarkWolters: updated to remove the new constructors and replaced them with a fluent Builder

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
MarkWolters: updated to remove the new constructors and replaced them with a fluent Builder

jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphIndexBuilder.java
ashkrisk: The new version will still silently pick `addHierarchy` and `refineFinalGraph` from global state with seemingly no way to opt-out without using deprecated constructors.

Why not include methods like `withAddHierearchy(boolean)` and `withRefineFinalGraph(boolean)` so users can opt out?

I would like it even more if managed state were strictly opt-in, i.e. JMX provided values would not be read unless the embedder explicitly requests them by calling `.withManagedParameters(true)` on the builder object, and otherwise using standard defaults like the existing constructors do.

MarkWolters: because we don't want them to opt in or opt out. We are taking the decision out of their hands, the deprecated constructors are there so as not to break existing customer code but the impetus behind this PR is to move control of something like hierarchy, which should be used across the board, out of the customers hands and put it in control of the jvector library.

ashkrisk (approved):

PR #708, opened 2026-08-03
Add native SIMD testing and benchmark module

r-devulap:
This pull request adds comprehensive native C++ unit tests and benchmarks for the SIMD kernel dispatcher and vector similarity functions, and integrates them into the build and CI pipeline. The most important changes are:

**Testing and Benchmarking Enhancements:**

* Adds a new `test_simd_kernels` executable using Google Test, with tests for ISA detection and SIMD kernel correctness, and integrates it into the Meson build and GitHub Actions CI workflow. 
* Adds a new `bench_simd_kernels` executable using Google Benchmark to micro-benchmark `cosine_f32`, `dot_product_f32`, and `euclidean_f32` functions at realistic embedding dimensions. 

**Build System and Code Organization:**

* Moves the core SIMD kernel and dispatcher source files into a src/ subdirectory and updates the Meson build rules and include paths accordingly. The native code layout has been reorganized in preparation for modularizing it into a standalone library, improving structure and maintainability.

**CI Improvements:**

* Updates the GitHub Actions workflow to install `libgtest-dev`, build, and run the new native C++ tests for all relevant ISA tiers, ensuring dispatcher correctness across hardware.

PR #709, opened 2026-08-11
Add int8 / byte-vector similarity support

### LINKED ISSUE #665: Native INT8 (byte vector) HNSW build + search API
Hi! Filing this from the [ArcadeDB](https://github.com/ArcadeData/arcadedb) project, which uses JVector as the HNSW backend for its `LSM_VECTOR` index.

## Background

We're adding pre-quantized int8 ingest to ArcadeDB ([ArcadeData/arcadedb#4132](https://github.com/ArcadeData/arcadedb/issues/4132)) so callers using providers that emit int8 directly (Cohere `embed-english-v3.0`, OpenAI `text-embedding-3-large` reduced precision, Sentence Transformers with int8 quantization) can skip a precision-losing client-side `int8 → float32` round-trip.

We dug into JVector 4.0.0-rc.8 to wire the path through and found the HNSW graph API operates on `VectorFloat<?>` end-to-end:

- `RandomAccessVectorValues.getVector(int ordinal)` returns `VectorFloat<?>`.
- `GraphIndexBuilder` constructors take `RandomAccessVectorValues` (float-only).
- `VectorSimilarityFunction.compare(VectorFloat<?>, VectorFloat<?>)` - the abstract method signature.

So a caller with int8 input must dequantize to float32 for graph build *and* for every query, even when the application semantics are int8-throughout. `ByteSequence<?>` exists in the type system but is used only for PQ/BQ codes (sidecar against the float-vector graph), not as a primary HNSW vector type.

## Ask

A native byte (int8) vector path:

- A `RandomAccessByteVectorValues` (or a generalised `RandomAccessVectorValues<T>`).
- `VectorSimilarityFunction` overloads / variants for `byte[]` (cosine + dot product on byte vectors with per-block min/max calibration; euclidean on bytes is also straightforward).
- `GraphIndexBuilder` constructor(s) that accept the byte-vector RAVV + byte similarity function.
- Search-side equivalent in `GraphSearcher`.

## Why it matters

Modern embedding providers default to int8/binary outputs at scale - Cohere binary embeddings are 1/32× the size of float32, Cohere int8 is 1/4×. Forcing dequantize-on-build/search means:

- **Build cost**: O(N * dim * 4) bytes of transient float32 even though the source is bytes.
- **Search cost**: every query vector dequantises before comparison - JVector's SIMD intrinsics for the byte-similarity case never get exercised.
- **Storage cost**: applications keep bytes in their primary store but JVector wants floats, so RAM and on-disk size grow 4× beyond what the application needs.

Lucene 9.x added `VectorEncoding.BYTE` for similar reasons; we'd love the same in JVector to close the precision/size loop.

## ArcadeDB context

For reference, our matrix vs Qdrant / Milvus 2.5 ([docs/arcadedb-vs-leading-vector-dbms.md](https://github.com/ArcadeData/arcadedb/blob/main/docs/arcadedb-vs-leading-vector-dbms.md)) flags pre-quantized ingest as a P2 gap. We can ship an MVP that dequantizes int8 → float32 server-side (covered in [ArcadeData/arcadedb#4132](https://github.com/ArcadeData/arcadedb/issues/4132)) and that closes the API ergonomics gap, but the full "end-to-end int8 with no float32 transient" win requires the JVector-side support.

Happy to contribute if there's a design direction the maintainers are considering. Otherwise, this is a tracking request.

Thanks!

r-devulap:
### Overview
This branch adds full **INT8 (byte-vector) support** to JVector — from the math primitives all the way through graph building, on-disk storage, SIMD acceleration, and end-to-end tests.

---

#### 1. fd131a7 — Add random-access int8 vector value support
Introduces a new `VectorValues` super-interface that captures shared random-access behavior, then refactors `RandomAccessVectorValues` (float) to extend it, and adds `RandomAccessByteVectorValues` and its list-backed implementation `ListRandomAccessByteVectorValues` as the byte-vector counterpart. Also updates `BuildScoreProvider` to use the generalized thread-local supplier.

#### 2. c55574c — Add byte-similarity methods to VectorUtilSupport / VectorUtil / DefaultVectorUtilSupport
Adds `dotProduct`, `squaredL2`, and `cosine` for `byte[]` / `ByteSequence` to the vector utility stack.

#### 3. 9ba737a — Add ByteVectorSimilarityFunction enum
Mirrors the float `VectorSimilarityFunction` enum; wraps the three byte similarity metrics with a `compare()` API and score normalization.

#### 4. 5652b46 — Add BuildScoreProvider.byteVectorScoreProvider and searchProviderFor(ByteSequence)
Adds `byteVectorScoreProvider()` factory and `searchProviderFor(ByteSequence)` so the graph builder can score byte-vector candidates.

#### 5. 389c391 — Wire byteVectorScoreProvider into GraphIndexBuilder
`GraphIndexBuilder` now accepts `RandomAccessByteVectorValues` and routes to the byte score provider, enabling full graph construction over int8 data.

#### 6. c6dfea0 — Vectorize ByteSequence similarity metrics with SIMD
Panama SIMD and AVX-512 native kernels for int8 dot-product, L2, and cosine; adds native benchmarks and C++ unit tests.

#### 7. 3d4dada — Add InlineByteVectors feature for native int8 on-disk storage
New `FeatureId` and `InlineByteVectors` class for serializing byte vectors into the graph index file format, plus `OnDiskGraphIndex` reader support.

#### 8. 974f508 — Add INT8 end-to-end tutorial (Int8Example)
End-to-end walkthrough showing how to build, serialize, and query a byte-vector graph index; registered in `TutorialRunner`.

#### 9. b0ad5cd — Add int8 test coverage for byte-vector similarity, RABVV, BSP, SIMD, graph build, and disk round-trip
Unit tests for `ByteVectorSimilarityFunction`, `BuildScoreProvider`, `TestVectorizationProvider` (SIMD), `TestVectorGraph` (build), `TestOnDiskGraphIndex` (round-trip), and `ListRandomAccessByteVectorValues`.

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
@@ -25,6 +25,7 @@ import io.github.jbellis.jvector.vector.VectorUtil;
 import io.github.jbellis.jvector.vector.VectorizationProvider;
 import io.github.jbellis.jvector.vector.types.VectorFloat;
 import io.github.jbellis.jvector.vector.types.VectorTypeSupport;
+import java.util.function.Supplier;
 
 /**
  * Encapsulates comparing node distances for GraphIndexBuilder.
@@ -106,8 +107,10 @@ public interface BuildScoreProvider {
     static BuildScoreProvider randomAccessScoreProvider(RandomAccessVectorValues ravv, VectorSimilarityFunction similarityFunction) {
         // We need two sources of vectors in order to perform diversity check comparisons without
         // colliding.  ThreadLocalSupplier makes this a no-op if the RAVV is actually un-shared.
-        var vectors = ravv.threadLocalSupplier();
-        var vectorsCopy = ravv.threadLocalSupplier();
+        var vectorsRaw = ravv.threadLocalSupplier();
+        var vectorsCopyRaw = ravv.threadLocalSupplier();
+        Supplier<RandomAccessVectorValues> vectors = () -> (RandomAccessVectorValues) vectorsRaw.get();
+        Supplier<RandomAccessVectorValues> vectorsCopy = () -> (RandomAccessVectorValues) vectorsCopyRaw.get();
 
         return new BuildScoreProvider() {
             @Override
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
@@ -16,13 +16,16 @@
 
 package io.github.jbellis.jvector.graph.similarity;
 
+import io.github.jbellis.jvector.graph.RandomAccessByteVectorValues;
 import io.github.jbellis.jvector.graph.RandomAccessVectorValues;
 import io.github.jbellis.jvector.graph.RemappedRandomAccessVectorValues;
+import io.github.jbellis.jvector.vector.ByteVectorSimilarityFunction;
 import io.github.jbellis.jvector.quantization.BQVectors;
 import io.github.jbellis.jvector.quantization.PQVectors;
 import io.github.jbellis.jvector.vector.VectorSimilarityFunction;
 import io.github.jbellis.jvector.vector.VectorUtil;
 import io.github.jbellis.jvector.vector.VectorizationProvider;
+import io.github.jbellis.jvector.vector.types.ByteSequence;
 import io.github.jbellis.jvector.vector.types.VectorFloat;
 import io.github.jbellis.jvector.vector.types.VectorTypeSupport;
 import java.util.function.Supplier;
@@ -59,6 +62,20 @@ public interface BuildScoreProvider {
      * @param vector the query vector to provide similarity scores against
      */
     SearchScoreProvider searchProviderFor(VectorFloat<?> vector);
+
+    /**
+     * Create a search score provider to use *internally* during construction, for a byte (int8) query vector.
+     * <p>
+     * The default implementation throws {@link UnsupportedOperationException}; only
+     * {@link #byteVectorScoreProvider} overrides it.
+     *
+     * @param vector the int8 query vector to provide similarity scores against
+     */
+    default SearchScoreProvider searchProviderFor(ByteSequence<?> vector) {
+        throw new UnsupportedOperationException(
+                "This BuildScoreProvider does not support byte-vector queries; " +
+                "use byteVectorScoreProvider() to construct a byte-vector builder");
+    }
 
     /**
      * Create a search score provider to use *internally* during construction.
@@ -210,6 +227,68 @@ public interface BuildScoreProvider {
             @Override
             public VectorFloat<?> approximateCentroid() {
                 return pqv.getCompressor().getOrComputeCentroid();
+            }
+        };
+    }
+
+    /**
+     * Returns a BSP that performs exact score comparisons using the given
+     * {@link RandomAccessByteVectorValues} and {@link ByteVectorSimilarityFunction}.
+     * All scoring is byte×byte with no float32 round-trip.
+     */
+    static BuildScoreProvider byteVectorScoreProvider(RandomAccessByteVectorValues ravv, ByteVectorSimilarityFunction bvsf) {
+        var vectors     = ravv.threadLocalSupplier();
+        var vectorsCopy = ravv.threadLocalSupplier();
+
+        return new BuildScoreProvider() {
+            @Override
+            public boolean isExact() {
+                return true;
+            }
+
+            @Override
+            public VectorFloat<?> approximateCentroid() {
+                var vv = vectors.get();
+                var centroid = vts.createFloatVector(vv.dimension());
+                for (int i = 0; i < vv.size(); i++) {
+                    var v = vv.getVector(i);
+                    for (int d = 0; d < vv.dimension(); d++) {
+                        centroid.set(d, centroid.get(d) + v.get(d));
+                    }
+                }
+                VectorUtil.scale(centroid, 1.0f / vv.size());
+                return centroid;
+            }
+
+            @Override
+            public SearchScoreProvider searchProviderFor(VectorFloat<?> vector) {
+                throw new UnsupportedOperationException(
+                        "byteVectorScoreProvider does not support float query vectors; use searchProviderFor(int node)");
+            }
+
+            @Override
+            public SearchScoreProvider searchProviderFor(ByteSequence<?> vector) {
+                var vc = vectorsCopy.get();
+                var sf = (ScoreFunction.ExactScoreFunction) node2 -> bvsf.compare(vector, vc.getVector(node2));
+                return new DefaultSearchScoreProvider(sf);
+            }
+
+            @Override
+            public SearchScoreProvider searchProviderFor(int node1) {
+                var v  = vectors.get().getVector(node1);
+                return searchProviderFor(v);
+            }
+
+            @Override
+            public SearchScoreProvider diversityProviderFor(int node1) {
+                return searchProviderFor(node1);
+            }
+
+            @Override
+            public ScoreFunction diversityScoreFunctionFor(int node1) {
+                var v  = vectors.get().getVector(node1);
+                var vc = vectorsCopy.get();
+                return (ScoreFunction.ExactScoreFunction) node2 -> bvsf.compare(v, vc.getVector(node2));
             }
         };
     }
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ByteVectorSimilarityFunction.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/vector/ByteVectorSimilarityFunction.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright DataStax, Inc.
+ *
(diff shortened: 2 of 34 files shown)


r-devulap:
To do: 

- [x] Update all the docs/README to reflect these changes 
- [ ] Add release notes
- [ ] Add BenchYAML results

jshook (changes requested):
This PR needs to be updated.
1) merge conflicts need to be addressed
2) the base commit this branch depends on should probably determine the target of the merge, rather than main, since there are direct dependencies.

r-devulap:
> This PR needs to be updated.
> 
> 1. merge conflicts need to be addressed

Fixed, it is now rebased with the latest changes in main. 

> 2. the base commit this branch depends on should probably determine the target of the merge, rather than main, since there are direct dependencies.

This is no longer a problem since it includes the commits in #708

r-devulap:
Marking this as draft until I resolve the question of leveraging NVQ for INT8.

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
```diff
@@ -976,14 +976,268 @@ float assembleAndSumPQ_512(
...
+                    .reinterpretAsInts();
+            IntVector vb = fromByteSequence(ByteVector.SPECIES_128, b, bOff + i)
+                    .convertShape(VectorOperators.B2I, IntVector.SPECIES_512, 0)
+                    .reinterpretAsInts();
+            acc = acc.add(va.mul(vb));
+        }
+
+        int result = acc.reduceLanes(VectorOperators.ADD);
+        for (int i = limit; i < length; i++) {
+            result += a.get(aOff + i) * b.get(bOff + i);
```
MarkWolters, line 1023: ArrayByteSequence.slice(offset, length) (ArrayByteSequence.java:75-80) returns an ArraySliceByteSequence wrapping the original array with that offset stored. So when you call .slice(2, 3) on an ArrayByteSequence and then call .get(0), .get(1), .get(2), you get data[2], data[3], data[4] with the offset already resolved. This code calls a.get(aOff + i) where aOff = a.offset(). For a slice, that's a.get(2 + i), which internally becomes data.get(offset + (2 + i)) = data.get(2 + 2 + i) = data.get(4 + i), the offset is being applied twice.

r-devulap: AFAIK, this should always operate on a `ArrayByteSequence` and we are not calling `.slice` on it, right? Or am I missing something?

MarkWolters: It should work because for ArrayByteSequence and MemorySegmentByteSequence offset is always 0 so there's no impact. The point is just that there's no reason to calculate and add offset here, you can just use `.get(i)`

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
```diff
@@ -106,8 +124,10 @@ static BuildScoreProvider randomAccessScoreProvider(RandomAccessVectorValues rav
     static BuildScoreProvider randomAccessScoreProvider(RandomAccessVectorValues ravv, VectorSimilarityFunction similarityFunction) {
         // We need two sources of vectors in order to perform diversity check comparisons without
         // colliding.  ThreadLocalSupplier makes this a no-op if the RAVV is actually un-shared.
-        var vectors = ravv.threadLocalSupplier();
-        var vectorsCopy = ravv.threadLocalSupplier();
+        var vectorsRaw = ravv.threadLocalSupplier();
+        var vectorsCopyRaw = ravv.threadLocalSupplier();
+        Supplier<RandomAccessVectorValues> vectors = () -> (RandomAccessVectorValues) vectorsRaw.get();
```
MarkWolters, line 129: with the split of RAVV into the new VectorValues `threadLocalSupplier()` no longer returns `Supplier<RandomAccessVectorValues>` but instead returns `Supplier<VectorValues<V>>`, making this cast potentially dangerous if someone in the future implements a copy method that returns a `VectorValues<V>` that does not implement RandomAccessVectorValues. This might not be a likely scenario but deserves at least a comment if not a full `instanceof` check here

jvector-base/src/main/java/io/github/jbellis/jvector/vector/ByteVectorSimilarityFunction.java
```diff
@@ -0,0 +1,76 @@
...
+     * Dot product normalised to {@code [0, 1]}.
+     * Raw int8 dot product is divided by {@code n * 127^2} (the maximum possible magnitude)
+     * before applying the {@code (1 + x) / 2} mapping, so the result is always in [0, 1]
+     * regardless of dimension or whether the vectors are unit-norm.
+     * For already unit-norm int8 vectors (e.g. Cohere, OpenAI reduced-precision) prefer {@link #COSINE}.
+     */
+    DOT_PRODUCT {
+        @Override
+        public float compare(ByteSequence<?> v1, ByteSequence<?> v2) {
+            float maxMagnitude = v1.length() * (127.0f * 127.0f);
```
MarkWolters, line 54: The discrepancy between `127 * 127` here and the min/max values of `-128/127` for int8 was bugging me but I couldn't put my finger on why so I asked Bob and here's what it said:
```
The issue is on line 54. The maxMagnitude is computed as n * 127², but the minimum value of a signed byte is −128, not −127. This means the most negative possible dot product is n * (−128 × −128) = n * 128², which is larger than n * 127².

So with two vectors where every component is −128, the raw dot product is n * 16384, but maxMagnitude is only n * 16129. That gives:

(1 + 16384/16129) / 2  ≈  (1 + 1.0158) / 2  ≈  1.008

…which exceeds 1.0 and breaks the guaranteed [0, 1] range.

The fix should use 128 (the absolute maximum magnitude of a signed byte) as the normalisation constant:

// Should be:
float maxMagnitude = v1.length() * (128.0f * 128.0f);

This ensures the denominator covers the worst-case dot product for any valid byte input, keeping the result strictly within [0, 1]. The Javadoc comment on line 46 should also be updated from 127^2 to 128^2 to match.
```
with the caveat
```
your formula is correct if byte vectors are contractually limited to [-127,127]; otherwise the 127² denominator is incorrect. If -128 is legal and all you need is bounded monotonic dot-product similarity, use 128² as the normalization constant.
```

MarkWolters (changes requested):

jvector-twenty/src/main/java/io/github/jbellis/jvector/vector/PanamaVectorUtilSupport.java
r-devulap: yeah makes sense. Fixed in aa21fb3b8fb3cadc3c62202bc37ab8940393dfdb.

jvector-base/src/main/java/io/github/jbellis/jvector/vector/ByteVectorSimilarityFunction.java
r-devulap: ah, right. Thanks for catching it! Fixed in: 3654bd69f4e6fd844f08649eff4464d441e479c4

jvector-base/src/main/java/io/github/jbellis/jvector/graph/similarity/BuildScoreProvider.java
r-devulap: I think a full `instanceof` check is useful. Added in 2964864bf25b960d822788ed27fc2e43fd6726c2

MarkWolters (approved):
Thanks for addressing my requested changes!

tlwillke:
@r-devulap this missing acceptance criteria and evidence that it is met.  It involves new scoring on a new vector type.  We either need the requester(s) to provide some dataset benchmarks and metrics (what is acceptable recall and performance?), or we need to create int8 datasets and set our own standards.  How do you want to proceed?

r-devulap:
>  We either need the requester(s) to provide some dataset benchmarks and metrics (what is acceptable recall and performance?)

@lvca is it possible to provide some datasets that we can you use to measure recall and performance?

lvca:
@tae898 Can you help on this?

tae898:
@r-devulap 

I've been benchmarking ArcadeDB's jvector-backed vector index on SIFT-1M and DEEP-10M, fp32 and int8, for both build and search.

The datasets I've been using are not int8:
- `http://ann-benchmarks.com/sift-128-euclidean.hdf5` — 501 MiB, 1,000,000×128, L2
- `http://ann-benchmarks.com/deep-image-96-angular.hdf5` — 3.58 GiB, 9,990,000×96, angular

So each vector db used their own quantization function to turn fp32 into int8. If you are looking for more "formal" int8 vector dataset, perhaps these two are not the best ones.

Worth knowing what "int8" means on the ArcadeDB side: the index writes an int8 copy of each vector into its own pages, then dequantizes back to float32 on the read path, so jvector always receives float32. That's the gap filed in #665.

My results: https://humem.ai/projects/arcadedb. Those are databases, not libraries, so it isn't a direct comparison with jvector on its own.

Let me know if there's anything else I can help you with.

r-devulap:
@tae898 & @lvca Thanks for your response and patience. We are close to merging this. I will use quantize those datasets to INT8 and add them to our benchYAML pipeline and get some recall/perf numbers for our reference. That should hopefully complete this patch.

PR #712, opened 2026-08-18
adding verification script

MarkWolters:
This PR adds one new script, `verify_native_lib.sh`, to the jvector-native module. The purpose of this script is to verify that the libjvector.so shared library has been correctly generated and contains the expected content prior to pushing a release to Sonatype (although it can also be used ad hoc at any time). 

This PR contains no code changes or modifications to existing files.

Release instructions at https://github.com/riptano/VECTOR-SEARCH/blob/main/teamdocs/jvector_releases_admin.md have been updated to include a step for this check.

### DIFF
--- a/jvector-native/verify_native_lib.sh
+++ b/jvector-native/verify_native_lib.sh
@@ -17,8 +17,7 @@
 # Sanity-checks the libjvector.so that actually ships inside the built
 # jvector-native jar (not the loose copy in src/main/resources): correct
 # architecture, resolvable dynamic dependencies, and presence of the expected
-# exported symbols. Bound to the `verify` phase so it runs as part of
-# `mvn verify`. Non-fatal (warns and skips) for any check whose tool isn't
+# exported symbols. Non-fatal (warns and skips) for any check whose tool isn't
 # available on the current OS (e.g. readelf/ldd on macOS).
 
 set -euo pipefail
@@ -82,7 +81,7 @@ else
 fi
 
 # Note that an exhaustive check of exported symbols is not strictly necessary, because the
-# Java code will fail to load the library if any of the expected symbols are missing. But
+# Java code will load but fail at runtime if any of the expected symbols are missing. But
 # this check is a useful sanity-check to catch any accidental changes to the native code that
 # would break the Java code, and to catch any accidental changes to the build process that
 # would result in a library that doesn't export the expected symbols.
--- a/jvector-native/verify_native_lib.sh
+++ b/jvector-native/verify_native_lib.sh
@@ -81,6 +81,11 @@ else
   echo "WARNING: ldd not available on this OS (e.g. macOS); use 'otool -L ${LIBNAME}' manually if needed." >&2
 fi
 
+# Note that an exhaustive check of exported symbols is not strictly necessary, because the
+# Java code will fail to load the library if any of the expected symbols are missing. But
+# this check is a useful sanity-check to catch any accidental changes to the native code that
+# would break the Java code, and to catch any accidental changes to the build process that
+# would result in a library that doesn't export the expected symbols.
 echo "-- exported symbols (nm) --"
 REQUIRED_SYMBOLS=(
   jvector_simd_get_active_isa
--- a/jvector-native/verify_native_lib.sh
+++ b/jvector-native/verify_native_lib.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+# Copyright DataStax, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


jvector-native/verify_native_lib.sh
```diff
@@ -0,0 +1,113 @@
...
+fi
+
+echo "-- exported symbols (nm) --"
+REQUIRED_SYMBOLS=(
+  jvector_simd_get_active_isa
+  jvector_simd_get_max_isa_env
+  dot_product_f32
+  cosine_f32
+  euclidean_f32
+)
```
r-devulap: Why not add the whole list? 

```suggestion
REQUIRED_SYMBOLS=(
add_in_place_f32
add_scalar_in_place_f32
assemble_and_sum_f32
assemble_and_sum_pq_f32
calculate_partial_sums_dot_f32
calculate_partial_sums_euclidean_f32
calculate_partial_sums_self_magnitude_f32
cosine_f32
dot_product_f32
euclidean_f32
jvector_simd_get_active_isa
jvector_simd_get_max_isa_env
max_f32
min_in_place_f32
nvq_cosine_8bit_packed
nvq_dot_product_8bit
nvq_loss
nvq_quantize_8bit
nvq_shuffle_query_in_place_8bit
nvq_square_l2_distance_8bit
nvq_uniform_loss
pq_decoded_cosine_similarity_f32
sub_in_place_f32
sub_scalar_in_place_f32
)
```

jvector-native/verify_native_lib.sh
```diff
@@ -0,0 +1,113 @@
...
+  if echo "${LDD_OUT}" | grep -qi "not found"; then
+    echo "ERROR: unresolved shared library dependencies detected above." >&2
+    FAIL=1
+  fi
+else
+  echo "WARNING: ldd not available on this OS (e.g. macOS); use 'otool -L ${LIBNAME}' manually if needed." >&2
+fi
+
+echo "-- exported symbols (nm) --"
+REQUIRED_SYMBOLS=(
```
r-devulap, line 89: Maybe add a comment explaining that an exhaustive list is not required here. Successful loading of the .so is sufficient, since the Java side will fail if any expected kernel API is missing. This would also clarify that the list does not need to be maintained as additional kernels are introduced.

r-devulap (approved):
LGTM.

jvector-native/verify_native_lib.sh
MarkWolters: Done in latest checkin

r-devulap (approved):

jvector-native/verify_native_lib.sh
```diff
@@ -0,0 +1,118 @@
...
+  if echo "${LDD_OUT}" | grep -qi "not found"; then
+    echo "ERROR: unresolved shared library dependencies detected above." >&2
+    FAIL=1
+  fi
+else
+  echo "WARNING: ldd not available on this OS (e.g. macOS); use 'otool -L ${LIBNAME}' manually if needed." >&2
+fi
+
+# Note that an exhaustive check of exported symbols is not strictly necessary, because the
+# Java code will fail to load the library if any of the expected symbols are missing. But
```
ashkrisk: nit: I believe the library loads correctly but fails at use when Java tries to jump to a symbol that doesn't exist.

jvector-native/verify_native_lib.sh
```diff
@@ -0,0 +1,118 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Sanity-checks the libjvector.so that actually ships inside the built
+# jvector-native jar (not the loose copy in src/main/resources): correct
+# architecture, resolvable dynamic dependencies, and presence of the expected
+# exported symbols. Bound to the `verify` phase so it runs as part of
+# `mvn verify`. Non-fatal (warns and skips) for any check whose tool isn't
```
ashkrisk: It's not currently bound to `mvn verify`, maybe remove this bit?

ashkrisk (approved):

jvector-native/verify_native_lib.sh
MarkWolters: corrected comment to read "will load but fail at runtime"

jvector-native/verify_native_lib.sh
MarkWolters: nice catch, that was initially my plan but later changed my mind. I have removed this part of the comment.

ashkrisk (approved):

PR #714, opened 2026-08-20
bug fix for non-normalized dot product vectors returning no matches and causing assertion to fail if using hierarchy

### LINKED ISSUE #713: GraphSearcher Assertion can fail for dot product when all candidates score negative
See CNDB ticket https://github.com/riptano/cndb/issues/18237

Astra test testTrueDotproduct fails when we enable by default hierarchy for vector indexes. Logs show 
```DEBUG [ReadStage-1] 2026-06-18 14:50:52,060 QueryController.java:770 - [cql_test_keyspace.table_02.table_02_v_idx] Failed search on index TrieIndexSSTableReader(path='/Users/ekaterinadimitrova/projects/cassandra/build/test/cassandra/data/cql_test_keyspace/table_02-a0e60d306b4611f198d0137652fee767/cc-3h1d_1gcr_5vz682bnpw1x2ykyuf-bti-Data.db'), aborting query.
java.lang.AssertionError: 0
	at io.github.jbellis.jvector.graph.GraphSearcher.internalSearch(GraphSearcher.java:263)
	at io.github.jbellis.jvector.graph.GraphSearcher.search(GraphSearcher.java:228)
	at org.apache.cassandra.index.sai.disk.vector.CassandraDiskAnn.search(CassandraDiskAnn.java:264)
	at org.apache.cassandra.index.sai.disk.v2.V2VectorIndexSearcher.searchInternal(V2VectorIndexSearcher.java:203)
	at org.apache.cassandra.index.sai.disk.v2.V2VectorIndexSearcher.orderBy(V2VectorIndexSearcher.java:173)
	at org.apache.cassandra.index.sai.disk.v1.Segment.orderBy(Segment.java:160)
	at org.apache.cassandra.index.sai.disk.v1.V1SearchableIndex.orderBy(V1SearchableIndex.java:224)
	at org.apache.cassandra.index.sai.SSTableIndex.orderBy(SSTableIndex.java:292)
	at org.apache.cassandra.index.sai.plan.QueryController.lambda$getTopKRows$9(QueryController.java:620)
	at org.apache.cassandra.index.sai.plan.QueryController.searchSSTables(QueryController.java:760)
	at org.apache.cassandra.index.sai.plan.QueryController.searchTopKRows(QueryController.java:728)
	at org.apache.cassandra.index.sai.plan.QueryController.getTopKRows(QueryController.java:621)
	at org.apache.cassandra.index.sai.plan.QueryController.getTopKRows(QueryController.java:101)
	at org.apache.cassandra.index.sai.plan.Plan$ScoredIndexScan.execute(Plan.java:1578)
	at org.apache.cassandra.index.sai.plan.QueryController.buildIterator(QueryController.java:452)
	at org.apache.cassandra.index.sai.plan.StorageAttachedIndexSearcher.search(StorageAttachedIndexSearcher.java:153)
	at org.apache.cassandra.db.ReadCommand.searchStorage(ReadCommand.java:574)
	at org.apache.cassandra.db.ReadCommand.executeLocally(ReadCommand.java:460)
	at org.apache.cassandra.db.ReadCommandVerbHandler.doVerb(ReadCommandVerbHandler.java:74)
	at org.apache.cassandra.net.InboundSink.lambda$new$0(InboundSink.java:80)
	at org.apache.cassandra.net.InboundSink.accept(InboundSink.java:100)
	at org.apache.cassandra.net.InboundSink.accept(InboundSink.java:47)
	at org.apache.cassandra.net.InboundMessageHandler$ProcessMessage.run(InboundMessageHandler.java:440)
	at org.apache.cassandra.concurrent.Stage.lambda$withTimeMeasurement$13(Stage.java:446)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at org.apache.cassandra.concurrent.AbstractLocalAwareExecutorService$FutureTask.run(AbstractLocalAwareExecutorService.java:165)
	at org.apache.cassandra.concurrent.AbstractLocalAwareExecutorService$LocalSessionFutureTask.run(AbstractLocalAwareExecutorService.java:137)
	at org.apache.cassandra.concurrent.SEPWorker.run(SEPWorker.java:119)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:829)
ERROR [ReadStage-1] 2026-06-18 14:50:52,060 AbstractLocalAwareExecutorService.java:169 - Uncaught exception on thread Thread[ReadStage-1,5,main]
java.lang.RuntimeException: java.lang.AssertionError: 0
	at org.apache.cassandra.utils.Throwables.unchecked(Throwables.java:309)
	at org.apache.cassandra.utils.Throwables.cleaned(Throwables.java:328)
	at org.apache.cassandra.index.sai.plan.QueryController.searchSSTables(QueryController.java:772)
	at org.apache.cassandra.index.sai.plan.QueryController.searchTopKRows(QueryController.java:728)
	at org.apache.cassandra.index.sai.plan.QueryController.getTopKRows(QueryController.java:621)
	at org.apache.cassandra.index.sai.plan.QueryController.getTopKRows(QueryController.java:101)
	at org.apache.cassandra.index.sai.plan.Plan$ScoredIndexScan.execute(Plan.java:1578)
	at org.apache.cassandra.index.sai.plan.QueryController.buildIterator(QueryController.java:452)
	at org.apache.cassandra.index.sai.plan.StorageAttachedIndexSearcher.search(StorageAttachedIndexSearcher.java:153)
	at org.apache.cassandra.db.ReadCommand.searchStorage(ReadCommand.java:574)
	at org.apache.cassandra.db.ReadCommand.executeLocally(ReadCommand.java:460)
	at org.apache.cassandra.db.ReadCommandVerbHandler.doVerb(ReadCommandVerbHandler.java:74)
	at org.apache.cassandra.net.InboundSink.lambda$new$0(InboundSink.java:80)
	at org.apache.cassandra.net.InboundSink.accept(InboundSink.java:100)
	at org.apache.cassandra.net.InboundSink.accept(InboundSink.java:47)
	at org.apache.cassandra.net.InboundMessageHandler$ProcessMessage.run(InboundMessageHandler.java:440)
	at org.apache.cassandra.concurrent.Stage.lambda$withTimeMeasurement$13(Stage.java:446)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at org.apache.cassandra.concurrent.AbstractLocalAwareExecutorService$FutureTask.run(AbstractLocalAwareExecutorService.java:165)
	at org.apache.cassandra.concurrent.AbstractLocalAwareExecutorService$LocalSessionFutureTask.run(AbstractLocalAwareExecutorService.java:137)
	at org.apache.cassandra.concurrent.SEPWorker.run(SEPWorker.java:119)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.AssertionError: 0
	at io.github.jbellis.jvector.graph.GraphSearcher.internalSearch(GraphSearcher.java:263)
	at io.github.jbellis.jvector.graph.GraphSearcher.search(GraphSearcher.java:228)
	at org.apache.cassandra.index.sai.disk.vector.CassandraDiskAnn.search(CassandraDiskAnn.java:264)
	at org.apache.cassandra.index.sai.disk.v2.V2VectorIndexSearcher.searchInternal(V2VectorIndexSearcher.java:203)
	at org.apache.cassandra.index.sai.disk.v2.V2VectorIndexSearcher.orderBy(V2VectorIndexSearcher.java:173)
	at org.apache.cassandra.index.sai.disk.v1.Segment.orderBy(Segment.java:160)
	at org.apache.cassandra.index.sai.disk.v1.V1SearchableIndex.orderBy(V1SearchableIndex.java:224)
	at org.apache.cassandra.index.sai.SSTableIndex.orderBy(SSTableIndex.java:292)
	at org.apache.cassandra.index.sai.plan.QueryController.lambda$getTopKRows$9(QueryController.java:620)
	at org.apache.cassandra.index.sai.plan.QueryController.searchSSTables(QueryController.java:760)
	... 20 common frames omitted
```

The bug: internalSearch's hierarchy-descent loop hardcodes a 0.0f threshold that implicitly assumes similarity scores are always ≥ 0:
```
  // GraphSearcher.java:260-265
  for (int lvl = entry.level; lvl > 0; lvl--) {
      searchOneLayer(scoreProvider, 1, 0.0f, lvl, Bits.ALL);
      assert approximateResults.size() == 1 : approximateResults.size();   // <-- line 263
      setEntryPointsFromPreviousLayer();
  }
```
  Inside searchOneLayer, that threshold is used as a hard filter (GraphSearcher.java:409):
```
  if (acceptOrdsThisLayer.get(topCandidateNode) && topCandidateScore >= threshold) {
      addTopCandidate(topCandidateNode, topCandidateScore, rerankK); 
  }
```
  Any candidate whose score is < 0 is silently dropped, never added to approximateResults. If every candidate reachable from the entry point within that hierarchy layer happens to score negative for a given query, approximateResults stays empty and the size() == 1 assert fires — exactly the AssertionError: 0 in the stack trace.

  Why 0.0f is unsafe: it's only a valid "accept everything" sentinel for similarity functions that are mathematically bounded to (0, 1] — COSINE and EUCLIDEAN are. DOT_PRODUCT (VectorSimilarityFunction.java:51-56, (1 + dotProduct(v1,v2)) / 2) is not bounded unless callers pre-normalize vectors to unit length — the class's own javadoc says as much: "Using dot product with vectors that are not normalized can result in errors."

VectorDotProductWithLengthTest.testTrueDotproduct (datastax/cassandra) deliberately uses non-unit 2D vectors with components in [-100, 100]:
```
  // these need to NOT be unit vectors to test the difference between DP and cosine
  return new float[] { R.nextFloatBetween(-100, 100), R.nextFloatBetween(-100, 100) };
```
  With that range, raw dot products routinely fall well outside [-1, 1], so (1+dot)/2 is often strongly negative — exactly the condition that trips the filter above.

  Why it's hierarchy-specific (matches the reporter's observation that disabling hierarchy makes it pass): layer-0 search uses a wide candidate pool and a large rerankK, so the genuinely-best (positively-scored, well-aligned) neighbors are found regardless — it silently drops negative-scored candidates too, but rarely all of them. Upper hierarchy layers search with rerankK=1 from a single fixed entry point — a much narrower search where it's entirely plausible that every candidate examined at that layer scores negative relative to a given query, especially in a small (2000-vector, 2-D) dataset with widely-ranged random vectors.

MarkWolters:
This PR resolves issue https://github.com/datastax/jvector/issues/713.

When dot product vsf is used for graph search in conjunction with enabling hierarchy this can result in a failed Assertion when no candidate is found to be sufficiently close to the search vector at the level being searched, in this case where level > 0.

Any candidate whose score is < 0 is silently dropped, never added to approximateResults. If every candidate reachable from the entry point within that hierarchy layer happens to score negative for a given query, approximateResults stays empty and the size() == 1 assert fires.

0.0f is only a valid "accept everything" sentinel for similarity functions that are mathematically bounded to (0, 1] — COSINE and EUCLIDEAN are. DOT_PRODUCT is not bounded unless callers pre-normalize vectors to unit length. VectorDotProductWithLengthTest.testTrueDotproduct (the failing Cassandra unit test that surfaced this issue) deliberately uses non-unit 2D vectors with components in [-100, 100].

### DIFF
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
@@ -271,8 +271,13 @@ public class GraphSearcher implements Closeable {
 
         // Move downward from entry.level to 1
         for (int lvl = entry.level; lvl > 0; lvl--) {
-            // Search this layer with minimal parameters since we just want the best candidate
-            searchOneLayer(scoreProvider, 1, 0.0f, lvl, Bits.ALL);
+            // Search this layer with minimal parameters since we just want the best candidate.
+            // Threshold must be NEGATIVE_INFINITY, not 0.0f: 0.0f is only a valid "accept everything"
+            // floor for similarity functions bounded to (0, 1] (COSINE, EUCLIDEAN). DOT_PRODUCT with
+            // non-unit-normalized vectors can legitimately score below 0, and a 0.0f floor here would
+            // silently discard the only reachable candidate(s) in a hierarchy layer, leaving
+            // approximateResults empty and tripping the assertion below.
+            searchOneLayer(scoreProvider, 1, Float.NEGATIVE_INFINITY, lvl, Bits.ALL);
             assert approximateResults.size() == 1 : approximateResults.size();
             setEntryPointsFromPreviousLayer();
         }
--- a/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
+++ b/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
@@ -199,11 +199,14 @@ public class GraphSearcher implements Closeable {
 
     /**
      * @param scoreProvider   provides functions to return the similarity of a given node to the query vector
-     * @param topK            the number of results to look for. With threshold=0, the search will continue until at least
-     *                        `topK` results have been found, or until the entire graph has been searched.
+     * @param topK            the number of results to look for. With threshold=Float.NEGATIVE_INFINITY, the search will
+     *                        continue until at least `topK` results have been found, or until the entire graph has been
+     *                        searched.
      * @param rerankK         the number of (approximately-scored) results to rerank before returning the best `topK`.
-     * @param threshold       the minimum similarity (0..1) to accept; 0 will accept everything. May be used
-     *                        with a large topK to find (approximately) all nodes above the given threshold.
+     * @param threshold       the minimum similarity (0..1) to accept; 0 will accept everything for EUCLIDEAN and COSINE.
+     *                        DOT_PRODUCT is not bounded unless callers pre-normalize vectors to unit length and
+     *                        requires Float.NEGATIVE_INFINITY rather than 0. May be used with a large topK to find
+     *                        (approximately) all nodes above the given threshold.
      *                        If threshold > 0 then the search will stop when it is probabilistically unlikely
      *                        to find more nodes above the threshold, even if `topK` results have not yet been found.
      * @param rerankFloor     (Experimental!) Candidates whose approximate similarity is at least this value
@@ -251,8 +254,10 @@ public class GraphSearcher implements Closeable {
      * @param topK            the number of results to look for. With threshold=0, the search will continue until at least
      *                        `topK` results have been found, or until the entire graph has been searched.
      * @param rerankK         the number of (approximately-scored) results to rerank before returning the best `topK`.
-     * @param threshold       the minimum similarity (0..1) to accept; 0 will accept everything. May be used
-     *                        with a large topK to find (approximately) all nodes above the given threshold.
+     * @param threshold       the minimum similarity (0..1) to accept; 0 will accept everything for EUCLIDEAN and COSINE.
+     *                        DOT_PRODUCT is not bounded unless callers pre-normalize vectors to unit length and
+     *                        requires Float.NEGATIVE_INFINITY rather than 0. May be used with a large topK to find
+     *                        (approximately) all nodes above the given threshold.
      *                        If threshold > 0 then the search will stop when it is probabilistically unlikely
      *                        to find more nodes above the threshold, even if `topK` results have not yet been found.
      * @param acceptOrds      a Bits instance indicating which nodes are acceptable results.
@@ -290,8 +295,10 @@ public class GraphSearcher implements Closeable {
      * @param scoreProvider   provides functions to return the similarity of a given node to the query vector
      * @param topK            the number of results to look for. With threshold=0, the search will continue until at least
      *                        `topK` results have been found, or until the entire graph has been searched.
-     * @param threshold       the minimum similarity (0..1) to accept; 0 will accept everything. May be used
-     *                        with a large topK to find (approximately) all nodes above the given threshold.
+     * @param threshold       the minimum similarity (0..1) to accept; 0 will accept everything for EUCLIDEAN and COSINE.
+     *                        DOT_PRODUCT is not bounded unless callers pre-normalize vectors to unit length and
+     *                        requires Float.NEGATIVE_INFINITY rather than 0. May be used with a large topK to find
+     *                        (approximately) all nodes above the given threshold.
      *                        If threshold > 0 then the search will stop when it is probabilistically unlikely
      *                        to find more nodes above the threshold, even if `topK` results have not yet been found.
      * @param acceptOrds      a Bits instance indicating which nodes are acceptable results.
@@ -322,7 +329,7 @@ public class GraphSearcher implements Closeable {
                                int topK,
                                Bits acceptOrds)
     {
-        return search(scoreProvider, topK, 0.0f, acceptOrds);
+        return search(scoreProvider, topK, Float.NEGATIVE_INFINITY, acceptOrds);
     }
 
     @Experimental
--- a/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestHierarchyDotProductSearch.java
+++ b/jvector-tests/src/test/java/io/github/jbellis/jvector/graph/TestHierarchyDotProductSearch.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java
ashkrisk: Most of the Javadocs for the `*search*` functions in this file either imply or explicitly state that threshold = 0 will accept everything. We should probably update all of the affected lines to make it clear that this isn't true, and that users should use `Float.NEGATIVE_INFINITY` if they really want to accept everything.

ashkrisk, line 314: Should we set the rerankFloor parameter here to `Float.NEGATIVE_INFINITY` instead of `0.0f`?

ashkrisk: Similarly, here we might want to change the hard-coded threshold parameter from `0.0f` to `Float.NEGATIVE_INFINITY`

MarkWolters: I don't think so. The floor just determines candidates to be reranked with FP vectors and won't result in a failure the way threshold will. This will only affect a DOT_PRODUCT search on one of the upper hierarchical levels and won't impact L0, so it doesn't seem necessary to me.

MarkWolters: Not strictly necessary. In the offending code `0.0f` was hard-coded as the threshold for upper level searches (see https://github.com/datastax/jvector/blob/main/jvector-base/src/main/java/io/github/jbellis/jvector/graph/GraphSearcher.java#L275) but the passed value of threshold is honored for the l0 search only, where this issue isn't going to occur due to the number of candidates. However I have updated it to Float.NEGATIVE_INFINITY for consistency as this has no impact on the search results.

MarkWolters: updated where relevant (keep in mind there are 2 uses of the word `threshold` in this class which can be confusing: one of them refers to the minimum similarity for a candidate to be considered, the other refers to an early stop condition on searches).

ashkrisk (approved):

PR #718, opened 2026-08-28
Introduce the `InMemoryDataSet` interface.

ashkrisk:
Separate part of functionality provided by the `DataSet` interface into the `InMemoryDataSet` interface to allow for DataSets that are not held in-memory.

The current `DataSet` interface requires all the base vectors in the corpus to be made available as a `List<VectorFloat<?>>`. This makes it difficult to create a `DataSet` implementation that reads data from disk without manifesting it in-memory.

To solve this, the new `InMemoryDataSet` interface is created and the scope of the original `DataSet` interface is reduced. The `DataSet` interface no longer provides all base vectors in memory, however, it retains the ability to acces base vectors through a RAVV.

Summary of changes:
- Split-off `getBaseVectors` from the `DataSet` interface into the `InMemoryDataSet` interface
- References to `getBaseVectors` are scrubbed from `Grid` to allow it to use the reduced `DataSet` interface. No performance impact is expected since `Grid` only used the base vector list for counting the number of vectors.
- `CompactorBenchmark` (and by extension, `AutoBenchYAML`) now uses `InMemoryDataSet` instead of the plain `DataSet`. The list of base vectors is used for partitioning, and porting this to the reduced `DataSet` interface is left to future patches.


MarkWolters:
This one seems to have an issue with the build that needs to be resolved, it is failing for both Bench regression build and compaction regression build (see https://github.com/datastax/jvector/actions/runs/34936963388/job/104276905749?pr=718#step:13:2650)

ashkrisk:
I made a typo in one of the links in the Javadoc. `mvn compile` and `mvn test` didn't care about it, but `mvn package` certainly does.

ashkrisk:
It occurred to me that the existing `DataSet` interface can be used to handle out-of-memory DataSets without modification as long as we return a custom list implementation from `getBaseVectors()`, similar to but not exactly the below:
```java
@Override
public List<VectorFloat<?>> getBaseVectors() {
    return new AbstractList<>() {
        @Override
        public int size() {
            return baseRavv.size();
        }

        @Override
        public VectorFloat<?> get(int index) {
            return baseRavv.getVector(index);
        }
    };
}
```

Marking this PR as draft while I look into the possibility.

PR #723, opened 2026-08-31
Enable native code on AARCH64 

r-devulap:
This pull request adds full support for running and building `jvector-native` on AArch64 (ARM64) systems, including both NEON/SVE/SVE2 SIMD instruction sets. The changes introduce architecture detection macros, update build scripts and CI workflows for ARM64, and extend runtime detection and dispatch of SIMD capabilities to support ARM64 alongside x86_64. This enables native vectorization and testing on ARM64 platforms such as AWS Graviton and Apple Silicon.

Key changes:

**ARM64 (AArch64) support:**

* Added architecture detection macros (`JV_ARCH_X86_64`, `JV_ARCH_AARCH64`) in new header `jvector_arch.h` and refactored source code to use these macros for architecture-specific logic. 
* Implemented runtime CPU feature detection for NEON, SVE, and SVE2 on ARM64, supporting both Linux (using `getauxval`) and macOS (using `sysctlbyname`).
* Updated the native dispatch logic to handle ARM64 SIMD tiers (NEON, SVE and SVE2), mirroring the x86_64 dispatch mechanism. 
* Kernels needed minor changes to adapt to vector length agnostic SIMD implementations (28d0a8400a562669d87952fbaf46a198f7316fbd) 

**Build system and CI improvements:**

* Modified `meson.build` to add ARM64 ISA variants, select appropriate test sources, and error on unsupported architectures. 
* Added a new GitHub Actions workflow to run unit tests on ARM64 runners, building and testing all supported ISAs and JDK versions.

**Java integration:**

* Updated `NativeVectorizationProvider` to allow native SIMD operations on both x86_64 and aarch64, not just x86_64.
* Relaxed the Maven profile activation in `pom.xml` to allow non-amd64 Unix builds (enabling ARM64 builds).


jvector-native/src/main/native/meson.build
ashkrisk, line 1: Have you considered how packaging is going to work? Ideally, a JAR that we publish on maven central would run as correctly on both amd64 and aarch64 systems, but the current JAR will only contain the native lib corresponding to the architecture on which it was built.

I'd suggest looking into options for cross-compiling. This allows us to keep the existing packaging process, and build the aarch lib alongside the amd64 lib on the same machine, which allows them to be packaged together.

jvector-native/src/main/native/src/assert_hwy_targets.h
```diff
@@ -14,22 +14,49 @@
...
+// Compiler flags per tier and the Highway target each must produce:
+//   neon:  -march=armv8-a+crypto   → HWY_NEON
+//          (no BF16/dotprod/I8MM features, so never HWY_NEON_BF16)
+//   sve:   -march=armv8.4-a+sve    → HWY_SVE  (scalable, VL-agnostic)
+//          (Graviton 3 / Neoverse V1 and later)
+//   sve2:  -march=armv9-a+sve2     → HWY_SVE2 (scalable, VL-agnostic)
+//          (Graviton 4 / Neoverse V2/N2 and later)
+#if defined(JV_REQUIRE_HWY_SVE2)
+#  if HWY_STATIC_TARGET != HWY_SVE2
+#    error "Highway did not select HWY_SVE2 for the SVE2 build. Check compiler flags (-march=armv9-a+sve2) and Highway blocklists."
```
ashkrisk, line 50: This breaks JV source builds on my M1 mac. An ARM CPU is detected and JV tries to compile for NEON, SVE and SVE2, but the latter two builds fail this assertion.

```sh
$ mvn clean package
In file included from ../../src/main/native/src/jvector_simd_kernels.cpp:23:
../../src/main/native/src/assert_hwy_targets.h:50:6: error: "Highway did not select HWY_SVE2 for the SVE2 build. Check compiler flags (-march=armv9-a+sve2) and Highway blocklists."
```

ashkrisk (commented):
Had an initial look, ran the code on a Google Axion VM (which worked) and on an M1 Mac (which needed some tweaking). Numbers on both look better than Panama!

We do need a plan for publishing ARM artifacts before we officially claim support for it. Most people would like to just grab a universal JAR from Maven Central.

jvector-native/src/main/native/src/assert_hwy_targets.h
r-devulap: what compiler and version are building it with?

ashkrisk:
```
Apple clang version 17.0.0 (clang-1700.0.13.5)
```
Meson autoselects `clang`, but I don't know if that's the real problem, M1 doesn't support SVE so this should fail regardless.

Even after fixing this there are other things needed to support MacOS (the library is a `dylib` and not a `so` for example). I'd say we just exclude non-Linux systems for now.

r-devulap: Yeah, i think its easier if we disable on macOS for now. Also, looks like SVE and SVE2 requires clang > v22. I have updated meson.build to ignore SVE and SVE2 builds when clang doesn't support it.

r-devulap: See 3b7b58b0a8daceb37198b0ed9aac913c7894420f

jvector-native/src/main/native/meson.build
r-devulap: This needs more work, I am able to cross build an arm library on x86 using `aarch64-linux-gnu-g++` which in theory should be equivalent to using `g++` on arm64 natively. Unfortunately, the shared libraries built with the same version of `g++` on arm and  `aarch64-linux-gnu-g++` on x86 don't match exactly (not diff clean) and its hard to sift through what is different. Another way to reason this is: if I cross compile an arm build on x86 and then run it on aarch64, do I get the same performance? I will try to get some numbers to validate that.

r-devulap: >  Unfortunately, the shared libraries built with the same version of g++ on arm and aarch64-linux-gnu-g++ on x86 don't match exactly (not diff clean) 

This was probably because `g++` native on arm was 13.4.0 v/s `aarch64-linux-gnu-g++ (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0` on x86.  When the compiler versions match exactly the library is also identical. Also confirmed that the performance is the same: benchYAML recall@10 v/s latency curves pretty much look the same irrespective of where you compile them.

r-devulap:
> We do need a plan for publishing ARM artifacts before we officially claim support for it. Most people would like to just grab a universal JAR from Maven Central.

249fcffa84c9fdc20bd7e6a83b9367f6a740ad31 addresses that. Added a maven flag `-Dnative.crossarch` which enables compiling two versions of the library: one for aarch64 and one for x86_64. The build script has been updated to include both the libraries into the jar and at runtime java picks the right library depending on which platform you run on. 

@MarkWolters This needs your approval :) But given it has no performance implications (see https://github.com/datastax/jvector/pull/723#discussion_r3976982913) I am guessing this is the easiest option for building the release jar?

r-devulap:
> Added a maven flag -Dnative.crossarch

`-Dnative.crosscompile` sounds better? open to better name suggestions :)

MarkWolters:
Ran a small comparison test of this branch against main on an ARM machine (Neon) and the difference looks very promising. Way better on this branch at low OQ, although as the OQ goes up the differences converge a bit. Still better performance numbers at every level of OQ without losing an recall or nodes visited. 

Run on AWS `c6g.2xlarge`

main:
```
Index configuration:
  featureSetForIndex   [NVQ_VECTORS]
  M                    32
  efConstruction       100
  neighborOverflow     1.2
  addHierarchy         true
  refineFinalGraph     true

Query configuration:
  usePruning           true

Overquery    Avg QPS         ± Std Dev    CV %        Mean Latency    Avg Visited    Recall@10    Max heap        Max offheap    
             (of 3)                                   (ms)                                        usage (MB)      usage (MB)     
---------------------------------------------------------------------------------------------------------------------------------
1.00         8601.0          204.4        2.4         0.924           291.3          0.75         1031.0          164.4          
2.00         7716.5          43.4         0.6         1.049           398.4          0.91         1008.7          164.4          
5.00         5912.5          31.1         0.5         1.373           673.3          0.97         1008.7          164.4          
10.00        4323.5          7.8          0.2         1.900           1092.3         0.99         1008.1          164.4          


Overquery    Avg QPS         ± Std Dev    CV %        Mean Latency    Avg Visited    Recall@100    Max heap        Max offheap    
             (of 3)                                   (ms)                                         usage (MB)      usage (MB)     
----------------------------------------------------------------------------------------------------------------------------------
1.00         4292.0          27.3         0.6         1.901           1092.3         0.82          1233.1          164.4          
2.00         2758.1          20.9         0.8         2.912           1890.8         0.96          1241.1          164.4          
```

arm64 (this branch):
```
Index configuration:
  featureSetForIndex   [NVQ_VECTORS]
  M                    32
  efConstruction       100
  neighborOverflow     1.2
  addHierarchy         true
  refineFinalGraph     true

Query configuration:
  usePruning           true

Overquery    Avg QPS         ± Std Dev    CV %        Mean Latency    Avg Visited    Recall@10    Max heap        Max offheap    
             (of 3)                                   (ms)                                        usage (MB)      usage (MB)     
---------------------------------------------------------------------------------------------------------------------------------
1.00         15103.7         370.5        2.5         0.513           297.0          0.75         957.9           164.4          
2.00         12105.3         212.4        1.8         0.662           403.3          0.92         1145.2          164.4          
5.00         8093.8          43.7         0.5         0.980           676.7          0.97         1169.3          164.4          
10.00        5082.8          41.1         0.8         1.527           1096.4         0.99         1169.3          164.4          


Overquery    Avg QPS         ± Std Dev    CV %        Mean Latency    Avg Visited    Recall@100    Max heap        Max offheap    
             (of 3)                                   (ms)                                         usage (MB)      usage (MB)     
----------------------------------------------------------------------------------------------------------------------------------
1.00         5151.5          15.5         0.3         1.517           1096.4         0.82          1031.2          164.4          
2.00         3113.6          17.7         0.6         2.563           1895.4         0.96          1039.2          164.4          
```

MarkWolters (changes requested):
I think this looks very good, and the performance testing at small scale looks impressive. I'll run a larger regression run to verify against larger datasets but I don't see any problems from my review except for one..please add release notes ;-)

PR #730, opened 2026-09-17
Mark nodes completed during rescoring

MarkWolters:
This PR is a bug fix for the issue reported by the Cassandra team (https://github.com/riptano/cndb/issues/19259) that results in the first 128k vectors copied during compaction being invisible to GraphIndex.View, causing recall scores to plummet after compaction. 

When rescore() is called the nodes specified for rescoring are never marked as completed, so they are hidden from the subsequent Views. This PR adds a call to `markComplete()` for each node added during the call to `rescore()`. The workaround for this issue implemented by the Cassandra team marks added nodes complete in a batched fashion after rescoring but is otherwise equivalent. It is detailed at https://github.com/datastax/cassandra/pull/2617.


jvector-tests/src/test/java/io/github/jbellis/jvector/graph/GraphIndexBuilderTest.java
```diff
@@ -128,6 +128,55 @@ public void testRescore(boolean addHierarchy) {
...
     }
 
+    @Test
+    public void testRescoreMarksCopiedNodesComplete() {
+        testRescoreMarksCopiedNodesComplete(false);
+        testRescoreMarksCopiedNodesComplete(true);
+    }
+
+    /**
+     * Regression test: {@link GraphIndexBuilder#rescore} is used by Cassandra to refine the PQ
```
tlwillke: I don't think we should have Cassandra-specific comments in the library.

tlwillke (approved):
LGTM

jvector-tests/src/test/java/io/github/jbellis/jvector/graph/GraphIndexBuilderTest.java
MarkWolters: removed references to Cassandra from comments

jshook (approved):

